PR #3, opened 2022-08-07
Print more debug info using verbose=True

kmax12:


WillKoehrsen (commented):
I think this PR can be closed since most methods log info when `verbose=True`.

PR #50, opened 2022-10-18
Adding "infer" param to tz conversion

kbaranko:
Calling the historical prices method for PJM throws a `AmbiguousTimeError` on days when we "fall" back. Caught this while scraping price data for 2021 and assuming it will happen on daylight savings this year as well. 

To reproduce: 
`iso.get_historical_lmp(
    date="Nov 7, 2021",
    market='REAL_TIME_HOURLY', 
    locations=['50655'],
)`

### DIFF
--- a/isodata/pjm.py
+++ b/isodata/pjm.py
@@ -270,7 +270,7 @@ class PJM(ISOBase):
 
         data["Time"] = pd.to_datetime(data["Time"]).dt.tz_localize(
             self.default_timezone,
-            ambiguous='infer',
+            ambiguous="infer",
         )
 
         data = data[
--- a/isodata/pjm.py
+++ b/isodata/pjm.py
@@ -270,6 +270,7 @@ class PJM(ISOBase):
 
         data["Time"] = pd.to_datetime(data["Time"]).dt.tz_localize(
             self.default_timezone,
+            ambiguous='infer',
         )
 
         data = data[


kbaranko:
Ran the tests locally with and without the changes I made and the same three kept failing. Pretty sure it's due to the MISO API being offline. Love the lint-fix! 

Happy to add some test coverage and took a peek at what you have so far. Curious to get your thoughts on desired extensiveness... I could add a `Nov 7th, 2021` test day assertion to all tests for historical data methods for complete visibility into which method x ISO x market combinations have issues with DST. I have not checked extensively myself, but know that CAISO markets had no issues with DST while PJM required adding the `ambiguous='infer'` param. Using a 2021 date would also cause problems down the line due to the rolling window (at least for CAISO)

I also could just add a test for the PJM case I found and edited then move on.

kmax12:
looking into this more, I think there is a better fix.

basically, the issue is we are parsing the time column where time is represented by Eastern Time which has a abiquity on days when the time shift backwards and 1am repeats itself. We can instead use the API to return UTC time (which isn't affected by DST) and then convert that to Eastern Time. that way there is no ambiguity. 

this is a screen shot from pjm's site. first column is utc, second is eastern time
image

I am going to fix this a part of resolving another issue, so I think we can close the PR for now. 

Thanks for bringing this issue to my attention!

Issue #52, opened 2022-10-20
ISO.get_historical_lmp : add feature to query historical time period


m4rz910:
Hey, would be nice to be able to pull large time spans instead of individual days at a time. Thanks!

m4rz910:
In general getting a key error "items" when I run the code...
``
o = isodata.PJM()
o.get_historical_lmp(date='1/1/2022', market="REAL_TIME_5_MIN", locations='ALL')
``

kmax12:
Good timing since I'm working on this in this pr to support date ranges: #53. I will add support for PJM ASAP (hopefully later today) and release it. 

In terms of the key error, there are 2 issues.

1.  "ALL" isn't supported for PJM. If you pass `None` or don't provide parameter, it will return all the Hubs. Alternatively, you can call `pjm.get_pnode_ids()` to see all the valid pnode ids. 
2. The PJM API  doesn't appear to have data going back before 4/17/2022. It's possible it is a 6 month rolling window. I'll need to look into more, but let me know if you are able to access earlier data anywhere. 

Here's a query that works for me

```
o = isodata.PJM()
o.get_historical_lmp(date='4/17/2022', market="REAL_TIME_5_MIN", locations=None)
```
```
	Time	Market	Location	Location Name	Location Type	LMP	Energy	Congestion	Loss
0	2022-04-17 00:00:00-04:00	REAL_TIME_5_MIN	51217	EASTERN HUB	HUB	44.378613	45.3	-0.725136	-0.196251
1	2022-04-17 00:00:00-04:00	REAL_TIME_5_MIN	51287	WEST INT HUB	HUB	45.638333	45.3	0.364444	-0.026111
2	2022-04-17 00:00:00-04:00	REAL_TIME_5_MIN	51288	WESTERN HUB	HUB	44.359686	45.3	-0.601354	-0.338960
3	2022-04-17 00:00:00-04:00	REAL_TIME_5_MIN	4669664	NEW JERSEY HUB	HUB	44.252114	45.3	-0.676695	-0.371191
4	2022-04-17 00:00:00-04:00	REAL_TIME_5_MIN	33092311	CHICAGO GEN HUB	HUB	44.118462	45.3	0.088923	-1.270462
...	...	...	...	...	...	...	...	...	...
3463	2022-04-18 00:00:00-04:00	REAL_TIME_5_MIN	34497125	AEP GEN HUB	HUB	58.237632	56.7	1.995789	-0.458158
[...]

PR #54, opened 2022-10-20
Improve PJM Historical Data Querying

kmax12:
PJM has a "standard" and "archived" LMP data at the same endpoints, but they support different query patterns based on how far back you query. Additionally, date range queries crossing the archive/standard switch over date or year boundaries (e.g 2021 to 2022) aren't supported. This PR resolves those issues, so user doesn't have to handle this details themselves.

## Example Notebook

Here is an [example notebook](https://github.com/kmax12/isodata/blob/57ccef6777de4eb19b3a63a462750d0e2f32972a/Examples/PJM%20LMP%20Data.ipynb) accessing historical LMP data for PJM

## Details

Documentation of archiving behavior can be found [here](https://www.pjm.com/-/media/etools/data-miner-2/data-miner-2-historic-data-guide.ashx) and excerpted below

image

References:
* Fixes historical queries for PJM mentioned in #52 
* Fixes DST issues mentioned in #50


kmax12:
Phew..this turned into a doozy, but looks everything is working now :)

PR #55, opened 2022-10-21
Nyiso LMP updates 

kbaranko:
Hi @kmax12  I've been making a few edits in my local version of `nyiso.py` file in order to grab the data I need for client and thought you might be interested. 

The commit enables a user to get LMP data for each generator, which I saw you had already marked as a #todo. I added an optional param to the `get_historical_lmps` and `get_lmp_today` methods and created some helper methods to set the filename because the endpoints are different depending on whether you want zone or gen. I can help add / fix tests once you review and modify the approach I took (or scrap it entirely!). 

Additionally: 
- Running into the same DST problem as we discussed in #50, although since the nyiso logic is downloading a bunch of archived csvs I don't think you can use the pjm solution. I also suspect that simply adding the 'infer' param won't work either because the response doesn't have a unique timestamp per row (due to there being multiple locations in the request) and therefore throws a `*** pytz.exceptions.AmbiguousTimeError: There are 29 dst switches when there should only be 1.`
- I've only seen hourly day ahead market data coming out of the nyiso methods (and had only ever heard of hourly day ahead markets in general) so I think `DAY_AHEAD_5_MIN` might be misnamed?


kmax12 (commented):
looking good! a few comments and this should be good to merge

kmax12:
interesting point re:DST. bummer that nyiso format doesn't make this easy to parse these dates correctly

one approach I can think of is to determine if the day is a DST switch, the parse the file in two chunks (one before change and one after), and then combine. 

what do you think of this approach? 

also, I played around with things a bit and this code could be used to determine if it is a transition day

```
def is_transition(date):
    return date.dst() != (date + pd.DateOffset(1)).dst()

assert is_transition(pd.Timestamp("11/7/2021", tz=isodata.NYISO.default_timezone)) == True
assert is_transition(pd.Timestamp("3/14/2021", tz=isodata.NYISO.default_timezone)) == True
assert is_transition(pd.Timestamp("12/01/2021", tz=isodata.NYISO.default_timezone)) == False
```

kmax12:
just noticed something interesting...for load, fuel mix NYISO returns a Time Zone column. For forecast and LMP data it doesnt.

here's load as an example: http://mis.nyiso.com/public/P-58Blist.htm

kmax12:
just a head up, I'm working on a fix for the DST issue in #58. Stepping away for a bit now, but will try to circle back tonight or tomorrow to wrap up and merge

kbaranko:
Handling some local issues invoking `pytest` and getting a `ImportError while loading conftest` from `ModuleNotFoundError: No module named '_tkinter'`. I think I've figured it out though. Will make the rest of your suggested edits later today / as soon as I can and will update this branch to keep it fresh with `main` once you merge #58  

At first glance your approach looks good to me, can give it a deeper check later.

kmax12:
> ImportError while loading conftest from ModuleNotFoundError: No module named '_tkinter'

looks like an errant import in pjm.py or maybe in another file. that can be removed

kbaranko:
@kmax12 hey no worries, handled the conflicts without any issues but having two nyiso tests fail in similar ways both locally and upon pushing. Curious if you see the issue; I'm implementing some of your earlier feedback on this PR then can investigate.

kmax12:
pretty sure that issue was my PR changed some function arguments. just pushed a fix that updates the call to use named keyword arguments.

kbaranko:
Pushed some commits in response to your feedback but let me know if you'd like me to modify the approach to testing etc. Was having trouble with the automated checks but looks like you've fixed it.

isodata/tests/test_nyiso.py
```diff
@@ -81,3 +82,33 @@ def test_nyiso_est_to_edt():
...
+
+    date = "2022-06-09"
+
+    df_zone = iso.get_historical_lmp(
+        date=date, market=Markets.DAY_AHEAD_HOURLY, location_type="zone"
+    )
+    df_gen = iso.get_historical_lmp(
+        date=date, market=Markets.DAY_AHEAD_HOURLY, location_type="generator"
+    )
+    assert df_gen.shape[0] > df_zone.shape[0]
```
kmax12, line 99: I think a better with would be to check that all the rows in the response match the location type 

for example

```python
assert (df_gen["Location Type"] == "Generator").all()
```

isodata/nyiso.py
```diff
@@ -189,24 +203,24 @@ def get_historical_lmp(
         end=None,
         market: str = None,
         locations: list = None,
+        location_type: str = None,
     ):
         """
-        Supported Markets: REAL_TIME_5_MIN, DAY_AHEAD_5_MIN
+        Supported Markets: REAL_TIME_5_MIN, DAY_AHEAD_HOURLY
+
+        Supported Location Types: ZONE, GENERATOR
```
kmax12, line 211: is it supposed to "zone" and "generator" all lowercase?

kmax12:
left two comments, but otherwise looks ready to merge.

isodata/nyiso.py
kbaranko: yes the strings are lowercase but they were used so frequently I put them as constants at the top of the file. is it less confusing to write out the strings here?

kmax12: ya, because the strings are what a user needs to pass to the function

kmax12:
💯 looks ready to merge to me. anything else you want to add? if not, I'm going to merge and cut a new release of isodata

kbaranko:
All set on my end, will raise an issue if I run into any trouble. Thank you!

PR #63, opened 2022-10-26
Add Python 3.11 markers and pytest-xdist to run unit tests in parallel

gsheni:
- Python 3.11 was released, so this adds markers in `pyproject.toml`
- Add Python 3.11 to unit tests and install test
- Add pytest-xdist to run unit tests in parallel (213s went down to 54s locally, on GitHub actions unit tests time went down to half)


kmax12:
Awesome – looks great. I'm seeing a 2x speed up locally

PR #75, opened 2022-11-02
NYISO feature: capacity prices

m4rz910:
1. Added a method to nyiso.py that pulls the latest capacity price report and parses it for latest record of prices (past ~5 years of monthly prices for the three auctions)
2. Added a simple test for the method in tests/test_nyiso.py
3. Added an example notebook to show what the resulting dataframe looks like

Things to consider in your review...
1. I noticed in the base url there is a 8 digit number (after 210142) that changes with every year and I am not sure what it represents. Come 2023, and every new year after that, I expect we will need to update the code or figure out what it means to handle it programmatically.


gridstatus/nyiso.py
```diff
@@ -476,6 +476,33 @@ def _download_nyiso_archive(self, date, end=None, dataset_name=None, filename=No
 
         return df
 
+    def get_capacity_prices(
+        self, date=pd.Timestamp.now(tz="US/Eastern"), verbose=False
```
kmax12, line 480: small python thing -- default arguments are only evaluated once when function is initiated. that means `date` will be the same value on every call. probably not a huge deal for this method since it only cares about month, but might as well change


```python
>>> import pandas as pd
def default_args_test(test=pd.Timestamp.now()):
    print(test)>>> def default_args_test(test=pd.Timestamp.now()):
...     print(test)
... 
>>> default_args_test()
2022-11-02 09:40:19.540002
>>> default_args_test()
2022-11-02 09:40:19.540002
```

you probably want to do something like this instead

```python
>>> def default_args_test(test=None):
...     if test is None:
...         test = pd.Timestamp.now()
...     print(test)
... 
>>> default_args_test()
2022-11-02 09:42:30.128012
>>> default_args_test()
2022-11-02 09:42:31.481343
```

docs/Examples/NYISO Capacity Prices.ipynb
```diff
@@ -0,0 +1,331 @@
+{
```
kmax12, line 1: this notebook looks good as is.

if you want it to show up in the [Examples Section](https://docs.gridstatus.io/en/latest/Examples/index.html) of the documentation, add it to `Examples/index.md`. if you do want to do that, I suggest add at least one graph/visualization with plotly like the other notebooks have

kmax12: perhaps something similar to what they have in last few tabs of the excel file youre grabbing the data from

image

gridstatus/nyiso.py
```diff
@@ -476,6 +476,33 @@ def _download_nyiso_archive(self, date, end=None, dataset_name=None, filename=No
...
+        # todo: it looks like the "27447313" component of the base URL changes every year but I'm not sure what the link between that and the year is...
+        capacity_market_base_url = (
+            "https://www.nyiso.com/documents/20142/27447313/ICAP-Market-Report"
+        )
+
+        url = f"{capacity_market_base_url}-{date.month_name()}-{date.year}.xlsx"
+        if verbose:
+            print(url)
+
+        df = pd.read_excel(url, sheet_name="MCP Table", header=[0, 1])
```
kmax12, line 500: any value in all adding data from the "UCAP Table" or "Summary Table" sheets? if not now, we can always do it in a future PR if someone needs it

kmax12 (commented):
Thanks for the contribution! Just a few comments and then we will be good to merge

gridstatus/nyiso.py
m4rz910: Good question, I had a similar thought when making it but decided to stick to just grabbing prices for now since that's the most important/commonly needed data. Unless someone is trying to get super detailed and trying to understand the underlying mechanics of the market and maybe predict the prices would they benefit from looking at that data; in my experience this a lot less common than just grabbing the prices for a historical analysis or something. As you said we can easily add it later if needed.

kmax12:
Thanks for the contribution! Will go out in next release

PR #80, opened 2022-11-07
Update Ercot Fuel Mix Parsing

kmax12:
Change in ercot return format. This PR updates parsing


kmax12:
failing tests are related to an ISONE daylight savings time parsing issue unrelated to the change, so im going to merge

PR #86, opened 2022-11-21
Error handling on likely missing date range in CAISO LMPS

Kladar:
For some methods that hit CAISO OASIS (such as `get_lmp()` ) if a start_date entered does not have valid data, the whole query fails (rather than return the data it can find, or if it can't find any return a useful error message). So I added a simple try/except block for the point in the loop where this is failing (technically fails and returns a `ValueError` on not having the correct columns for `Usecols` within `pd.read_csv()`). There is definitely more robust error handling than what I added that could be here, but wanted to figure out if I was thinking about this correctly first before going hog wild. 

- Example good date range, e.g. October 1st, 2022 start_date:

image

- Example bad date range, e.g. Jan 1st, 2017 start_date:

image

I haven't exhausted that this could be a data volume issue rather than a valid date range issue, since Jan 2017 doesn't seem like it would be that far back for a hub price, but I did successfully pull a few years worth and the loader bar appears each time, so it seems it can handle the multi-year pull just fine (but not when there isn't data for the start_date)

Let me know if this makes sense. Kind of an excuse to make my first contribution!

### DIFF
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -714,10 +714,14 @@ def _get_oasis(url, usecols=None, verbose=False):
 
     z = ZipFile(io.BytesIO(r.content))
 
-    df = pd.read_csv(
-        z.open(z.namelist()[0]),
-        usecols=usecols,
-    )
+    try:
+        df = pd.read_csv(
+            z.open(z.namelist()[0]),
+            usecols=usecols,
+        )
+    except ValueError:
+        print(f"There was an issue converting the returned zipfile from CAISO OASIS. It could be that\
+        there was no data for the date range selected. Try again with a different date range.")
 
     return df
 


kmax12:
thanks for the contribution.  I've heard through the grapevine that CAISO oasis only publishes that last ~39 months of data. That adds up since the earliest date that is working for me right now is August 8, 2019

```python
iso = gridstatus.CAISO()
iso.get_lmp(date="Aug 8, 2019", market="DAY_AHEAD_HOURLY", verbose=True)
```

When you request a date before that, CAISO actually returns an XML file in the zip which contains an error message. If you call `get_lmp` with `verbose=True`, gridstatus will print out the url that it is attempting to download and then it's easy to go grab the zip file and see the error yourself. 

```python
# doesn't work
iso.get_lmp(date="Aug 7, 2019", market="DAY_AHEAD_HOURLY", verbose=True)
```

image

I think a more generalizable fix here would be to try to parse the XML in the try except and print the error message that is contained in there. 

```python
from bs4 import BeautifulSoup
data = z.open(z.namelist()[0])
parsed_xml = BeautifulSoup(data, "xml")
parsed_xml.find_all('ERR_DESC')
```
that finds
```
[<m:ERR_DESC>No data returned for the specified selection</m:ERR_DESC>]
```

And then perhaps we can raise that text error message with some text like

```The following error occured when retreiving data from CAISO OASIS : "No data returned for the specified selection"```

Finally, if you're up for it can you add a test that the error is raise in `test_caiso.py`? An example of writing a test to catch an error can be seen in `test_pjm.py::test_pjm_no_data`

**Comment by Kladar** [comment]
Nice, sounds good I will take a look and 
1. parse the actual XML error
2. write the corresponding test

**Comment by WillKoehrsen** [comment]
This PR still seems useful insofar as you get an error if trying to request CAISO LMP data too far in the past. I think we could add an early `return` to `get_lmp`:

```python
if df.empty:
    log(f"No data found for date: {date} and end: {end}", verbose=verbose)
    return df
```

Then, if you request data from a time range spanning both missing and present data, the request should still succeed. 

BTW, playing around with the data, it appears the earliest LMP data for CAISO is from 1200 days ago (about 40 months).

WillKoehrsen:
Should be resolved by #331.

PR #94, opened 2022-12-05
ERCOT: get_load supports more historical data

jcsalterego:
For ERCOT, adds support for load data from https://www.ercot.com/content/cdr/html/actual_loads_of_forecast_zones.html as mentioned in the TODO comments.

Highlights:

* Still pulls from the JSON endpoint when `date == 'currentDay'`
* Otherwise, pulls from HTML endpoint, up to 14 days ago
* Requires `lxml` due to `pd.read_html`

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -101,6 +101,7 @@ class Ercot(ISOBase):
         else:
             raise NotSupported()
 
+    @support_date_range("1D")
     def get_load(self, date, verbose=False):
         if date == "latest":
             d = self._get_load_json("currentDay").iloc[-1]
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -86,7 +86,7 @@ class Ercot(ISOBase):
             df = pd.DataFrame(r["currentDay"]["data"].values())
             df = df.dropna(subset=["actualSolar"])
 
-            df = self._handle_data(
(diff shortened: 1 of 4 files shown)


gridstatus/ercot.py
```diff
@@ -101,28 +103,37 @@ def get_fuel_mix(self, date, verbose=False):
 
     def get_load(self, date, verbose=False):
```
kmax12, line 104: now that this supports multiple days, it would make sense add the following decorator.

```python
@support_date_range("1D")
def get_load(self, date, verbose=False):
...
```

This allows the user to call the function with a start and and date, but the `1D` tells grid status to only call the underlying function with a frequency one day at a time. 

It also has handle for parsing date strings and objects, so a user can pass in something like `ercot.get_load("dec 1, 2022")`.

you can see an example on the `get_as_prices` function

kmax12 (changes requested):
this is great! just one comment and then I think it's ready to merge.

gridstatus/ercot.py
jcsalterego: Makes sense!

jcsalterego:
It's definitely timezone-related but I cannot reproduce the issue locally, yet. It doesn't help that it's comparing to live data from ERCOT.

kmax12:
looks great. merging now. really appreciate the contribution!

PR #97, opened 2022-12-16
ERCOT: Settlement Point Prices

jcsalterego:
### Description

Adds `Ercot::get_spp()` with support for:

* Dates (Real-Time Market): `latest`, `today`, varying historical
* Dates (Day-Ahead Market): `today`, varying historical and future (+1 day) dates
* Markets: `REAL_TIME_15_MIN`, `DAY_AHEAD_HOURLY`
* Location Types: `ZONE`, `HUB`, `NODE`

Example notebook here: [https://isodata--97.org.readthedocs.build/en/97/Examples/ercot/Settlement%20Point%20Prices.html](https://isodata--97.org.readthedocs.build/en/97/Examples/ercot/Settlement%20Point%20Prices.html)


jcsalterego:
DAM tests failed now (Sun Dec 18 11:08:53 CST 2022) when there isn't a published report for today. Hmm...

image

jcsalterego:
OK, fixed the failing test, as I never used a -1 day offset for pulling DAM data.

gridstatus/ercot.py
```diff
@@ -432,6 +501,253 @@ def get_interconnection_queue(self, verbose=False):
...
+                "Market",
+                "Location Type",
+                "LMP",
+            ]
+        ]
+
+        df = df.reset_index(drop=True)
+        return df
+
+    def _get_lmp_rtm15_latest(
```
kmax12, line 635: i wonder if this method should be combined with `_get_lmp_rtm15_today`.

it seems they are mostly duplicate logic with the except of getting the documents. from what i can tell everything else is the same. 

I propose we simplify `get_lmp` to call a `get_spp_rtm_15(date, ...)` method and we move the logic of handling `latest` vs `today` to that function so it's easier to reuse everything else.

gridstatus/ercot.py
```diff
@@ -432,6 +501,253 @@ def get_interconnection_queue(self, verbose=False):
...
+        self,
+        locations: list = None,
+        location_type: str = None,
+        verbose=False,
+    ):
+        """Get Real-time 15-minute Market LMP data for ERCOT
+
+        https://www.ercot.com/mp/data-products/data-product-details?id=NP6-788-CD
+        """
+        today = pd.Timestamp(pd.Timestamp.now(tz=self.default_timezone).date())
```
kmax12, line 645: i believe this can be written using `normalize()` instead

```
>>> pd.Timestamp(pd.Timestamp.now().date())
Timestamp('2022-12-20 00:00:00')
>>> pd.Timestamp.now().normalize()
Timestamp('2022-12-20 00:00:00')
>>> pd.Timestamp.now().normalize() ==  pd.Timestamp(pd.Timestamp.now().date())
True
```

i'm guessing you were following existing code used elsewhere in this file, so if you could update those places too, that'd be very appreciated.

gridstatus/ercot.py
```diff
@@ -432,6 +501,253 @@ def get_interconnection_queue(self, verbose=False):
...
+        self,
+        locations: list = None,
+        location_type: str = None,
+        verbose=False,
+    ):
+        """Get Real-time 15-minute Market LMP data for ERCOT
+
+        https://www.ercot.com/mp/data-products/data-product-details?id=NP6-905-CD
+        """
+        today = pd.Timestamp(pd.Timestamp.now(tz=self.default_timezone).date())
```
kmax12, line 698: same comment about normalize as above

utils/ercot/fully-describe-data-product.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
kmax12, line 1: these scripts are nifty! 

just to confirm, they were manually run and then the results copied and pasted into `ercot.py`? just trying to think through in the future when someone want to implement a new report.

is the idea that they would run these scripts again to add to `ercot.py`? if so, we should leave a comment in that file about these so they are discoverable by future contributors

gridstatus/ercot.py
```diff
@@ -491,6 +845,73 @@ def _handle_html_data(self, df, columns):
...
+            df = df[df["SettlementPointType"].isin(RESOURCE_NODE_SETTLEMENT_TYPES)]
+        elif norm_location_type == LOCATION_TYPE_ZONE:
+            df = df[df["SettlementPointType"].isin(LOAD_ZONE_SETTLEMENT_TYPES)]
+        elif norm_location_type == LOCATION_TYPE_HUB:
+            df = df[df["SettlementPointType"].isin(HUB_SETTLEMENT_TYPES)]
+        else:
+            raise ValueError(f"Invalid location_type: {location_type}")
+        return df
+
+    def _filter_by_locations(self, df, field_name, locations):
```
kmax12, line 861: there is an existing method `utils.py::filter_lmp_locations` for this operation

rather than pass field name as a parameter I would just do the rename to "Location" before calling since that is the standard name across grid status

gridstatus/ercot.py
```diff
@@ -491,6 +845,73 @@ def _handle_html_data(self, df, columns):
...
+            raise ValueError(f"Invalid location_type: {location_type}")
+
+    def _get_settlement_point_mapping(self, verbose=False):
+        """Get dataframe whose columns can help us filter out values"""
+
+        report_type_id = SETTLEMENT_POINTS_LIST_AND_ELECTRICAL_BUSES_MAPPING_RTID
+        url = f"https://www.ercot.com/misapp/servlets/IceDocListJsonWS?reportTypeId={report_type_id}"
+        if verbose:
+            print(f"Fetching document {url}", file=sys.stderr)
+        docs = self._get_json(url)["ListDocsByRptTypeRes"]["DocumentList"]
```
kmax12, line 885: any reason you can use `_get_document` for this?

gridstatus/tests/test_ercot.py
```diff
@@ -86,3 +87,107 @@ def test_ercot_get_fuel_mix():
...
+        "Market",
+        "Location Type",
+        "LMP",
+    ]
+    df = iso.get_lmp(
+        date="today",
+        market=Markets.REAL_TIME_15_MIN,
+        location_type="zone",
+    )
+    assert df.shape[0] >= 0
```
kmax12, line 107: the validation is so similar across these these tests i'd just create a `check_ercot_spp(df, market_type, location_type)` method that contains all the assertions and then use that in each test

at the top of `tests/test_isos.py` you can see how i used this approach elsewhere

gridstatus/ercot.py
```diff
@@ -432,6 +501,253 @@ def get_interconnection_queue(self, verbose=False):
...
+            elif utils.is_today(date):
+                return self._get_lmp_dam_today(locations, location_type, verbose)
+            else:
+                unsupported = True
+        else:
+            unsupported = True
+
+        if unsupported:
+            raise NotSupported(
+                f"Market {market} and/or date {date} are not supported for ERCOT",
```
kmax12, line 559: should the error just say market isn't supported? I think you check for the date at the top of the method

gridstatus/ercot.py
```diff
@@ -432,6 +501,253 @@ def get_interconnection_queue(self, verbose=False):
...
+        elif market == Markets.DAY_AHEAD_HOURLY:
+            if date == "latest":
+                return self._get_lmp_dam_latest(locations, location_type, verbose)
+            elif utils.is_today(date):
+                return self._get_lmp_dam_today(locations, location_type, verbose)
+            else:
+                unsupported = True
+        else:
+            unsupported = True
+
```
kmax12, line 556: I think some form of the following logic is in each of the `_get_spp` methods. can we just include it in this function to avoid repetition and increase maintainability?

```python
df = self._filter_by_locations(df, "SettlementPointName", locations)

df = df.rename(
    columns={
        "SettlementPointPrice": "LMP",
        "SettlementPointName": "Location",
    },
)

df = df[
    [
        "Location",
        "Time",
        "Market",
        "Location Type",
        "LMP",
    ]
]

df = df.reset_index(drop=True)```

File: gridstatus/ercot.py [hunk 10]
```diff
@@ -432,6 +501,253 @@ def get_interconnection_queue(self, verbose=False):
...
+        return df
+
+    def _get_lmp_dam_today(
+        self,
+        locations: list = None,
+        location_type: str = None,
+        verbose=False,
+    ):
+        """Get day-ahead hourly Market LMP data for ERCOT"""
+        today_date = pd.Timestamp(pd.Timestamp.now(tz=self.default_timezone).date())
```
kmax12, line 585: same comment about `normalize`

kmax12 (changes requested):
Wow! This is the most significant outside contribution gridstatus has had so far. really appreciate it! 

I have several review comments here, but one more to add. Feel free to push back on of my comments or file issues you feel should be skipped and completed in another PR by you/someone else

Can we move the usage from the PR description to an example notebook file in the docs? The notebook can live in `docs/Examples/ercot`. I need to document this process in a contributing guide, but the steps are 

1. Add a `.ipynb` file with cell outputs saved to appropriate folder in `docs/Examples`. The docs build process will not run the code to generate output so must save it as you want it to be display.
2. Add a link to the notebook in `docs/Examples/<folder>/index.md`
3. Confirm the notebook made it in the docs by running `make html` from the root of the `docs/`. You can also look at the docs as generated by the CI by clicking on the github action from the PR
4. If you are submitting a change with a PR, I recommend you link to the usage notebook on read the docs in the PR description

Screenshot 2022-12-20 at 8 31 18 PM

gridstatus/ercot.py
```diff
@@ -469,6 +787,42 @@ def _get_document(
         url = f"https://www.ercot.com/misdownload/servlets/mirDownload?doclookupId={doc[1]}"
         return url, doc[0]
 
+    def _get_documents(
```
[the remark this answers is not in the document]
jcsalterego: I thought about this a bit, and they're distinct since `_get_document` is opinionated about picking the latest AND raising an exception if there's no match. Would you want to preserve that behavior, even if it reuses `_get_documents`?

jcsalterego (commented):
Thanks for the review. It'll probably take a bit to make these changes.

gridstatus/ercot.py
jcsalterego: I think we need to preserve `tz=self.default_timezone` because if this code is running in UTC but local time (US Central) is still in the previous day, then we want the local date, right?

So then this becomes:

```python
pd.Timestamp.now(tz=self.default_timezone).normalize()
```

kmax12: yep, that looks correct

gridstatus/ercot.py
kmax12: can you use `_get_document` to get the url, but then still inline the zip file extraction? 

basically, this logic seems to already be in `_get_document`

```python
report_type_id = SETTLEMENT_POINTS_LIST_AND_ELECTRICAL_BUSES_MAPPING_RTID
url = f"https://www.ercot.com/misapp/servlets/IceDocListJsonWS?reportTypeId={report_type_id}"
if verbose:
    print(f"Fetching document {url}", file=sys.stderr)
docs = self._get_json(url)["ListDocsByRptTypeRes"]["DocumentList"]
latest_doc = sorted(
    docs, key=lambda x: x["Document"]["PublishDate"])[-1]
doc_id = latest_doc["Document"]["DocID"]
doc_url = f"https://www.ercot.com/misdownload/servlets/mirDownload?doclookupId={doc_id}"
```

gridstatus/ercot.py
```diff
@@ -432,6 +501,253 @@ def get_interconnection_queue(self, verbose=False):
...
+        location_type: str = None,
+        verbose=False,
+    ):
+        """Get LMP data for ERCOT
+
+        Supported Markets: REAL_TIME_15_MIN, DAY_AHEAD_HOURLY
+
+        Supported Location Types: "zone", "hub", "node"
+        """
+        if not (date == "latest" or utils.is_today(date)):
```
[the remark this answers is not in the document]
kmax12: i think the existing reports you are using have some historical data though. for example, the DAM SPP report seems to have last 30 days. 

I'm looking here: https://www.ercot.com/mp/data-products/data-product-details?id=NP4-190-CD

gridstatus/ercot.py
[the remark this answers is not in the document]
kmax12: i think the approach to preserve that behavior while reusing `_get_documents` is good

gridstatus/ercot.py
jcsalterego: Yep, I see what you mean now. I was misreading `_get_document` and thinking it wouldn't match anything without a `date`, but I was wrong, and there will be a new `_get_document` anyhow.

gridstatus/tests/test_ercot.py
```diff
@@ -2,11 +2,36 @@
 import pytest
 
 import gridstatus
+from gridstatus import Markets
+
+
+def check_ercot_spp(df, market, location_type):
```
jcsalterego, line 8: Regarding your earlier comment, we can move these into `test_isos.py` once other ISOs get `get_spp`.

gridstatus/ercot.py
```diff
@@ -432,42 +509,257 @@ def get_interconnection_queue(self, verbose=False):
...
+        location_type: str = None,
+        verbose=False,
+    ):
+        """Get SPP data for ERCOT
+
+        Supported Markets: REAL_TIME_15_MIN, DAY_AHEAD_HOURLY
+
+        Supported Location Types: "zone", "hub", "node"
+        """
+        if locations is None:
```
kmax12, line 527: would it make more sense to just make "ALL" the default value in the function signature? same comment for `LOCATION_TYPE_ZONE` below

gridstatus/ercot.py
```diff
@@ -432,42 +509,257 @@ def get_interconnection_queue(self, verbose=False):
 
         return queue
 
+    def get_spp(
```
kmax12, line 512: I believe this now handles historical data, if that's the case, we can add the `@support_date_range(frequency="1D")` decorator.

gridstatus/ercot.py
```diff
@@ -432,42 +509,257 @@ def get_interconnection_queue(self, verbose=False):
...
+
+        if market == Markets.REAL_TIME_15_MIN:
+            return self._get_spp_rtm15(
+                date,
+                locations,
+                location_type,
+                verbose,
+            )
+        elif market == Markets.DAY_AHEAD_HOURLY:
+            return self._get_spp_dam(date, locations, location_type, verbose)
```
kmax12, line 544: this is small, but I think you should do the call to `Ercot._finalize_spp_df` here since both dam and rtm call it. 

what I am thinking

```python
if market == Markets.REAL_TIME_15_MIN:
    df = self._get_spp_rtm15(
        date,
        locations,
        location_type,
        verbose,
    )
elif market == Markets.DAY_AHEAD_HOURLY:
    df = self._get_spp_dam(date, locations, location_type, verbose)

else:
    raise NotSupported(
        f"Market {market} not supported for ERCOT",
    )
    
return Ercot._finalize_spp_df(df, "SettlementPoint", locations)
```

gridstatus/ercot.py
```diff
@@ -432,42 +509,257 @@ def get_interconnection_queue(self, verbose=False):
...
+
     def _get_document(
         self,
         report_type_id,
         date=None,
         constructed_name_contains=None,
         verbose=False,
-    ):
-        """Get document for a given report type id and date. If multiple document published return the latest"""
+    ) -> Document:
```
kmax12, line 703: this function looks great!

gridstatus/ercot.py
```diff
@@ -432,42 +509,257 @@ def get_interconnection_queue(self, verbose=False):
...
+    def _get_spp_dam(
+        self,
+        date,
+        locations: list = None,
+        location_type: str = None,
+        verbose=False,
+    ):
+        """Get day-ahead hourly Market SPP data for ERCOT"""
+        today = pd.Timestamp.now(tz=self.default_timezone).normalize()
+        if date == "latest":
```
kmax12, line 558: just realized, there is probably no practical need to support `latest` for the DAM. That is because the entire day of prices is priced and released simultaneously. That means "latest" in the day ahead market will always mean the last hour of the day. This is different than the RTM which clears tick by tick. 

can we just remove support for latest? also remove from notebook example

jcsalterego: DAM is definitely interesting. What happens if we pass in `latest` for DAM then? Should it raise a ValueError or just return `today`?

Also if I wanted to get tomorrow's data today, I would pass in tomorrow's date explicitly?

gridstatus/ercot.py
jcsalterego: Sounds good

gridstatus/ercot.py
jcsalterego: Yes, will do

gridstatus/ercot.py
jcsalterego: location_type defaulting to ZONE was following NYISO. I think the code currently expects exactly one location type - so just to confirm, you're looking to support returning data showing all location types?

gridstatus/ercot.py
jcsalterego: Oh right, I couldn't do that because the settlement point fields are named differently, because of course:

```
ercot.py
585:        return Ercot._finalize_spp_df(df, "SettlementPoint", locations)
695:        return Ercot._finalize_spp_df(df, "SettlementPointName", locations)
```

I could do something like this:
```python
if market == Markets.REAL_TIME_15_MIN:
    df = self._get_spp_rtm15(
        date,
        locations,
        location_type,
        verbose,
    )
    settlement_point_field = "SettlementPoint"
elif market == Markets.DAY_AHEAD_HOURLY:
    df = self._get_spp_dam(date, locations, location_type, verbose)
    settlement_point_field = "SettlementPointName"
else:
    raise NotSupported(
        f"Market {market} not supported for ERCOT",
    )
    
return Ercot._finalize_spp_df(df, settlement_point_field, locations)
```

Does that work?

gridstatus/ercot.py
kmax12: > DAM is definitely interesting. What happens if we pass in latest for DAM then? Should it raise a ValueError or just return today?

i think it should error for `latest`. the error message can say to use "today" instead. 

> DAM is definitely interesting. What happens if we pass in latest for DAM then? Should it raise a ValueError or just return today?
> 
> Also if I wanted to get tomorrow's data today, I would pass in tomorrow's date explicitly?

i need to do an audit across the ISOs of how dates are used to refer to day ahead prices and make sure grid status is consistent. I believe the most common approach is for the date to refer to when the energy is delivered. but because it is the day ahead market, the results are published the day before. 

I think this is what you are achieving with `publish_date = publish_date - pd.Timedelta("1D")`. 

So, the short answer is yes. You pass in tomorrows date.

gridstatus/ercot.py
kmax12: sorry, I meant that we could just do this instead of these conditionals checking for `None`

```python
def get_spp(
        self,
        date,
        end=None,
        market: str = None,
        locations: list = "ALL",
        location_type: str = "ZONE",
        verbose=False,
    ):
    ...
```

gridstatus/ercot.py
kmax12: ahh, i see. i prefer to have the `finalize_spp_df` call in this function since it makes it more clear its common to all the markets. i think your approach works.

gridstatus/ercot.py
jcsalterego: Oh, whoops. Misread that by a mile.

jcsalterego:
ISO APIs are getting wonky this time of day (see 27 minute test runtime), but I think the latest changes are good to go.

kmax12 (commented):
All the tests are passing locally for me, so I agree it's good.

Can you merge in `main` and also add a brief entry in `CHANGELOG.md`? After those two things, we should be good to go.

jcsalterego:
> Can you merge in `main` and also add a brief entry in `CHANGELOG.md`? After those two things, we should be good to go.

Done and done

kmax12 (approved):
Excellent job! This is great work. I really appreciate the contribution.

PR #98, opened 2022-12-18
Update LMP Pricing Data Table Format

jcsalterego:
## Description

Present the LMP Pricing Data Availability Table in the same green checkmark and red X fashion as the methods availability table, plus an extra column to indicate the main method for accessing LMP/SPP data.

## Before
image

## After
Screen Shot 2022-12-23 at 8 59 48 AM


jcsalterego:
Unrelated test failure with ```self = <json.decoder.JSONDecoder object at 0x7f95a30a63d0>
[493](https://github.com/kmax12/gridstatus/actions/runs/3726071149/jobs/6319276936#step:5:494)
s = 'ï»¿<!DOCTYPE html>\r\n<html>\r\n<head>\r\n    <title>An internal server error has occurred. Please check the URL and ...coration:none;font-size:14px;">Go Back to the Homepage</a>\r\n        </div>\r\n    </div>\r\n</body>\t\r\n</html>\r\n'
[494](https://github.com/kmax12/gridstatus/actions/runs/3726071149/jobs/6319276936#step:5:495)```

kmax12:
wanted to get #97 merged in before returning to this. This is good but we just need to make sure Ercot still works after merging in `main` and makes sense to just mention somewhere on this page (maybe below table?) that Ercot uses `get_spp` instead of `get_lmp`

jcsalterego:
> This is good but we just need to make sure Ercot still works after merging in `main`. Also i think it makes sense to just mention somewhere on this page (maybe below table?) that Ercot uses `get_spp` instead of `get_lmp`

Do you foresee the other ISOs getting the `get_spp` method, thus making this a separate table (or more columns)?

kmax12:
no, I think just ercot has spp. everyone else uses `lmp`

jcsalterego:
OK. I'll review the detection code and add a note right after the table.

jcsalterego:
The LMP doc could use a fresh coat of paint I think.

> We are currently adding Locational Marginal Price (LMP). Even though each BA offers different markets, but you can query them with a standardized API

What does BA stand for here?

jcsalterego:
Ready for review

kmax12:
> What does BA stand for here?

Stands for balancing authority. Basically, the organization responsible for balancing supply/demand on the grid. I'm fairly confident it is save to say that all the ISOs and RTOs are BAs. However, there are definitely more BAs that aren't ISOs or RTOs. 

Not sure it's usage is necessary/helpful here

jcsalterego:
> Stands for balancing authority.

Ah, OK - thanks.

kmax12 (approved):
Looks great. Thank you!

PR #105, opened 2022-12-27
SPP: Recent LMP (RTM & DAM)

jcsalterego:
Adds Southwest Power Pool (SPP) support for Locational Marginal Prices (LMP) supporting the following combinations:

* date = latest, today
* markets = Real Time (5 minutes), Day Ahead (Hourly)
* location types = hub, interface

with the exception of fetching latest DAM resulting in a ValueError.

Additional notes:

* Historical is within reach but the data is compacted into larger time periods the further back we go in time, so the best solution is likely a smart planner which can list files and figure out the most efficient path to downloading the right data set
* Data is provided by Settlement Locations and Physical Nodes (PNode), but I'm unsure how they should be represented in `location_type`
* We rely on MapGIS feature data endpoints to get a definitive list of Hubs and Interfaces and use them to filter the CSV data
  * Hub: https://pricecontourmap.spp.org/arcgis/rest/services/MarketMaps/RTBM_FeatureData/MapServer/1/
  * Interface: https://pricecontourmap.spp.org/arcgis/rest/services/MarketMaps/RTBM_FeatureData/MapServer/2/


gridstatus/spp.py
```diff
@@ -272,6 +297,311 @@ def get_interconnection_queue(self, verbose=False):
...
+            )
+            paths = files_df["path"].tolist()
+        if verbose:
+            print(f"Found {len(paths)} files for {date}", file=sys.stderr)
+        return paths
+
+    def _fetch_rtbm_lmp_by_location(self, date, verbose=False):
+        all_dfs = []
+        paths = self._fs_get_rtbm_lmp_by_location_paths(date, verbose=verbose)
+        for path in paths:
```
kmax12, line 515: what do you think about adding a `tqdm` progress bar here? I do that in other places where there is a lot of files that need to de downloaded leading to longer than normal runtimes. Even though there is no way for the user to turn it off, i think it is better to have stuff printed to stdout than to wait at screen that isn't updating

```python
for path in tqdm.tqdm(paths):
    ...
```

example: https://github.com/kmax12/gridstatus/blob/d47ff339930d35c5fd73ef6aea80039eb47ad751/gridstatus/pjm.py#L406

kmax12 (changes requested):
This is a beautiful scraper. Thank you so much for continuing to contribute :) 

>Historical is within reach but the data is compacted into larger time periods the further back we go in time, so the best solution is likely a smart planner which can list files and figure out the most efficient path to downloading the right data set

Got it. Do you think you could put up a separate github issue just documenting your current thoughts? If you want to pick it up yourself in near future, no need to be too verbose.

>Data is provided by Settlement Locations and Physical Nodes (PNode), but I'm unsure how they should be represented in location_type 

page 209 of this [file](https://www.spp.org/documents/34096/integrated%20marketplace%20protocols%2035b%201.pdf) describes PNodes, Settlement Locations, and Hub.

My take away is that we should just add a generic "Settlement Location" location_type and have it return all the data. If possible would still be good to mark the hubs and interfaces as such, but everything else would just be a Settlement Location. Open to other options.

kmax12:
one more thing, can you make sure to add an entry to `CONTRIBUTING.md`?

gridstatus/spp.py
```diff
@@ -272,6 +297,311 @@ def get_interconnection_queue(self, verbose=False):
...
+                date,
+                end,
+                market,
+                locations,
+                location_type,
+                verbose,
+            )
+        else:
+            raise NotSupported(
+                f"Location type {location_type} is not supported for SPP",
```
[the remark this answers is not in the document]
jcsalterego: Yes

gridstatus/spp.py
```diff
@@ -272,6 +297,311 @@ def get_interconnection_queue(self, verbose=False):
...
+        Returns:
+            pd.DataFrame of features
+        """
+        args = {
+            "f": "json",
+            "where": "OBJECTID IS NOT NULL",
+            "returnGeometry": "false",
+            "outFields": "*",
+        }
+        url = utils.url_with_query_args(base_url, args)
```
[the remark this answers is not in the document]
jcsalterego: FWIW, I ripped out `url_with_query_args` but am still using `requests` for the CSV fetches.

gridstatus/utils.py
```diff
@@ -262,3 +263,10 @@ def load_folder(path, time_zone=None, verbose=True):
     # todo make sure rows are sorted by time
 
     return data
+
+
+def url_with_query_args(url: str, qsargs: dict) -> str:
+    """Return a URL with query string arguments"""
+    qs = "&".join("{k}={v}".format(k=k, v=quote(v)) for k, v in qsargs.items())
```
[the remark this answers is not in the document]
jcsalterego: Ah, thanks.

gridstatus/spp.py
jcsalterego: Will add

jcsalterego:
Minutia: I went with an underscore for the `SETTLEMENT_LOCATION` location_type which is consistent with PJM. Happy to change it to `SETTLEMENT LOCATION` if you have strong feelings.

gridstatus/spp.py
```diff
@@ -272,6 +302,348 @@ def get_interconnection_queue(self, verbose=False):
...
+            df (DataFrame): DataFrame with SPP data
+            market (str): Market
+            locations (list): List of locations to filter by
+            location_type (str): Location type
+            verbose (bool): Verbose output
+        """
+
+        df["Market"] = market.value
+
+        if location_type == LOCATION_TYPE_SETTLEMENT_LOCATION:
```
kmax12, line 437: I like this implementation!

gridstatus/spp.py
```diff
@@ -272,6 +302,348 @@ def get_interconnection_queue(self, verbose=False):
...
+            )
+        else:
+            # filter
+            location_list = self._get_location_list(location_type, verbose=verbose)
+            df["Location Type"] = SPP._get_location_type_name(location_type)
+            df = df[df["Location"].isin(location_list)]
+
+        df = df.rename(
+            columns={
+                "LMP": "LMP",  # for posterity
```
kmax12, line 470: 😂 

i do this too when doing renames to make it easier to understand code

gridstatus/spp.py
```diff
@@ -272,6 +302,348 @@ def get_interconnection_queue(self, verbose=False):
...
+            df = pd.DataFrame(list_results.json())
+            return df
+        else:
+            return pd.DataFrame()
+
+    def _file_browser_download_url(self, fs_name, params=None):
+        qs = "?" + urlencode(params) if params else ""
+        return f"{FILE_BROWSER_API_URL}download/{fs_name}{qs}"
+
+    def _log_url(self, message, url, params={}):
```
kmax12, line 637: is this function used? seems helpful but lets remove if not used

kmax12 (changes requested):
this is looking ready to merge. just one last small comment. perhaps also worth merging in main since there were some small linting changes

gridstatus/spp.py
jcsalterego: Good catch

kmax12 (approved):
looks good to me. 

one failing SPP test is related to #109 

The other is an unrelated ISONE test

PR #115, opened 2022-12-30
Add Github Action cache to speed up build docs workflow

### LINKED ISSUE #113: Use Github Action Caching to Speed Up Tests
More info: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

gsheni:
- Fixes #113

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -17,14 +17,16 @@ jobs:
       - uses: actions/cache@v3
         id: cache
         with:
-          path: ~/.cache/pip
-          key: py-${{ matrix.python-version }}-tests-${{ hashFiles('pyproject.toml') }}-v01
-          restore-keys: |
-            py-${{ matrix.python-version }}-tests-
-      - name: Install dependencies (if cache has not changed)
+          path: ${{ env.pythonLocation }}
+          key: py-${{ matrix.python-version }}-tests-${{ env.pythonLocation }}-${{ hashFiles('pyproject.toml') }}-v01
+      - name: Install core deps (if cache has changed)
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -12,7 +12,15 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+          cache-dependency-path: "pyproject.toml"
+      - uses: actions/cache@v3
+        id: cache
+        with:
+          path: ~/.cache/pip
+          key: python-deps-${{ matrix.python-version }}-${{ hashFiles('pyproject.toml') }}-tests-v01
       - name: Install dependencies
```
gsheni, line 22:
```suggestion
      - name: Install dependencies (if cache has not changed)
```

kmax12:
can you tell if it is actually speeding up the tests? it's a bit more complexity than I thought, so probably only worth it if the tests are speeding up significantly

gsheni:
#### The Build Docs workflow did speed up.
- No Cache ~ 2m 38s (https://github.com/kmax12/gridstatus/actions/runs/3809303349/jobs/6480574975)
- Cache ~1m 56s (https://github.com/kmax12/gridstatus/actions/runs/3813177529/jobs/6486773457)

gsheni:
I was unable to get the cache working for the unit tests. The cache will only be saved if the workflow succeed, but the unit tests are currently failing. 
I tried some tricks (disable fail-fast, silence pytest output) with no luck. **My guess is you would save ~23s per each unit test run per python version (~23 x 5 = 115)**

Issue #118, opened 2023-01-02
Refactor tests


kmax12:
Let's refactor `test_isos.py`. Below are some initial thoughts

## Goal

1. Make it easy to run only tests for a specific ISO
3. Make it easy to identify which ISOs have failing tests
4. Keep shared testing code in a central location
5. Still allow iso specific tests to be written

## Proposed approach

Outlined in this branch: https://github.com/kmax12/gridstatus/blob/5ad2795d397c400d20d0a4b71023eab01ef638ff/gridstatus/tests/test_subclass.py

## Open questions

* How do we specify tests that are shared by some, but not all ISO. e.g `test_get_latest_status`

jcsalterego:
👀

WillKoehrsen:
I think this has been completed. Running tests for a specific iso can be done with `pytest gridstatus/tests/test_{iso}.py`. Shared testing functionality is in `base_test_iso.py`.

PR #124, opened 2023-01-03
Refactor Test Part I

jcsalterego:
Addresses the beginning of #118:

* moves parametrized tests from `test_isos.py` into `class BaseTestISO`
* standardizes on `def test_{method}_*` for ease of use, e.g. `test_get_historical_lmp` becomes `test_get_lmp_historical`
* creates subclasses, e.g. `class TestErcot(BaseTestISO)` in parallel files, e.g. `gridstatus/tests/test_ercot2.py`
* adds a `@with_markets` decorator which wraps `pytest.mark.parametrize("market")`

Once we finalize the approach, we can either:

* merge this PR, possibly creating double testing,
* OR continue porting tests into the subclasses, and then collapsing the parallel files into the final files, e.g. `gridstatus/tests/test_ercot2.py -> gridstatus/tests/test_ercot.py`

### DIFF
--- a/gridstatus/tests/test_ercot2.py
+++ b/gridstatus/tests/test_ercot2.py
@@ -17,3 +17,6 @@ class TestErcot(BaseTestISO):
     def test_get_load_forecast_historical(self):
         with pytest.raises(NotSupported):
             super().test_get_load_forecast_historical()
+
+    def test_get_load_forecast_historical_with_date_range(self):
+        pass
--- a/gridstatus/tests/test_ercot2.py
+++ b/gridstatus/tests/test_ercot2.py
@@ -7,6 +7,7 @@ from gridstatus.tests.base_test_iso import BaseTestISO
 class TestErcot(BaseTestISO):
     iso = Ercot()
 
+    @pytest.mark.skip
     def test_get_fuel_mix_date_or_start(self):
         pass
 
@@ -14,9 +15,11 @@ class TestErcot(BaseTestISO):
         with pytest.raises(NotSupported):
             super().test_get_fuel_mix_historical()
 
+    @pytest.mark.skip
     def test_get_fuel_mix_historical_with_date_range(self):
         pass
 
+    @pytest.mark.skip
     def test_get_lmp_historical(self, markets=None):
         pass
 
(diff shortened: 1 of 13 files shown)


gridstatus/tests/test_ercot2.py
```diff
@@ -0,0 +1,48 @@
...
+from gridstatus import Ercot, NotSupported
+from gridstatus.tests.base_test_iso import BaseTestISO
+
+
+class TestErcot(BaseTestISO):
+    iso = Ercot()
+
+    """get_fuel_mix"""
+
+    @pytest.mark.skip
```
jcsalterego, line 12: This is an extraneous decorator, but I thought it was a good way of noting the ISO is opting-out of a test in addition to the `pass`.

The other options I see are adding a decorator to the BaseTestISO method which includes or excludes certain subclasses, but at the point it seems like six of one, half of a dozen of the other.

gridstatus/tests/test_isos.py
```diff
@@ -1,52 +1,13 @@
...
-    )
-
-
-def check_status(df):
-    assert set(df.columns) == set(
-        ["Time", "Status", "Notes"],
-    )
-
-
 def check_is_datetime_type(series):
```
kmax12, line 11: i think this was moved to `base_test_iso.py` and is no longer used here

gridstatus/tests/test_isos.py
```diff
@@ -1,52 +1,13 @@
 import pandas as pd
```
kmax12, line 1: I think we can rename this file now that is just the misc tests that remain. perhaps `test_gridstatus.py`?

perhaps also just leave a comment that ISO general tests go in `base_test_iso` and ISO specific test go in their respective files, so it clear what goes here.

gridstatus/tests/test_ercot2.py
kmax12: I like this approach. i think it's most scalable to have the subclass handle their own opting out. seems better if the base test class doesn't have to be aware of all it's subclasses 

one thought I have to make it even more clear is add a reason to the marker like this `@pytest.mark.skip(reason="Not Applicable")`.

kmax12 (changes requested):
Nice. This must have been some tedious work. Thanks!

I dont see why we can't merge before completing the second part. In fact, the longer this is unmerged, the more git conflicts it'll lead to if new tests are added.

Just a few comments

PR #129, opened 2023-01-05
Update GitHub action used to publish PyPI package

gsheni:
- Use the official GitHub action to release packages on PyPi
- @kmax12 You will need to create an PyPI API Token and add it as an GitHub env secret (`PYPI_API_TOKEN`)


kmax12:
What changes to the release process does this introduce?

For example, it looks like it ill no longer check for the release tag to match the package version.

How can we test to make sure it works before switching over? 

Also, what benefits does switching have? I don't currently have any issue with the existing github action

PR #130, opened 2023-01-05
Refactoring Tests: Part II

jcsalterego:
Part II of #118 

The resulting changeset is pretty difficult to read, but the process was such:

* functions from old `test_*.py` files become methods in their counterpart `test_*2.py` test classes
* methods are renamed to follow `test_{iso_method}_{description}` where possible, e.g. `get_historical_lmp_with_date_range` becomes `get_lmp_historical_with_date_range`
* local utility methods such as `assert_*` and `check_*` become `_assert_*` and `_check_*`
* each test suite is organized into sections by method names, with accompanying short docstring, e.g. `"""get_lmp"""`
* finally, `test_*2.py` replaces `test_*.py`

### Test suite comparisons

[main
](https://github.com/kmax12/gridstatus/commit/e1d24a88973384a53e9589635377274fe5c4f06b)

```
pytest gridstatus/tests  --collect-only | tail -n 1
========================= 243 tests collected in 0.03s =========================
```

This branch:
```
pytest gridstatus/tests  --collect-only | tail -n 1
========================= 238 tests collected in 0.03s =========================
```

### Per commit method additions/subtractions

e754306ff6cd5d2b4abe49260cef0506c6ab4ac1 Reorganize get_interconnection_queue tests
```diff
-def check_queue(queue):
-def get_interconnection_queues():
-def test_get_interconnection_queue(iso):
+def test_get_interconnection_queue(self):
```

435c94f0fd2bcb82b413ff96acce3330187e7408 BaseTestISO reorg
```diff
+def test_get_load_forecast_historical(self):
-def test_get_load_forecast_historical(self):
+def test_get_load_forecast_historical_with_date_range(self):
-def test_get_load_forecast_historical_with_date_range(self):
+def test_get_load_forecast_today(self):
-def test_get_load_forecast_today(self):
```

89d6fcd3ecf546d2aabe437658e2fbcc2c7124df Consolidation: test_isone.py
```diff
-def test_get_fuel_mix(date):
+def test_get_fuel_mix(self, date):
+def test_get_fuel_mix_nov_7_2022(self):
-def test_get_lmp(date, market):
+def test_get_lmp(self, date, market):
+def test_get_lmp_historical(self, market):
-def test_get_lmp_historical(self, market):
+def test_get_lmp_latest(self, market):
-def test_get_lmp_latest(self, market):
+def test_get_lmp_today(self, market):
-def test_get_lmp_today(self, market):
-def test_get_load(date):
+def test_get_load(self, date):
[...]

PR #131, opened 2023-01-06
Remove Redundant Tests

kmax12:
Remove redundant tests that became clear from #130. 

Move the extra checks to the base function so all ISO benefit


jcsalterego (commented):
Sweet

PR #133, opened 2023-01-08
Enable PJM Realtime Markets

kmax12:
Enables PJM realtime markets for today and latest, even though gridstatus cannot current access the data since PJM publishes it with delay.


jcsalterego (approved):
:tada:

PR #135, opened 2023-01-12
@lmp_config Decorator

jcsalterego:
Quality of life improvement: declarative market and date (`latest`, `today`, `historical`) support for `get_lmp`/`get_spp`.

This will enforce args/kwargs presence of `date` (or `start`, from `@support_date_range`) and `market`.

`lmp_config.supports` also gives us an imperative way to check the config:

```python
import gridstatus
from gridstatus import Markets
from gridstatus.lmp_config import lmp_config

if lmp_config.supports(
    gridstatus.CAISO.get_lmp, Markets.REAL_TIME_5_MIN, "latest"):
    print("CAISO supports REAL_TIME_5_MIN")
```

The `make_lmp_availability_df` and table method can/should probably be rewritten to provide a more granular breakdown of what's all supported.

### Considerations

* Could live in `decorators.py`
* Could be named something else... `market_date_config`
* Or, it could be specialized for `get_lmp`/`get_spp` and do other things later such as:
  * `location_type` filtering
  * automatic wrapping `date=latest` with `_latest_lmp_from_today`

### Unknown

* Don't know how it interacts with `@support_date_range`, other than having to add `start=` support.

### DIFF
--- a/gridstatus/lmp_config.py
+++ b/gridstatus/lmp_config.py
@@ -1,0 +1,152 @@
+import functools
+
+import pandas as pd
+
+from gridstatus.base import ISOBase, Markets, NotSupported
+from gridstatus.utils import _handle_date, is_today
+
+
+class lmp_config:
+    def __init__(self, supports, tz=None):
+        self.supports = supports
+        self.tz = tz
+
+    def __call__(self, func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            if self._is_class_method(args, kwargs):
+                instance = args[0]
+                func_args = args[1:]
+                func_kwargs = kwargs
+                func_sig = self._parse_signature(
+                    instance,
+                    func_args,
+                    func_kwargs,
+                    tz=self.tz,
+                )
+                return self._class_method_wrapper(instance, func, func_sig)
+            else:
+                func_sig = self._parse_signature(None, args, kwargs, tz=self.tz)
+                return self._function_wrapper(func, func_sig)
+
+        return wrapper
+
+    @staticmethod
+    def _is_class_method(args, kwargs):
+        # args[0] must be "self" and an instance of ISOBase
+        return len(args) > 0 and isinstance(args[0], ISOBase)
+
+    @staticmethod
+    def _validate_market_arg_signature(market):
+        return isinstance(market, str) or isinstance(market, Markets)
+
+    @staticmethod
+    def _parse_date(date, tz):
+        if date in (
+            "latest",
+            "today",
+        ):
+            return _handle_date("today", tz=tz)
+        elif isinstance(date, pd.Timestamp):
+            final_date = _handle_date(date, tz=tz)
+            if not isinstance(final_date, pd.Timestamp):
+                raise ValueError(f"Cannot parse date {repr(date)}")
+            return final_date
+        else:
+            raise ValueError("date must be string or pd.Timestamp")
+
+    @staticmethod
+    def _class_method_wrapper(instance, func, func_sig):
+        instance_args = tuple([instance] + list(func_sig["args"]))
+        instance_kwargs = func_sig["kwargs"]
+        return func(*instance_args, **instance_kwargs)
+
+    @staticmethod
+    def _function_wrapper(func, func_sig):
+        return func(*func_sig["args"], **func_sig["kwargs"])
+
+    @staticmethod
+    def _parse_market(market):
+        return Markets(market)
+
+    def _parse_signature(self, instance, args, kwargs, tz):
+        date = None
+        market = None
+
+        if tz is None:
+            if instance is not None:
+                if not hasattr(instance, "default_timezone"):
+                    raise ValueError(
+                        "ISO does not have default_timezone set; cannot determine tz",
+                    )
+                else:
+                    tz = instance.default_timezone
+        if tz is None:
+            raise ValueError("Must set tz= arg or ISO default timezone")
+
+        args = list(args)
+
+        if len(args) == 0:
+            if "date" not in kwargs or "market" not in kwargs:
+                raise ValueError("date and market are required")
+            date = kwargs["date"]
+            market = kwargs["market"]
+        elif len(args) == 1:
+            if "date" in kwargs and "market" in kwargs:
+                date = kwargs["date"]
+                market = kwargs["market"]
+            elif "date" in kwargs:
+                date = kwargs["date"]
+                market = args[0]
+            elif "market" in kwargs:
+                date = args[0]
+                market = kwargs["market"]
+            else:
+                raise ValueError("Missing date or market")
+        else:
+            date = args[0]
+            market = args[1]
+
+        original_date_arg = date
+        date = self._parse_date(date, tz=tz)
+        market = self._parse_market(market)
+
+        self._check_support(original_date_arg, date, market, tz)
+
+        if len(args) == 0:
+            kwargs["date"] = date
+            kwargs["market"] = market
+        elif len(args) == 1:
+            if "date" in kwargs and "market" not in kwargs:
+                kwargs["date"] = date
+                args[0] = market
+            elif "date" not in kwargs and "market" in kwargs:
+                args[0] = date
+                kwargs["market"] = market
+        else:
+            args[0] = date
+            args[1] = market
+
+        return {
+            "args": tuple(args),
+            "kwargs": kwargs,
+        }
+
+    def _check_support(self, original_date_arg, date_arg, market_arg, tz):
+        if market_arg not in self.supports:
+            raise NotSupported(f"{market_arg} not supported")
+
+        if original_date_arg in (
+            "latest",
+            "today",
+        ):
+            supported = original_date_arg in self.supports[market_arg]
+        elif is_today(date_arg, tz):
+            supported = "today" in self.supports[market_arg]
+        else:
+            supported = "historical" in self.supports[market_arg]
+
+        if not supported:
+            raise NotSupported(
+                f"{market_arg} does not support {repr(original_date_arg)}",
+            )
--- a/gridstatus/lmp_config.py
+++ b/gridstatus/lmp_config.py
@@ -67,6 +67,7 @@ class lmp_config:
 
     def _parse_signature(self, instance, args, kwargs, tz):
         date = None
+        date_field = None
         market = None
 
         if tz is None:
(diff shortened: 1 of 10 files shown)


kmax12:
I really like this idea. I actually had been kicking around a similar idea for not only lmp, but all methods. Especially since it would make docs a lot easier.

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,211 @@
...
+
+    @staticmethod
+    def _function_wrapper(func, func_sig):
+        return func(*func_sig["args"], **func_sig["kwargs"])
+
+    @staticmethod
+    def _parse_market(market):
+        return Markets(market)
+
+    def _parse_signature(self, instance, args, kwargs, tz):
```
kmax12, line 81: I'm not sure how to simplify right now, but adding all this higher-level logic to parsing the signature seems potentially less maintainable and eventually more error prone than doing nothing right now. 

one approach might be to put some of this logic for standardizing date/start/end in a different decorator. there is some similar code doing this validation for support date range here: https://github.com/kmax12/gridstatus/blob/860565ae81ae888df5e3442bfe9c12a2cbac45f7/gridstatus/decorators.py#L38

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,211 @@
...
+            supported = "historical" in self.supports[market_arg]
+
+        if not supported:
+            raise NotSupported(
+                f"{market_arg} does not support {repr(original_date_arg)}",
+            )
+
+    # copied from utils.py to avoid circular import
+    @staticmethod
+    def __handle_date(date, tz=None):
```
kmax12, line 183: you can avoid a circular import putting the import inside the method call where you use it. that way you do have to copy code

kmax12 (changes requested):

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, supports, tz=None):
+        self.supports = supports
+        self.tz = tz
+
+    def __call__(self, func):
+        lmp_config.configs[func.__qualname__] = self.supports
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            if self._is_class_method(args, kwargs):
```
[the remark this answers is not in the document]
jcsalterego: Makes sense - I'll rip out the function/explicit tz support.

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,211 @@
...
+
+        return wrapper
+
+    @staticmethod
+    def _is_class_method(args, kwargs):
+        # args[0] must be "self" and an instance of ISOBase
+        return len(args) > 0 and isinstance(args[0], ISOBase)
+
+    @staticmethod
+    def _validate_market_arg_signature(market):
```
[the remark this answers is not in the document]
jcsalterego: Good call

gridstatus/lmp_config.py
jcsalterego: Yeah this whole method is... complex. I only saw the decorator stuff after putting up this PR - so I'll see how we can use args_dict here too.

gridstatus/lmp_config.py
jcsalterego: Ah, thanks

gridstatus/lmp_config.py
jcsalterego: OK I cleaned this up quite a bit: similar to `_get_args_dict`, I have `_get_fn_params` return OrderedDict `args`, and `kwargs` as-is. That allows me to recompose the original args after validating and transforming (if needed) `date` and `market`. The replacement logic is similarly simplified.

jcsalterego:
Now it turns out the `__code__` inspection stuff gets confused when the method is also wrapped with `@support_date_range`, so that's the next thing to tackle.

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,154 @@
...
+from gridstatus.base import ISOBase, Markets, NotSupported
+
+
+class lmp_config:
+
+    configs = {}
+
+    def __init__(self, supports, tz=None):
+        self.supports = supports
+        self.tz = tz
```
kmax12, line 16: do we need this `tz` parameter at all? not seeing it used anywhere.

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,154 @@
...
+        lmp_config.configs[func.__qualname__] = self.supports
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            bound_args = self._get_bound_args(func, args, kwargs)
+            if len(args) > 0 and isinstance(args[0], ISOBase):
+                bound_args = self._verify_bound_args(bound_args)
+                return self._class_method_wrapper(func, bound_args)
+            else:
+                raise ValueError("Must be class method on ISOBase")
```
kmax12, line 29: any way to do this validation check outside of `wrapper` so that it get checked before the method is actually called?

im searching and i can't find an easy way...

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,154 @@
...
+        """Verify date/start and market args/kwargs. Transform values and injects
+        them back into the original signature if needed.
+
+        Raises:
+            ValueError: If date/start or market are missing or invalid
+        """
+        arguments = bound_args.arguments
+        instance = arguments["self"]
+
+        date = self._get_first(arguments, ["date", "start"])
```
kmax12, line 72: not sure i see the need for the _get_first function. why can't it just be 

```python
if "date" in arguments:
   date = arguments["date"]
elif "start" in arguments:
   date = arguments["start"]
else:
   raise ValueError("date/start is required")
   
if "market" in arguments:
   market = arguments["market"]
else:
   raise ValueError("market is required")
```

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,154 @@
...
+            raise ValueError("market is required")
+
+        date_value = self._parse_date(date, tz=tz)
+        market_value = Markets(market)
+
+        self._check_support(date, date_value, market_value, tz)
+
+        if date != "latest" and date_value != date:
+            self._set_first(arguments, ["date", "start"], date_value)
+        if market != market_value:
```
kmax12, line 88: why even do this check? why not always just do `arguments["market"] = market_value`?

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,154 @@
...
+        elif market is None:
+            raise ValueError("market is required")
+
+        date_value = self._parse_date(date, tz=tz)
+        market_value = Markets(market)
+
+        self._check_support(date, date_value, market_value, tz)
+
+        if date != "latest" and date_value != date:
+            self._set_first(arguments, ["date", "start"], date_value)
```
kmax12, line 87: similar comment to `get_first` above. i think it's better to just be explicit with if/else statements here

gridstatus/miso.py
```diff
@@ -126,6 +127,12 @@ def _get_load_and_forecast_data(self, verbose=False):
         r = self._get_json(url, verbose=verbose)
         return r
 
+    @lmp_config(
+        supports={
+            Markets.REAL_TIME_5_MIN: ["latest"],
+            Markets.DAY_AHEAD_HOURLY: ["latest"],
+        },
+    )
```
kmax12, line 135: once we have this decorator, should we removed logic such as the following in all these files?
```
        if date != "latest":
            raise NotSupported()
```

gridstatus/lmp_config.py
```diff
@@ -0,0 +1,154 @@
+import datetime
+import functools
+import inspect
+
+import pandas as pd
+
+from gridstatus.base import ISOBase, Markets, NotSupported
+
+
+class lmp_config:
```
kmax12, line 10: we can do it in a separate PR, but I assume this decorator could also support validating `locations` and `location_type` parameters. maybe we can file an issue or leave a todo in the code?

kmax12 (changes requested):
nice, much simpler. a few comments

gridstatus/lmp_config.py
jcsalterego: Whoops

gridstatus/lmp_config.py
jcsalterego: Previously there have been debug statements here to verify the conversion of strings to the Market enum, hence the conditional. I'm fine with always setting it.

gridstatus/miso.py
jcsalterego: Yes

gridstatus/lmp_config.py
jcsalterego: #137

gridstatus/lmp_config.py
kmax12: i think we can leave it as-is for now. 

perhaps just leave a comment in the code  for posterity that this check has to happen at function runtime

jcsalterego: Done

kmax12 (commented):
This is looking pretty good to me. Does it make sense to up `make_lmp_availability_df` in this PR?

jcsalterego:
> This is looking pretty good to me. Does it make sense to up `make_lmp_availability_df` in this PR?

Yep, will do.

jcsalterego:
> again, this might be a separate PR, but can we get rid of the `@with_markets` decorator that happens in the tests and use this annotation instead?

`@with_markets` is a test decorator that is kinda shorthand for `pytest.mark.parametrize` - how do you see `@lmp_config` being used as a replacement?

kmax12:
At a high-level, I am wondering if it makes sense to explicitly list the markets to test for each ISO since we should just be testing the ones the ISO says they support @lmp_config.

In terms of implementing, perhaps you redefine `class with_markets` to just look at lmp_config to get the markets that should be tested? you might still need the decorator, but seems like it shouldn't have to repeat the markets listed in lmp_config

jcsalterego:
@kmax12 ah yes, I think that makes sense. A separate issue/PR? May work well with a larger decorator cleanup so other methods get the same safeguards and declarative supports.

jcsalterego:
#138 captures eliminating `@with_markets`.

`make_lmp_availability_df` has been updated:

### Current

image

### Previously

image

kmax12 (approved):
This looks good to me. Test errors seem unrelated, but going to run them again tomorrow morning to be sure and then I'll merge

Also, filled #139 

Thank you!

PR #136, opened 2023-01-16
New ISO: Alberta Electric System Operator (AESO)

jcsalterego:
This PR adds some support for Alberta Electric System Operator (AESO). They don't have a lot of data to fulfill the common gridstatus API from what I can tell.

The bulk of the work is parsing Current Supply & Demand: http://ets.aeso.ca/ets_web/ip/Market/Reports/CSDReportServlet?beginDate=&endDate=&contentType=html

If AESO isn't part of the roadmap, I can close the PR. Otherwise there is missing documentation, and possibly some historical data to be added.


WillKoehrsen (commented):
Is this still something we want to get merged? It looks pretty close.

PR #141, opened 2023-01-24
Improve today handling

kmax12:


gridstatus/utils.py
```diff
@@ -97,7 +97,7 @@ def make_availability_table():
 
 def _handle_date(date, tz=None):
     if date == "today":
-        date = pd.Timestamp.now(tz=tz)
+        date = pd.Timestamp.now(tz=tz).replace(hour=0, minute=0, second=0)
```
jcsalterego, line 100: I think `normalize` is used for exactly this. Otherwise you're left with milliseconds.

```suggestion
        date = pd.Timestamp.now(tz=tz).normalize()
```

```python
>>> pd.Timestamp.now().replace(hour=0, minute=0, second=0)
Timestamp('2023-01-24 00:00:00.237088')
```

kmax12: ahh, you're right. updated

Issue #146, opened 2023-02-02
SPP Day Ahead Market returning tomorrow's data for today


kmax12:
I am running the following code at 530pm CST on Feb 2. The expected behavior is to return the day ahead market for Feb 2, but it is actually giving data from Feb 3

```python
import gridstatus
iso = gridstatus.SPP()
iso.get_lmp(
    date="today",
    market=gridstatus.Markets.DAY_AHEAD_HOURLY,
    verbose=True,
)
```
Here is the output

<img width="1679" alt="image" src="https://user-images.githubusercontent.com/371792/216474692-75917d89-567c-4d72-861f-33734bcabe49.png">

PR #149, opened 2023-02-03
Fix SPP dam pulling tomorrow for today (#146)

xmattyshea:


gridstatus/spp.py
```diff
@@ -611,8 +611,17 @@ def _fs_get_dam_lmp_by_location_paths(self, date, verbose=False):
...
+        files_df["date"] = files_df.name.apply(
+            lambda x: pd.to_datetime(
+                x.strip(".csv").split("-")[-1],
+                format="%Y%m%d%H%M",
+            )
+            .normalize()
+            .tz_localize(self.default_timezone),
+        )
+
+        max_file = files_df[files_df["date"] == date]
```
kmax12, line 624: this looks great!

I would rename `max_file` to be something like `matched_file` since it is no longer the max date, just the file that matches the requested date.

kmax12 (changes requested):
just one comment and then ready to merge after

gridstatus/spp.py
xmattyshea: Updated. Thanks!

kmax12:
test errors are unrelated to your change. merging now

Issue #154, opened 2023-02-04
Add ERCOT historical Ancillary Service Prices


xmattyshea:
Ancillary service prices, like the ERCOT RT SPP prices, are available as annual archives. 

Source : https://www.ercot.com/mp/data-products/data-product-details?id=NP4-181-ER

kmax12:
implemented by #155

PR #155, opened 2023-02-04
Ercot historial ancillary

xmattyshea:
#154 

Adding support to pull ERCOT annual historial ancillary prices. The exiting functionality in `get_as_prices` supports single date up to t-30days.

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -630,6 +630,18 @@ class Ercot(ISOBase):
 
         data.rename(columns=rename, inplace=True)
 
+        # redorder to match other ancillary function
+        col_order = [
+            "Time",
+            "Market",
+            "Non-Spinning Reserves",
+            "Regulation Down",
+            "Regulation Up",
+            "Responsive Reserves",
+        ]
+
+        data = data[col_order].copy()
+
         return data
 
     def _get_spp_dam(
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -53,6 +53,9 @@ SETTLEMENT_POINT_PRICES_AT_RESOURCE_NODES_HUBS_AND_LOAD_ZONES_RTID = 12301
 # https://www.ercot.com/mp/data-products/data-product-details?id=NP3-560-CD
 SEVEN_DAY_LOAD_FORECAST_BY_FORECAST_ZONE_RTID = 12311
 
+# Historical DAM Clearing Prices for Capacity
+# https://www.ercot.com/mp/data-products/data-product-details?id=NP4-181-ER
+HISTORICAL_DAM_CLEARING_PRICES_FOR_CAPACITY_RTID = 13091
 
 """
 Settlement	Point Type	Description
@@ -575,6 +578,59 @@ class Ercot(ISOBase):
                 f"Market {market} not supported for ERCOT",
             )
         return Ercot._finalize_spp_df(df, settlement_point_field, locations)
+
+    def get_as_prices_historical(self, year, verbose=False):
+        """Get historical ancillary service clearing prices in hourly intervals
+            in Day Ahead Market
+
+        Arguments:
+            year(int): year to get data for
+
+        Returns:
+            pandas.DataFrame: A DataFrame with prices for "Non-Spinning Reserves", \
+                "Regulation Up", "Regulation Down", "Responsive Reserves".
+
+        Source:
+            https://www.ercot.com/mp/data-products/data-product-details?id=NP4-181-ER
+        """
+
+        doc_info = self._get_document(
+            report_type_id=HISTORICAL_DAM_CLEARING_PRICES_FOR_CAPACITY_RTID,
+            constructed_name_contains=f"{year}.zip",
+            verbose=verbose,
+        )
+
+        x = utils.get_zip_file(doc_info.url)
+        data = pd.read_csv(x)
+
+        data["Time"] = pd.to_datetime(
+            data["Delivery Date"]
+            + " "
+            + (data["Hour Ending"].str.split(":").str[0].astype(int) - 1)
+            .astype(str)
+            .str.zfill(2)
+            + ":00",
+        ).dt.tz_localize(
+            self.default_timezone,
+            ambiguous=data["Repeated Hour Flag"] == "Y",
+        )
+
+        data["Market"] = "DAM"
+
+        # NSPIN  REGDN  REGUP  RRS
+        # some columns from workbook contain trailing/leading whitespace
+        data.columns = [x.strip() for x in data.columns]
+
+        rename = {
+            "NSPIN": "Non-Spinning Reserves",
+            "REGDN": "Regulation Down",
+            "REGUP": "Regulation Up",
+            "RRS": "Responsive Reserves",
+        }
+
+        data.rename(columns=rename, inplace=True)
+
+        return data
 
     def _get_spp_dam(
         self,
--- a/gridstatus/tests/test_ercot.py
+++ b/gridstatus/tests/test_ercot.py
@@ -32,6 +32,30 @@ class TestErcot(BaseTestISO):
         assert df.columns.tolist() == as_cols
         assert df["Time"].unique()[0].date() == date
 
+    def test_get_as_prices_historical(self):
+        as_cols = [
+            "Time",
+            "Market",
+            "Non-Spinning Reserves",
+            "Regulation Down",
+            "Regulation Up",
+            "Responsive Reserves",
+        ]
+
+        year = 2019
+        first_date = pd.Timestamp(year, 1, 1).date()
+        df = self.iso.get_as_prices_historical(year)
+        assert df.shape[0] >= 0
+        assert df.columns.tolist() == as_cols
+        assert df["Time"].unique()[0].date() == first_date
+
+        year = year - 1
+        first_date = pd.Timestamp(year, 1, 1).date()
+        df = self.iso.get_as_prices_historical(year)
+        assert df.shape[0] >= 0
+        assert df.columns.tolist() == as_cols
+        assert df["Time"].unique()[0].date() == first_date
+
     """get_fuel_mix"""
 
     def test_get_fuel_mix(self):


kmax12 (changes requested):
Thanks for the contribution. This is actually a bit trickier to integrate than first meets the eye

in an ideal world, this would be combined with the existing `Ercot.get_as_prices` method and the code would decide whether or not to call this endpoint or the other one based on the date requested.  that way the user doesn't have to know which function to call based on date. the two functions currently share a lot of code anyways.

One challenge I see is the currently `get_as_prices` has the decorator `@support_date_range("1D")`. What that does: if a user requests a date range e.g `iso.get_as_prices(start="Jan 1, 2021", end="Dec 31, 2021")`, then the function will only get called with one day at a time. The decorator than handles combining everything back together.

This will cause some efficiency issue since this end point can support an entire year at a time.  Mean requesting the data for a full year would cause the data to be downloaded 365 times!. 

My thought is to just ignore this problem for now and fix it in a follow up PR. In that PR, we'll have to think through some gridstatus API changes probably

kmax12:
I'll throw out one idea

It seems that the annual [report](https://www.ercot.com/mp/data-products/data-product-details?id=NP4-181-ER) you are adding has all the same data as daily report with the exceptions of the last couple of days. 

The report also isn't that big of a file. one approach could be

1. Switch to using this report by default
2. Update decorator to be `@support_date_range("1Y")` so that it chunks things by year
3. Add special handling to call the daily endpoint if the last couple days are requested

xmattyshea:
Ran into some issues with the suggested fix of adding the `@support_date_range("1Y")` decorator to the function. Instead, what I did was the following:

- renamed the `get_as_prices()` original function to an internal `_get_as_prices()`, which is invoked by the new `get_as_prices()` function if only the `date` argument is passed. Otherwise, it takes the `date` and `end` arguments to iterate over the required annual files based on the years implied by the start - end range. 
- If the requested end date is not in the annual file, it takes the range of dates between the `max(data.date())` and `end` and iterates through them using the `_get_as_prices()` function. This is only necessary if `end >= today-3`. 
- Pros: retains decorator functionality for the single date request
- Cons: no argument checks from decorator for annual pulls.

gridstatus/tests/test_ercot.py
```diff
@@ -32,6 +32,14 @@ def test_get_as_prices(self):
         assert df.columns.tolist() == as_cols
         assert df["Time"].unique()[0].date() == date
 
+        date = pd.Timestamp(2022, 11, 8).date()
+        today = utils._handle_date("today").date()
```
kmax12, line 36: depending on when the test run, this can give different results for "today". can be fixed with 

```today = utils._handle_date("today", tz=selfiso.default_timezone).date()```

however, looks like `today` is already defined higher up in this test, so not sure this is needed

File: gridstatus/tests/test_ercot.py [hunk 2]
```diff
@@ -32,6 +32,14 @@ def test_get_as_prices(self):
...
+        df = self.iso.get_as_prices(date, end="today")
+        assert df.shape[0] >= 0
+        assert df.columns.tolist() == as_cols
+        assert df.Time.min().date() == date
+        assert df.Time.max().date() == today
```
kmax12, line 41: i think it's worth also checking to make sure every day between `date` and `today` is also defined in the response

gridstatus/tests/test_ercot.py
```diff
@@ -32,6 +32,14 @@ def test_get_as_prices(self):
         assert df.columns.tolist() == as_cols
         assert df["Time"].unique()[0].date() == date
 
+        date = pd.Timestamp(2022, 11, 8).date()
```
kmax12, line 35: i would add a test that requests just a single day over a year ago. for example, i believe this currently fails


```python
self.iso.get_as_prices(date=today-pd.Timedelta(days=365))
```

gridstatus/ercot.py
```diff
@@ -576,6 +579,120 @@ def get_spp(
...
+        if not end:
+            return self._get_as_prices(date)
+
+        start = utils._handle_date(date, tz=self.default_timezone)
+
+        # add one day since end is exclusive
+        end = utils._handle_date(end, tz=self.default_timezone) + pd.DateOffset(days=1)
+
+        # annual files, so get all years we'll need to iter over
+        years = {x for x in range(start.year, end.year + 1)}
```
kmax12, line 616: if you use the `@support_date_range("1Y")` decorator you shouldn't get any calls to this function that have more than one year

kmax12 (changes requested):

gridstatus/ercot.py
```diff
@@ -576,6 +579,120 @@ def get_spp(
             )
         return Ercot._finalize_spp_df(df, settlement_point_field, locations)
 
+    def get_as_prices(
```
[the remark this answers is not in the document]
xmattyshea: [adding for context]

Basic example: When you call the function with the `@support_date_range("1Y")` and `date=1/1/2020, end=1/1/2023`, it runs a loop over the following timestamps:
```
dates=[
    Timestamp('2020-01-01 00:00:00-0600', tz='US/Central'), 
    Timestamp('2020-12-31 00:00:00-0600', tz='US/Central', freq='A-DEC'), 
    Timestamp('2021-12-31 00:00:00-0600', tz='US/Central', freq='A-DEC'), 
    Timestamp('2022-12-31 00:00:00-0600', tz='US/Central', freq='A-DEC'), 
    Timestamp('2023-01-01 00:00:00-0600', tz='US/Central')
]
```
so you actually do get multiple distinct year values per iteration passed to the function (per your comment). There's not more than 1 year between dates, but that doesn't necessarily matter because the file identifier is a distinct year integer.

The function actually does still work if you use the decorator, but you still need the logic checking if there are different discreet years between `start` and `end`.

gridstatus/decorators.py
```diff
@@ -330,3 +330,28 @@ def pjm_update_dates(dates, args_dict):
...
+
+
+def ercot_update_dates(dates, args_dict):
+    date = args_dict["date"]
+    end = args_dict["end"]
+    years = {x for x in range(date.year, end.year + 1)}
+
+    fixed_dates = []
+
+    for i, year in enumerate(years):
```
kmax12, line 342: I'm not sure this logic is right.

what happens if you request data like

start = "Feb 1, 2022"
end = "March 2, 2022"

I think in this case something will go wrong.

kmax12:
the new approach looks right, but can you update tests, remove print statements, and also respond to the all unresolved comments. i reopen some that didnt actually look resolved to me

gridstatus/decorators.py
kmax12: if not, can you add a test showing it works?

gridstatus/ercot.py
```diff
@@ -576,6 +580,123 @@ def get_spp(
...
+        doc_info = self._get_document(
+            report_type_id=HISTORICAL_DAM_CLEARING_PRICES_FOR_CAPACITY_RTID,
+            constructed_name_contains=f"{date.year}.zip",
+            verbose=verbose,
+        )
+
+        x = utils.get_zip_file(doc_info.url)
+        data = pd.read_csv(x)
+
+        data["Time"] = pd.to_datetime(
```
kmax12, line 633: most of this logic looks copy and pasted from the existing get_as_price. we should find away to not have to repeat it.

gridstatus/ercot.py
```diff
@@ -576,6 +580,123 @@ def get_spp(
...
+            pandas.DataFrame: A DataFrame with prices for "Non-Spinning Reserves", \
+                "Regulation Up", "Regulation Down", "Responsive Reserves".
+
+        Source:
+            https://www.ercot.com/mp/data-products/data-product-details?id=NP4-181-ER
+        """
+
+        # use to check if we need to pull daily files
+        if (
+            not end
```
kmax12, line 611: why this `not end`? what if I request data from last 3 days? does this work?

xmattyshea: it actually works whether it's `not end` or if it's `end`. It should be changed to check if end is declared, and then pass it to `_get_as_prices_recent()` as the default. If you don't check for `end` and you pass it as an argument, it gets picked up in the `args_dict` and throws an `AttributeError` when trying to convert `None` to a `timestamp`. I'll make the appropriate edits.

xmattyshea: updated decorator logic to delete `end` from `args_dict` if value is `None`. Allowed for simplification of above issue by removing the intermediate check on the value of `end`.

gridstatus/tests/test_ercot.py
xmattyshea: Removed extra implementation of `today` and used original.

gridstatus/ercot.py
```diff
@@ -576,6 +579,120 @@ def get_spp(
...
+                "Regulation Up", "Regulation Down", "Responsive Reserves".
+
+        Source:
+            https://www.ercot.com/mp/data-products/data-product-details?id=NP4-181-ER
+        """
+
+        if not end:
+            return self._get_as_prices(date)
+
+        start = utils._handle_date(date, tz=self.default_timezone)
```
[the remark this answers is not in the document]
xmattyshea: Updated to support `@support_date_range`

gridstatus/ercot.py
```diff
@@ -311,7 +314,7 @@ def get_load_forecast(self, date, verbose=False):
         return doc
 
     @support_date_range("1D")
-    def get_as_prices(self, date, verbose=False):
+    def _get_as_prices(self, date, verbose=False):
```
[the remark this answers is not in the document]
xmattyshea: function renamed to `_get_as_prices_recent()`

gridstatus/decorators.py
```diff
@@ -330,3 +335,33 @@ def pjm_update_dates(dates, args_dict):
...
+
+
+def ercot_update_dates(dates, args_dict):
+    date = args_dict["date"]
+    end = args_dict["end"]
+    years = {x for x in range(date.year, end.year + 1)}
+
+    fixed_dates = []
+
+    if date.year == end.year:
```
kmax12, line 347: i think in the case where this is true you can just return `dates`, since no modification is needed.

also, i would move this log up to the top of the function.

gridstatus/decorators.py
```diff
@@ -150,6 +154,7 @@ def wrapped_f(*args, **kwargs):
 
             # every None removes two possible queries
             total = len(dates) - dates.count(None) * 2 - 1
+            print(dates)
```
kmax12, line 157: remove print statement

gridstatus/ercot.py
```diff
@@ -576,6 +580,120 @@ def get_spp(
...
+            "Market",
+            "Non-Spinning Reserves",
+            "Regulation Down",
+            "Regulation Up",
+            "Responsive Reserves",
+        ]
+
+        max_date = data.Time.max().date()
+
+        df_list = []
```
kmax12, line 669: can just write `df_list = [data]`

gridstatus/ercot.py
```diff
@@ -576,6 +580,120 @@ def get_spp(
...
+        # join, sort, filter and reset data index
+        data = (pd.concat(df_list).sort_values(by="Time"))[col_order].copy()
+
+        data = (
+            data.loc[
+                (data.Time.dt.date >= date.date()) & (data.Time.dt.date <= end.date())
+            ]
+            .drop_duplicates(subset=["Time"])
+            .reset_index(drop=True)
+            .copy()
```
kmax12, line 692: is this copy needed?

gridstatus/ercot.py
```diff
@@ -311,7 +315,7 @@ def get_load_forecast(self, date, verbose=False):
         return doc
 
     @support_date_range("1D")
-    def get_as_prices(self, date, verbose=False):
+    def _get_as_prices_recent(self, date, verbose=False):
         """Get ancillary service clearing prices in hourly intervals in Day Ahead Market
```
kmax12, line 319: need to update doc string to say it's for recent only

kmax12 (changes requested):
it's looking good! once all the remaining comments are resolved we'll be able to merge

gridstatus/ercot.py
```diff
@@ -646,6 +710,63 @@ def _finalize_spp_df(df, settlement_point_field, locations):
         df = df.reset_index(drop=True)
         return df
 
+    @staticmethod
+    def _finalize_as_price_df(doc, timezone, ambiguous_column, pivot=False):
```
kmax12, line 714: you shouldn't need to pass `timezone` since you can just use `Ercot.default_timezone`

gridstatus/ercot.py
xmattyshea: refactored shared logic into `_finalize_as_price_df`

gridstatus/ercot.py
```diff
@@ -576,6 +579,120 @@ def get_spp(
...
+            doc_info = self._get_document(
+                report_type_id=HISTORICAL_DAM_CLEARING_PRICES_FOR_CAPACITY_RTID,
+                constructed_name_contains=f"{year}.zip",
+                verbose=verbose,
+            )
+
+            x = utils.get_zip_file(doc_info.url)
+            data = pd.read_csv(x)
+
+            data["Time"] = pd.to_datetime(
```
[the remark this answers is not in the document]
xmattyshea: refactored shared logic into `_finalize_as_price_df`

gridstatus/ercot.py
xmattyshea: resolved using suggestion

kmax12 (approved):
Looks great!

PR #167, opened 2023-02-20
Add code coverage with Codecov

### LINKED ISSUE #165: Add Code Coverage
Would be nice to generate a codecoverage number like libraries such as https://github.com/alteryx/featuretools

image

Should to add to both readme and github actions that run on every PR

gsheni:
- Fixes #165


gsheni:
@kmax12 ready for review. It should work once the repo is setup on Codecov

kmax12:
i added the codecov secret to the repo, but the 3.11 tests arent even reaching the point at which it runs. any idea what might be going on? going to try to rerun

gsheni:
@kmax12 fixed it. 
I forgot to add the slow/re-run args to `make test-cov` command (leading the CI to run the slow unit tests).

gsheni:
The tests failed but I've seen this before. They failed towards the end of day ET (because the data is not available to query for the new day?).

kmax12:
looking good. the only issue seems to be with codecov token now. i suspect it has to do with the being a fork so the secret token isn't available. going to merge in and see if it resolves itself. 

thanks again for the contribution!

gsheni:
Codecov isn't working on `main` is because the unit tests are **not run on** merge to `main`. The `tests.yaml` workflow needs to be updated.
Does it matter if all Python unit tests are run on main? Do you only want to run unit tests for Python 3.11 on merge to `main`?

kmax12:
Ah, let's just do 3.11 on main

PR #168, opened 2023-02-21
Change MISO Default Timezone

kmax12:


jcsalterego:
Did they, uh, move?

kmax12:
😂 no but i believe their api returns EST weather or not it's daylight savings time, so this is undoing me trying to handle that when i first implemented it

PR #175, opened 2023-03-02
Fix isone dst

blakeShalom:
Fixes this issue

https://github.com/kmax12/gridstatus/issues/171

Small refactor on shared code between the two markets. 

May run into the same issue on the "latest" code as well. Could either fix or open a new issue.


kmax12:
still need to review, but just wanted to comment that the fail tests are unrelated to this PR. sorry about the broken tests!

Issue #171, opened 2023-02-27
ISO Get LMP Fails on 2018-11-04


kmax12:
## Reproduce

```python
import gridstatus

iso = gridstatus.ISONE()
iso.get_lmp(
    date="2018-11-04",
    market="DAY_AHEAD_HOURLY",
)
```

```
AmbiguousTimeError                        Traceback (most recent call last)
Cell In[38], line 4
      1 import gridstatus
      3 iso = gridstatus.ISONE()
----> 4 iso.get_lmp(
      5     date="2018-11-04",
      6     market="DAY_AHEAD_HOURLY",
      7 )

File ~/Documents/isodata/gridstatus/lmp_config.py:26, in lmp_config.__call__.<locals>.wrapper(*args, **kwargs)
     24 if len(args) > 0 and isinstance(args[0], ISOBase):
     25     bound_args = self._verify_bound_args(bound_args)
---> 26     return self._class_method_wrapper(func, bound_args)
     27 else:
     28     # This is a runtime check after method is called.
     29     # Possible improvement: move this to "compile"-time.
     30     raise ValueError("Must be class method on ISOBase")

File ~/Documents/isodata/gridstatus/lmp_config.py:61, in lmp_config._class_method_wrapper(func, bound_args)
     59 instance_args = bound_args.args
     60 instance_kwargs = bound_args.kwargs
---> 61 return func(*instance_args, **instance_kwargs)

File ~/Documents/isodata/gridstatus/decorators.py:78, in support_date_range.__call__.<locals>.wrapped_f(*args, **kwargs)
...
File ~/Library/Application Support/pypoetry/venv/lib/python3.10/site-packages/pandas/_libs/tslibs/tzconversion.pyx:313, in pandas._libs.tslibs.tzconversion.tz_localize_to_utc()

File ~/Library/Application Support/pypoetry/venv/lib/python3.10/site-packages/pandas/_libs/tslibs/tzconversion.pyx:553, in pandas._libs.tslibs.tzconversion._get_dst_hours()

AmbiguousTimeError: There are 3 dst switches when there should only be 1.
```

Likely a problem in the pandas date parsing code. This would be a good first issue.

blakeShalom:
I'll take a look at this one for a first task

blakeShalom:
Ah. 

[...]

PR #177, opened 2023-03-04
Fixes duplicate and multiple locations with same id #171

blakeShalom:


blakeShalom:
Looks like sometimes the inverse is true and Location ID is present and Location isn't. Trying to replicate locally and then I'll push the change.

kmax12:
looks good. thanks!

PR #178, opened 2023-03-13
Fix ERCOT DST issue

kmax12:
Fix DST issue by using Dateoffset instead of Timedelta

This shifts things by one day rather than 24 hrs. Should probably audit the rest of codebase for any potential similar issues


kmax12:
3.7 tests still failing. Time to drop 3.7 support

Issue #259, opened 2023-08-10
ERCOT: Load Forecasts


colinpbowen:
Load forecasting is one of the most important operations in ISO Planning. It therefore behooves us to understand the different types of forecasts issued by ERCOT. 

The [Load Forecast](https://www.ercot.com/gridinfo/load/forecast) page on ERCOT summarizes key information. ERCOT supplies two types of Load Forecasts: the Long-Term Load Forecast (LTLF) and the Medium-Term Load Forecast (MTLF).

The LTLF is an hourly forecast for the next 10 years based on forecasted economic data and historical weather. They offer two forecast products:
1. Monthly Peak Demand (in MW) and Total Energy consumed in the month (in MWh) for the next 10 years. 
2. The main product, the Hourly Forecast for the next 10 years, by Weather Zone, for gross load, PV forecast, and net load. 

They also include Load Forecast Scenarios, which include peak demand scenarios, including both coincident and non-coincident peaks. 

The MLTF is an hourly forecast for the next 168 hours (7 days) based on current weather forecast parameters within each weather zone. Load forecast error is measured by Mean Absolute Percentage Error; last month, the day-ahead forecast had a MAPE of ~2.7%, the 6 hour ahead ~2.4%, and the 3 hour ahead ~1.6%. 

In EMIL, we can find the following data products which significantly expand upon the scope described above. These include some intra-hour products. Descriptions are provided.

(LTLF) [Long Term Hourly Peak Demand and Energy Forecast](https://www.ercot.com/mp/data-products/data-product-details?id=NP3-784-M)
(MLTF) [Seven-Day Load Forecast by Weather Zone](https://www.ercot.com/mp/data-products/data-product-details?id=NP3-561-CD). Seven-Day Load Forecast by Forecast Zone. This is supplemented by the already-scraped Actual System Load by Forecast Zone. 
(MLTF) [Seven-Day Load Forecast by Forecast Zone](https://www.ercot.com/mp/data-products/data-product-details?id=NP3-560-CD)
(MLTF) [Seven-Day Load Forecast by Model and Weather Zone](https://www.ercot.com/mp/data-products/data-product-details?id=NP3-565-CD). Hourly system-wide Mid-Term Load Forecasts (MTLFs) for all forecast models with an indicator for which forecast was in use by ERCOT at the time of publication for current day plus the next 7.
[...]

PR #260, opened 2023-08-14
ERCOT Load Forecasts #259

colinpbowen:
Added functionality to grab all types of medium-term load forecasts from ERCOT, and refactored to follow the _handle_doc() functionality as implemented elsewhere. 

Part of #259

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -542,8 +542,6 @@ class Ercot(ISOBase):
                 "Forecast Time",
             ],
         )
-        # forecast_time = df.pop('Forecast Time')
-        # df.insert(loc=2, column='Forecast Time', value=forecast_time)
 
         return df
 
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -1,7 +1,7 @@
 import io
 from dataclasses import dataclass
-from zipfile import ZipFile
 from enum import Enum
+from zipfile import ZipFile
 
 import pandas as pd
 import requests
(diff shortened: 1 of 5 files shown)


gridstatus/ercot.py
```diff
@@ -453,24 +473,37 @@ def _get_todays_outlook_non_forecast(self, date, verbose=False):
...
 
         # intrahour https://www.ercot.com/mp/data-products/data-product-details?id=NP3-562-CD
         # there are a few days of historical date for the forecast
         today = pd.Timestamp.now(tz=self.default_timezone).normalize()
-        doc_info = self._get_document(
-            report_type_id=SEVEN_DAY_LOAD_FORECAST_BY_FORECAST_ZONE_RTID,
+        
+        doc = self._get_document(
+            report_type_id=forecast_type.value,
             date=today,
```
kmax12, line 499: since we're revisiting this function, let's model it after the `get_hourly_wind_report` and use `published_before=None if date == "latest" else date` rather than date. 

when we do this, we also need to remove this line above that rounds the supplied date to the beginning of the day

```python
today = pd.Timestamp.now(tz=self.default_timezone).normalize()
```

kmax12 (changes requested):
thank you for the contribution — really good work for first issue on the repo! I left a few comments. 

also, check out the [contributing guide](https://github.com/kmax12/gridstatus/blob/main/CONTRIBUTING.md#running-tests-and-linting) for how to run tests and linting locally.

we will need to update the testing for this function. `test_get_hourly_wind_report` in `test_ercot.py` is a good example. 

ping me on slack to discuss any of this.

gridstatus/ercot.py
```diff
@@ -65,10 +66,29 @@
 # https://www.ercot.com/mp/data-products/data-product-details?id=NP6-905-CD
 SETTLEMENT_POINT_PRICES_AT_RESOURCE_NODES_HUBS_AND_LOAD_ZONES_RTID = 12301
 
+
+class SevenDayLoadForecast(Enum):
```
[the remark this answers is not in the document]
colinpbowen: How have you made a distinction between reports and non-reports within the EMIL ecosystem? To me it seems like either everything is a report or nothing is. And technically, they aren't named "reports" [on EMIL.](https://www.ercot.com/mp/data-products/data-product-details?id=NP3-560-CD)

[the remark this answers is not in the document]
kmax12: not sure i follow. at the link, the number is called a "Report Type Id". 

image

gridstatus/tests/base_test_iso.py
```diff
@@ -230,7 +231,11 @@ def test_get_load_today(self):
 
     def test_get_load_forecast_historical(self):
         test_date = (pd.Timestamp.now() - pd.Timedelta(days=14)).date()
-        forecast = self.iso.get_load_forecast(date=test_date)
+        test_fcast_type = ERCOTSevenDayLoadForecastReport.BY_FORECAST_ZONE
```
kmax12, line 234: this is the base class that is used across all the tests for each individual iso. you want to update `test_ercot.py`. Because the return value for ercot is no different than the isos, you'll have to override this test on `TestErcot` by defining it

gridstatus/ercot.py
```diff
@@ -453,40 +477,75 @@ def _get_todays_outlook_non_forecast(self, date, verbose=False):
 
         return data.reset_index(drop=True)
 
-    def get_load_forecast(self, date, verbose=False):
-        """Returns load forecast
+    @support_date_range("HOUR_START")
+    def get_load_forecast(
+        self,
+        date,
+        forecast_type: ERCOTSevenDayLoadForecastReport,
```
kmax12, line 484: to maintain existing behavior, can we make the `BY_FORECAST_ZONE` report the default parameter? 

this will also partiaily fix some of the ercot tests that are failing

kmax12 (commented):
code itself is looking good! just need to get the ercot load_forecast related tests to pass (ignore the others that are failing)

kmax12 (approved):
Just one last comment. will merge after that!

kmax12:
Thanks for the contribution! Just merged

Issue #269, opened 2023-08-28
Add ERCOT System Lambda 


kmax12:
https://www.ercot.com/mp/data-products/data-product-details?id=NP6-322-CD

PR #272, opened 2023-09-13
Add ERCOT  System Lambda

Michael-fore:
#269 

For some reason I decided to work on this one first. 

I added a function to pass through the self.read_doc to override the columns renaming of the parse_doc
```
        def rename_func(df):
            df[["DeliveryDate", "HourEnding"]] = df["SCEDTimeStamp"].str.split(
                " ",
                expand=True,
            )
            return df

        df = self.read_doc(doc=doc, rename_func=rename_func)
```
and
```
if rename_func:
            doc = rename_func(doc)
else:
            doc.rename( ...)
```

Let me know if you don't want to use this approach

Excited for my first contribution!

#


gridstatus/ercot.py
```diff
@@ -1773,6 +1777,41 @@ def _make_bid_curve(df):
 
         return self.parse_doc(df, verbose=verbose)
 
+    # @support_date_range("DAY_START") ??
```
kmax12, line 1780: we can use `@support_date_range("DAY_START")` as is if we use the get document approach I mention below

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,41 @@ def _make_bid_curve(df):
...
+        Arguments:
+            date (str, datetime): date to get data for
+            verbose (bool, optional): print verbose output. Defaults to False.
+
+        Returns:
+            pandas.DataFrame: A DataFrame
+
+        """
+
+        doc = self._get_document(
```
kmax12, line 1797: let's make this function return a full day of data if `date != 'latest'`

in that case where `date != latest`, we can just pass the date along to use self._get_documents instead and it will return a list of documents for that day

you can then parse and concatenate them

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,41 @@ def _make_bid_curve(df):
...
+            report_type_id=SCED_SYSTEM_LAMBDA_RTID,
+            date=date,
+            verbose=verbose,
+        )
+
+        def rename_func(df):
```
kmax12, line 1803: does it make sense for this dataset to get parsed to an interval start and interval end? 

it seems to me like it would be fine to just leave the columns as`SCED Time Stamp` and `System Lambda` for now. if we do want to do parse to intervals, we'll need some logic like you see [here](https://github.com/kmax12/gridstatus/blob/main/gridstatus/ercot.py#L1143). Basically, this converts the sced time stamp to the 15 minute spp interval. 

in either case,  you'll have to update `read_doc` function or just create a new one, since it doesn't really handle this well. right now. 

We will face a similar question when we integrate the real time 5 minute LMPs (we currently only do SPP).

kmax12 (changes requested):
thanks for the contribution. here are some initial comments

gridstatus/ercot.py
Michael-fore: I am so down to not  deal with interval ending if we don't need to 

IIRC this is an interval ending data type, but we can reach out to ERCOT to ask if need be

gridstatus/ercot.py
Michael-fore: Is speed an issue you solve for?

As of now, fetching a whole day takes awhile.

I can speed it up if by batch fetching, but only want to do that if necessary

kmax12: i think supporting a whole day and latest is enough for now. in theory it be nice to be able to specify a finer granularity, but that would require some reworking to get_documents. i would totally welcome it, but maybe out of scope

gridstatus/ercot.py
kmax12: let's just keep it as non interval ending for now and we can also enhance later

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,50 @@ def _make_bid_curve(df):
...
+
+            return df
+        else:
+            doc = self._get_documents(
+                report_type_id=SCED_SYSTEM_LAMBDA_RTID,
+                date=date,
+                verbose=verbose,
+                extension="csv",
+            )
+            df = pd.DataFrame()
```
kmax12, line 1812: small thing, it's more efficient to rewrite this do just one concat

```python
dfs = []
for doc in docs:
    log(f"Reading {doc.url}", verbose)
    new_df = pd.read_csv(doc.url, compression="zip")
    dfs.append(new_df)
df = pd.concat(dfs)
```

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,50 @@ def _make_bid_curve(df):
...
+            pandas.DataFrame: A DataFrame
+
+        """
+        if date == "latest":
+            doc = self._get_document(
+                report_type_id=SCED_SYSTEM_LAMBDA_RTID,
+                date=date,
+                verbose=verbose,
+            )
+            df = pd.read_csv(doc.url, compression="zip")
```
kmax12, line 1802: maybe just do `docs = [doc]` here and then have shared logic for reading the csv

gridstatus/ercot.py
Michael-fore: Just went the easy route on this one

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,50 @@ def _make_bid_curve(df):
...
+
+            return df
+        else:
+            doc = self._get_documents(
```
kmax12, line 1806: i think making this plural variable called `docs` make more sense

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,49 @@ def _make_bid_curve(df):
...
+                report_type_id=SCED_SYSTEM_LAMBDA_RTID,
+                date=date,
+                verbose=verbose,
+                extension="csv",
+            )
+            df = pd.DataFrame()
+            for i in doc:
+                new_df = pd.read_csv(i.url, compression="zip")
```
kmax12, line 1814: this appears to be missing handle of the date time parsing, dst changes, and setting of the time zone

for example: https://github.com/kmax12/gridstatus/blob/main/gridstatus/ercot.py#L2028

if you can modify parse_doc function you maybe be able to inherit some of this logic for free

kmax12 (changes requested):

gridstatus/ercot.py
Michael-fore: I usually prefer more verbose logic, but changed as well

gridstatus/ercot.py
Michael-fore: nice catch, fixed

gridstatus/ercot.py
Michael-fore: Oh i thought we were good to ignore that, i may just redo the rename_func logic to inherit that logic

Michael-fore:
Still fixing dst stuff

Michael-fore:
@kmax12 Do you also shift the short day hour? So from 134 to 124?

gridstatus/ercot.py
Michael-fore: I uh, missed this one. lmk if you want it fixed

gridstatus/ercot.py
```diff
@@ -1773,6 +1778,57 @@ def _make_bid_curve(df):
...
+            # shift back to display correct how with new tz
+            if dst_flag == "Y":
+                nt = time + pd.Timedelta(hours=1)
+                nt.dt.tz_localize(self.default_timezone, infer=True)
+                nt = time - pd.Timedelta(hours=1)
+            else:
+                return time
+
+        # ANCHOR - handle dst
+        df["SCEDTimeStamp"] = np.vectorize(handle_dst)(
```
kmax12, line 1825: i think it's better to avoid numpy here and just use the pandas version, which i think is sometihng like this 

```
df["SCEDTimeStamp"] = df.apply(lambda row: handle_dst(row["SCEDTimeStamp"], row["RepeatedHourFlag"]), axis=1)
```

kmax12: scratch that, i would just do this 

df["SCEDTimeStamp"] = pd.to_datetime(df["SCEDTimeStamp"]).dt.tz_localize(self.default_timezone, ambiguous=df["RepeatedHourFlag"] == "Y")

gridstatus/ercot.py
```diff
@@ -1773,6 +1778,57 @@ def _make_bid_curve(df):
...
+            df["SCEDTimeStamp"],
+            df["RepeatedHourFlag"],
+        )
+
+        return df
```
kmax12, line 1830: to be consistent with how we name variables, we want the final column names to be "SCED Time Stamp" and "System Lambda". Drop repeated hour since the time stamp column is properly typed and has timezone info

gridstatus/ercot.py
kmax12: actually, can you check the logic of how the ambiguious param works and the repeated hour flag? 

unclear to me if RepeatedHourFlag is Y or N when it is DST. maybe need to be ambiguous=df["RepeatedHourFlag"] == "N"

here's the documentation: https://pandas.pydata.org/docs/reference/api/pandas.Series.tz_localize.html

Michael-fore: from the docs

```bool-ndarray where True signifies a DST time, False designates a non-DST time (note that this flag is only applicable for ambiguous times)```

I believe those with `"Y"` are the ambiguous times, but we can also wait to hear back from ERCOT

Comment (kmax12) [reply gridstatus/ercot.py]: based on our slack discussion, i thought we settled that it is set to Y when the time repeats itself during the transition from DST to Standard Time, which would mean that the value of N represents DST. 

but yes, i have a question into ercot and let's see what they say

Comment (stephennimmo) [reply gridstatus/ercot.py]: Using an hourly example, when there is a DST event, you'll have 23 hourly events or 25 hourly events. In the xml, that Y flag denotes the event that is a duplicate based on the timestamps. Your assumption is correct. 

FWIW, I've typically seen this implemented by getting rid of "timezone on write", but rather storing the data as utc and then applying timezone on read. Then, you could just use Unix-epoch milliseconds as a format and store the appropriate timezone info in some sort of metadata. However, there are pros/cons to that as well like what happens with timezones change (they do) or having to do timezone conversions 1000s of times for every historical read. 

I've also seen the records flattened, where there is a single row for each day, with a column for each hourly ending value with the column name being the time sans punctuation. To support DST, in the hourly example, you would have a row with a date column followed by 25 potential hourly columns (0000, 0100, 0200, etc). This look great from a storage efficiency because now you are eliminating a bunch of repetitive data however if you want to read a specific column, you'll end up having to write some query magic and would at a minimum typically be reading a full daily row back.

File: gridstatus/ercot.py [hunk 10]
```diff
@@ -1985,6 +2033,9 @@ def parse_doc(self, doc, verbose=False):
                 "Delivery Interval": "DeliveryInterval",
                 # fix whitespace in column name
                 "DSTFlag    ": "DSTFlag",
+                # for SCED LAMBDA
```
kmax12, line 2036: i think you can back out these changes

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,49 @@ def _make_bid_curve(df):
...
+            extension="csv",
+        )
+
+        if date == "latest":
+            doc = self._get_document(**kwargs)
+            docs = [doc]
+        else:
+            docs = self._get_documents(**kwargs)
+
+        df = pd.concat([pd.read_csv(i.url, compression="zip") for i in docs])
```
kmax12, line 1811: to make this easier to use with backfilling data that is sent over directly by ERCOT, can you separate out the logic for handling documents into a _handle_sced_system_lambda function? you can see an example of this here: https://github.com/kmax12/gridstatus/blob/main/gridstatus/ercot.py#L1889

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,49 @@ def _make_bid_curve(df):
...
+
+        df["SCEDTimeStamp"] = pd.to_datetime(df["SCEDTimeStamp"]).dt.tz_localize(
+            self.default_timezone,
+            ambiguous=df["RepeatedHourFlag"] == "Y",
+        )
+
+        df["System Lambda"] = df["SystemLambda"].astype("float64")
+
+        df.drop("SystemLambda", axis=1, inplace=True)
```
kmax12, line 1820: the final column names should be just: "SCED Time Stamp", "System Lambda". I dont think that is what this code does or what you are testing for

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,49 @@ def _make_bid_curve(df):
...
+            doc = self._get_document(**kwargs)
+            docs = [doc]
+        else:
+            docs = self._get_documents(**kwargs)
+
+        df = pd.concat([pd.read_csv(i.url, compression="zip") for i in docs])
+
+        df["SCEDTimeStamp"] = pd.to_datetime(df["SCEDTimeStamp"]).dt.tz_localize(
+            self.default_timezone,
+            ambiguous=df["RepeatedHourFlag"] == "Y",
```
kmax12, line 1815: per our voluminous other discussion I think this should be 

`ambiguous=df["RepeatedHourFlag"] == "N"`

the repeated hour happens during standard time. we want ambiguous to be true during DST time.

gridstatus/tests/test_ercot.py
```diff
@@ -10,6 +10,19 @@
 class TestErcot(BaseTestISO):
     iso = Ercot()
 
+    def test_get_sced_system_lambda(self):
```
kmax12, line 13: fix this test to make sure it passes. add test for  date="today" in addition to date='latest'

kmax12 (changes requested):
no need to wait for response from ercot to merge this in. left some more comments and it should be ready to merge once they are addressed

Michael-fore:
Changes implemented, if not this push then the next we should be good

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,59 @@ def _make_bid_curve(df):
...
+
+        df["System Lambda"] = df["SystemLambda"].astype("float64")
+
+        df.drop(
+            ["SystemLambda", "SCEDTimeStamp", "RepeatedHourFlag"],
+            axis=1,
+            inplace=True,
+        )
+
+        return df
```
kmax12, line 1831: sorry, just realized this in my final review. we normal return data in chronologically ascending order. 

can you add a `df.sort_values("SCED Time Stamp")`?

gridstatus/ercot.py
```diff
@@ -1773,6 +1777,59 @@ def _make_bid_curve(df):
...
+        if date == "latest":
+            doc = self._get_document(**kwargs)
+            docs = [doc]
+        else:
+            docs = self._get_documents(**kwargs)
+
+        df = self._handle_sced_system_lambda(docs)
+        return df
+
+    def _handle_sced_system_lambda(self, docs):
```
kmax12, line 1814: this should take a `verbose` parameter. 

before reading the URL, add logging statement.

also wrap the loop in tqdm for a progress bar. helping since a day downloads many files 

here is an example: https://github.com/kmax12/gridstatus/blob/main/gridstatus/ercot.py#L950

kmax12 (changes requested):
just a couple more comments. the last thing is can you add a line item to `CHANGELOG.md`? 

before you do that, I recommend merging in `main` since I already created an entry for the next release

Michael-fore:
re merged to main and implemented the  asks

gridstatus/ercot.py
```diff
@@ -1787,6 +1791,65 @@ def _make_bid_curve(df):
...
+            extension="csv",
+        )
+
+        if date == "latest":
+            doc = self._get_document(**kwargs)
+            docs = [doc]
+        else:
+            docs = self._get_documents(**kwargs)
+
+        df = self._handle_sced_system_lambda(docs)
```
kmax12, line 1825: tests arent passing because this is missing the required `verbose` parameter. let me know if you need help setting up tests to run. 

ill run it on github action, but unfortunately its current broken due to an unrelated issue of requests to SPP's server not working on github

Michael-fore: Oh i have no idea how to set up those tests. Help would be nice

Michael-fore: does spp block the github server or something?

Michael-fore: verbose added tho

kmax12: start here for tests: https://github.com/kmax12/gridstatus/blob/main/CONTRIBUTING.md#running-tests-and-linting

kmax12: i think the problem with SPP has to do with SSL cert configuration on the docker image used for the github action. i need to investigate more

gridstatus/ercot.py
```diff
@@ -1787,6 +1791,65 @@ def _make_bid_curve(df):
...
+        else:
+            docs = self._get_documents(**kwargs)
+
+        df = self._handle_sced_system_lambda(docs, verbose=verbose)
+
+        return df
+
+    def _handle_sced_system_lambda(self, docs, verbose):
+
+        if verbose:
```
kmax12, line 1831: you dont need this, `if verbose`. log already handles it

Michael-fore:
failing from gridstatus/tests/test_spp.py::TestSPP::test_get_lmp_historical[Markets.DAY_AHEAD_HOURLY]

kmax12 (approved):
Looks good and ready to merge. Thanks for contributions!

PR #273, opened 2023-09-14
allow specification of facets in EIA.get_dataset

JMorganUSU:
Minor change to allow specification of facets in get_dataset function for EIA. I'm trying to get generation data for one balancing authority, so being able to include the facets I care about in get_dataset allows me to only request data for the one, instead of getting all BAs then throwing away most of the data. facets is an optional parameter, ensuring backwards compatibility with previous versions.


gridstatus/eia.py
```diff
@@ -54,7 +54,7 @@ def _fetch_page(self, url, headers):
         df = pd.DataFrame(response["data"])
         return df, response["total"]
 
-    def get_dataset(self, dataset, start, end, n_workers=1, verbose=False):
+    def get_dataset(self, dataset, start, end, facets={}, n_workers=1, verbose=False):
```
kmax12, line 57: Using mutable default arguments like facets={} is risky in Python. The default value is shared across all calls to the function. If facets is modified in one call, the modified value will be used in subsequent calls, leading to unexpected behavior.

Use None and set to a new dictionary inside the function instead:

```python
def get_dataset(self, dataset, start, end, facets=None, n_workers=1, verbose=False):
    if facets is None:
        facets = {}
```

This way, a new dictionary is created for each function call, avoiding the shared state issue.

kmax12 (changes requested):
thanks! other than one comment the only thing missing a test. you can define that in test_eia.py

JMorganUSU:
Made requested changes, including adding a test (which passes) and updating the changelog.

WillKoehrsen:
If the build passes, this should be ready to merge. There is a minor conflict in `CHANGELOG.md` that needs to be fixed after merging the latest `main` into the PR branch.

JMorganUSU:
Are any other changes needed or is this ready to merge?

WillKoehrsen (approved):
👍🏽 Thanks for the contribution!

PR #274, opened 2023-09-16
NYISO Ancillary Services

Michael-fore:
Does this data need to be normalized or de normalized? Is there a special format output you would like as well?


kmax12:
take a look at how `get_as_prices` in ercot and caiso and try to make it look as similar as possible.

Michael-fore:
I don't see that we are handling DST for NYISO,  so i didn't include logic. Just going to work on `get_rt_ancillary_services`, once that is good will apply the logic to da

I ended up copying get_fuel_mix from NYISO, the data types al already de normalized so no need to pivot. 

If the function looks good to you i will go ahead and write a few tests

kmax12:
Since this has been open awhile. Closing this and added new issue to track: #333

PR #283, opened 2023-10-10
Use pyarrow-backed DataFrame with read_csv

gsheni:
- Use PyArrow backed data types for at reading data since PyArrow can be significantly faster and more memory-efficient than NumPy
- Starting with pandas 3.0 will change the default for string strings to PyArrow (such as when calling read_csv). 
  - [Source](https://twitter.com/pandas_dev/status/1697608898960675261) 
- Example:
  - Screenshot 2023-10-10 at 7 35 58 PM

### DIFF
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -529,7 +529,7 @@ class CAISO(ISOBase):
         msg = f"Downloading interconnection queue from {url}"
         log(msg, verbose)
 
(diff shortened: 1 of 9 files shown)


WillKoehrsen:
@kmax12 Is this still something we want to do? Should be relatively easy to make the changes.

kmax12:
i think all else equal, worth doing. not sure if it'll have a huge impact on us right now, so if it causes downstream problems, we can punt for now

PR #288, opened 2023-11-03
remove get_status and related funcs from spp and test_spp

MaxPowerWasTaken:


gridstatus/tests/base_test_iso.py
```diff
@@ -255,11 +255,12 @@ def test_get_load_forecast_today(self):
     """get_status"""
 
     def test_get_status_latest(self):
-        status = self.iso.get_status("latest")
-        assert isinstance(status, GridStatus)
+        if not isinstance(self.iso, gridstatus.spp.SPP):
```
kmax12, line 258: rather than skip the test in the base class, we can implement it a test that just checks for not implemented. see example here in pjm tests: https://github.com/kmax12/gridstatus/blob/main/gridstatus/tests/test_pjm.py#L173

kmax12 (changes requested):
looks good. a lot of needed code deleted. just one small comment

gridstatus/spp.py
```diff
@@ -894,100 +885,6 @@ def _get_leaf_elements(elems):
                 accum.append(elem)
         return accum
 
-    def _get_status_candidate_texts(self, html):
```
kmax12, line 897: i think _get_leaf_elements above can be deleted too

gridstatus/spp.py
```diff
@@ -894,100 +885,6 @@ def _get_leaf_elements(elems):
...
-            tz=self.default_timezone,
-        )
-        return next(iter(last_updated_timestamps), None)
-
-    def _get_status_status_and_notes(self, candidate_texts):
-        """Extracts (status, notes,) tuple from candidates texts"""
-        status_texts = self._match(
-            RELIABILITY_LEVELS,
-            candidate_texts,
-            haystack_norm_fn=lambda x: self._clean_status_text(x),
```
kmax12, line 970: i think you need to delete `_clean_status_text` as well

gridstatus/spp.py
```diff
@@ -894,100 +885,6 @@ def _get_leaf_elements(elems):
...
-            """
-            for keyword in RELIABILITY_LEVELS:
-                pos = text.lower().find(keyword.lower())
-                if pos > -1:
-                    pos += len(keyword)
-                    new_list.append(text[pos:])
-            new_list.append(text)
-        timestamp_texts = new_list
-
-        last_updated_timestamps = self._extract_timestamps(
```
kmax12, line 958: delete _extract_timestamps

PR #295, opened 2023-11-17
added additional nodes to pjm.py

cwaldoch:


gridstatus/pjm.py
```diff
@@ -80,7 +80,330 @@ class PJM(ISOBase):
         "124076095",
         "970242670",
         "1709725933",
-    ]
```
kmax12, line 83: are these zones? if not, perhaps we create a new variable called `selected_nodes`? i like this approach to because it's is backwards compatible too

gridstatus/pjm.py
```diff
@@ -80,7 +80,333 @@ class PJM(ISOBase):
...
+        '5021703',]
```
kmax12, line 83: looks like there is one extra here. is this supposed to be under `price_node_ids`

PR #301, opened 2023-12-02
Suport range for ERCOT load forecast

kmax12:


gridstatus/ercot.py
```diff
@@ -525,44 +525,68 @@ def _get_todays_outlook_non_forecast(self, date, verbose=False):
 
         return data.reset_index(drop=True)
 
-    @support_date_range("HOUR_START")
+    @support_date_range(frequency=None)
```
WillKoehrsen, line 528: If `frequency=None`, then `support_date_range` only runs the wrapped function once IIUC. What's the benefit of using the decorator without a frequency?

WillKoehrsen: Ah, I see, `support_date_range` also parses and standardizes dates.

WillKoehrsen (approved):

gridstatus/ercot.py
kmax12: right now it's needed because the logic to validate normalize dates. 

most important line is this one: https://github.com/kmax12/gridstatus/blob/main/gridstatus/decorators.py#L91

PR #307, opened 2023-12-27
Raw interconnection queue data

### LINKED ISSUE #305: Allow access to raw ISO interconnection queue data?
Hi! I work with @TrentonBush and @bendnorman at [Catalyst Cooperative](https://catalyst.coop/).

## Overview

As mentioned in #76 and #290, there are still some interesting questions about the different ways different ISOs report this data. At Catalyst, we'd like to access the raw data directly, to help understand and answer some of those questions.

## Rough Plan

I think we could do this by adding an `ISOBase.get_raw_interconnection_queue() -> BinaryIO` method to `ISOBase`, along with its subclasses. We'd then consume the output of this in the various `get_interconnection_queue()` methods.

I know, adding to the public API is risky, because that makes promises to all future users. But, if we want to allow people to access the raw data directly, then that behavior must be exposed somehow...

We're already splitting out the raw data extraction in some of the ISO classes, sort of.

For example, in `ISONE.get_interconnection_queue()` we already call `ISONE._extract_interconnection_queue()`, which *mostly* handles all the junk required to get a relevant blob of HTML into memory. It also handles transforming that into a Pandas dataframe at all, before passing control back to the caller which does further transformations on the data.

And in `ERCOT.get_interconnection_queue()` we call `._get_document()`, which handles much of the complicated aspects of the extraction process. We handle the rest of the process in the body of `get_interconnection_queue` itself.

For most of the other ISOs, this `get_raw_interconnection_queue()` would be a very simple method that just runs a GET request.

### Side benefit

This sets us up to split out the logic a little bit more in the future, so that the `get_interconnection_queue()` logic looks something like:

```python
def get_interconnection_queue(self):
    raw_data = self.get_raw_interconnection_queue()
    input_dataframe = pd.read_excel(raw_data, ...)
    cleaned_dataframe = self.transform_interconnection_queue(input_dataframe)
    return cleaned_dataframe
```

Which would allow us to write tests on each ISO's `.transform_interconnection_queue()` methods without having to hit various external APIs - if we so choose.

## Questions

I think the big question is, do you want to allow people to use GridStatus to access the raw interconnection queue data?

If you think that's a worthwhile goal, what do you think of the plan above? What will go wrong / be hard to maintain? Are you OK with making this a public method?

I'm happy to put in a PR if that seems like the right next step.

jdangerx:
Closes #305.

This adds a `get_raw_interconnection_queue` method to all `ISOBase` subclasses that outputs a `BinaryIO`. This allows users like Catalyst to archive the raw data, including data that we don't know how to interpret within GridStatus yet.

I parametrized the `test_interconnection_queue` test so that a failure in one ISO doesn't fail the whole thing. This helps us isolate test failures. I also added an assertion that the queue data is non-empty to slightly strengthen the tests, and fixed the NYISO column mapping code to handle a change in their data format.

### DIFF
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -416,7 +416,7 @@ class NYISO(ISOBase):
         )
 
         # the spreadsheet doesnt have a date, so make it null
-        completed["Proposed  In-Service"] = None
+        completed["Proposed In-Service"] = None
         completed["Proposed COD"] = None
         # assume it was finished when last updated
         completed["Actual Completion Date"] = completed["Last Updated Date"]
@@ -477,8 +477,8 @@ class NYISO(ISOBase):
             queue["Proposed COD"],
             errors="coerce",
         )
-        queue["Proposed  In-Service"] = pd.to_datetime(
-            queue["Proposed  In-Service"],
+        queue["Proposed In-Service"] = pd.to_datetime(
+            queue["Proposed In-Service"],
             errors="coerce",
         )
         queue["Proposed Initial-Sync Date"] = pd.to_datetime(
@@ -508,7 +508,7 @@ class NYISO(ISOBase):
         }
 
         extra_columns = [
-            "Proposed  In-Service",
+            "Proposed In-Service",
             "Proposed Initial-Sync Date",
             "Last Updated Date",
             "Z",
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -385,7 +385,35 @@ class NYISO(ISOBase):
             and "SGIA Tender Date" not in completed.columns
         ):
             active = active.drop(columns=["SGIA Tender Date"])
-        completed.columns = active.columns
+        completed_colnames_map = {
+            ("Queue", "Pos."): "Queue Pos.",
+            ("Queue", "Owner/Developer"): "Developer Name",
+            ("Queue", "Project Name"): "Project Name",
+            ("Date", "of IR"): "Date of IR",
+            ("SP", "(MW)"): "SP (MW)",
+            ("WP", "(MW)"): "WP (MW)",
+            ("Type/", "Fuel"): "Type/ Fuel",
+            ("Location", "County"): "County",
+            ("Location", "State"): "State",
+            ("Z", "Unnamed: 9_level_1"): "Z",
+            ("Interconnection", "Point"): "Points of Interconnection",
+            ("Interconnection", "Utility "): "Utility",
+            ("Interconnection", "S"): "S",
+            ("Last Update", "Unnamed: 13_level_1"): "Last Updated Date",
+            ("Availability", "of Studies"): "Availability of Studies",
+            ("SGIA Tender Date", ""): "SGIA Tender Date",
+            ("CY Complete Date", ""): "CY Complete Date",
+            ("Proposed Initial-Sync Date", ""): "Proposed Initial-Sync Date",
+            ("Proposed", " In-Service"): "Proposed In-Service Date",
+            ("Proposed", "COD"): "Proposed COD",
+            ("Proposed", "COD.1"): "Proposed COD.1",
+            ("Proposed", "COD.2"): "Proposed COD.2",
+            ("Proposed", "COD.3"): "Proposed COD.3",
+            ("Status", ""): "Status",
+        }
+        completed.columns = completed.columns.to_flat_index().map(
+            lambda c: completed_colnames_map[c],
+        )
 
         # the spreadsheet doesnt have a date, so make it null
         completed["Proposed  In-Service"] = None
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -1,4 +1,8 @@
+import io
+from typing import BinaryIO
+
 import pandas as pd
+import requests
 
 import gridstatus
 from gridstatus import utils
@@ -62,8 +66,9 @@ class NYISO(ISOBase):
                 row["Notes"] = [row["Status"]]
 
                 row["Status"] = row["Status"][
-                    row["Status"].index(STATE_CHANGE)
-                    + len(STATE_CHANGE) : -len(" state.**")
+                    row["Status"].index(STATE_CHANGE) + len(STATE_CHANGE) : -len(
+                        " state.**",
+                    )
                 ].capitalize()
 
             return row
@@ -323,6 +328,16 @@ class NYISO(ISOBase):
 
         return df
 
+    def get_raw_interconnection_queue(self, verbose=False) -> BinaryIO:
+        url = "https://www.nyiso.com/documents/20142/1407078/NYISO-Interconnection-Queue.xlsx"  # noqa
+
+        msg = f"Downloading interconnection queue from {url}"
+        log(msg, verbose)
+        response = requests.get(url)
+        if response.status_code != 200:
+            raise RuntimeError(f"GET {url} failed: {response}")
+        return io.BytesIO(response.content)
+
     def get_interconnection_queue(self, verbose=False):
         """Return NYISO interconnection queue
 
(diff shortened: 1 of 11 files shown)


kmax12:
thanks for getting the ball rolling on this. 

i only took a very quick look, but i suspect the issue has to do with NYISO changing their data format. you may just have to update the nyiso specific code until it works again

kmax12:
better testing for interconnection queue would be appreciated and i suspect would make them more likely to be used by others. i implemented these initially, but havent been using these scrapers on gridstatus.io, so they haven't been maintained like everything else.

jdangerx:
There were a few column name/order changes so I added an explicit remap and NYISO appears to be working again!

In terms of better testing/assertions, I made some incremental improvements (parametrization, asserting non-empty) but don't have the domain knowledge to really make other meaningful improvements right now - that might make sense as a separate PR.

gridstatus/nyiso.py
```diff
@@ -377,7 +386,35 @@ def get_interconnection_queue(self, verbose=False):
...
+            ("Z", "Unnamed: 9_level_1"): "Z",
+            ("Interconnection", "Point"): "Points of Interconnection",
+            ("Interconnection", "Utility "): "Utility",
+            ("Interconnection", "S"): "S",
+            ("Last Update", "Unnamed: 13_level_1"): "Last Updated Date",
+            ("Availability", "of Studies"): "Availability of Studies",
+            ("SGIA Tender Date", ""): "SGIA Tender Date",
+            ("CY Complete Date", ""): "CY Complete Date",
+            ("Proposed Initial-Sync Date", ""): "Proposed Initial-Sync Date",
+            ("Proposed", " In-Service"): "Proposed In-Service Date",
```
jdangerx, line 408: What's the relationship between the `"Proposed In-Service Date"` column and the `"Proposed  In-Service"` column that we create below with two spaces?

pyproject.toml
```diff
@@ -46,7 +46,7 @@ test = [
     "pytest-cov == 4.0.0",
 ]
 dev = [
-    "ruff == 0.0.202",
+    "ruff >= 0.0.291",
```
jdangerx, line 49: VS Code was complaining about `ruff == 0.0.202` being too old for the Ruff extension, so I updated it here - if you have any specific reason to keep it pinned to such an old version, we can definitely do that. Current version is 1.9, FWIW.

gridstatus/nyiso.py
```diff
@@ -63,7 +66,9 @@ def _parse_status(row):
 
                 row["Status"] = row["Status"][
                     row["Status"].index(STATE_CHANGE)
-                    + len(STATE_CHANGE) : -len(" state.**")
+                    + len(STATE_CHANGE) : -len(
```
jdangerx, line 69: Maybe an artifact of updating `ruff`?

gridstatus/nyiso.py
```diff
@@ -377,7 +386,35 @@ def get_interconnection_queue(self, verbose=False):
             and "SGIA Tender Date" not in completed.columns
         ):
             active = active.drop(columns=["SGIA Tender Date"])
-        completed.columns = active.columns
+        completed_colnames_map = {
```
jdangerx, line 389: The original code here assumed that the columns were conceptually the same between `active` and `columns`.

After taking a look I saw two differences:

1. There were three `"Proposed COD.X"` columns in `completed` that were not in `active`.
2. The order of `"Proposed Initial-Sync Date"` and `"Proposed In-Service Date"` were flipped between the two.

So I made the mapping explicit.

jdangerx:
@kmax12 I made a bunch of comments but I think this is pretty much ready for review!

kmax12:
Thanks @jdangerx. I’ll review as I have time. Might take a week or so but will try to be faster

gridstatus/nyiso.py
kmax12: must be

gridstatus/tests/test_interconnection_queue.py
```diff
@@ -5,6 +5,9 @@
 
 # todo implement this
 @pytest.mark.slow
-def test_get_interconnection_queue_all():
+@pytest.mark.parametrize("iso_class", gridstatus.utils.all_isos)
+def test_get_interconnection_queue_all(iso_class):
```
kmax12, line 9: we recently added `IESO` to [`all_isos`](https://github.com/kmax12/gridstatus/blob/c12055616f3da4e34ddde181469cd8d976133a8e/gridstatus/__init__.py#L31). there is an interconnection queue [here](https://www.ieso.ca/en/Sector-Participants/Connection-Process/Application-Status), but we haven't implemented it. let's just make sure this test can handle ieso not being implemented

gridstatus/nyiso.py
kmax12: i think the double space is a typo! you can fix that

if i remember my own intentions, the goal was to just remove the "I/S" or "IS" values so the type of the column would be a datetime. have the string would cause a mixed typed

kmax12 (commented):
sorry for the long delay on this. this is a nice refactor to make the raw queue data available for archiving. 

just a tiny bit left on nyiso and we can merge this. 

thanks again!

jdangerx:
I fixed the double space typo and updated the tests to not fail on IESO. I'm not quite sure if you want me to merge the `Proposed In-Service` and `Proposed In-Service Date` columns or not, but I can do that and strip out the `I/S` and `IS` values if that is the right thing to do!

kmax12:
> I fixed the double space typo and updated the tests to not fail on IESO. I'm not quite sure if you want me to merge the `Proposed In-Service` and `Proposed In-Service Date` columns or not, but I can do that and strip out the `I/S` and `IS` values if that is the right thing to do!

yes, i think they can be merged. i think the `I/S` value that shows up for completed projects is just saying that they are in fact "in service". this column can be null for completed project

gridstatus/nyiso.py
```diff
@@ -441,8 +478,8 @@ def get_interconnection_queue(self, verbose=False):
             queue["Proposed COD"],
             errors="coerce",
         )
-        queue["Proposed  In-Service"] = pd.to_datetime(
-            queue["Proposed  In-Service"],
+        queue["Proposed In-Service"] = pd.to_datetime(
```
kmax12, line 481: should this be `"Proposed In-Service Date"` since that is the value in `active`?

gridstatus/nyiso.py
```diff
@@ -472,7 +509,7 @@ def get_interconnection_queue(self, verbose=False):
         }
 
         extra_columns = [
-            "Proposed  In-Service",
+            "Proposed In-Service",
```
kmax12, line 512: same comment as above

should this be "Proposed In-Service Date" since that is the value in active?

gridstatus/nyiso.py
WillKoehrsen: Yeah, it sounds like others have noticed issues between the active sheet and other sheets so thanks for taking care of this (#323).

WillKoehrsen (approved):
@jdangerx The tests pass and all the changes look good. Is this ready to merge?

jdangerx:
Yeah, it's ready for merge! There was a merge conflict which I just dealt with - hopefully CI continues to pass 😅

WillKoehrsen:
@jdangerx It looks like there's a linting issue. Running `make lint-fix` then committing changes should solve that.

WillKoehrsen:
It looks like there is [one test failure](https://github.com/kmax12/gridstatus/actions/runs/7788746640/job/21239283143?pr=307#step:6:2458) on `TestNYISO.test_get_interconnection_queue_handles_new_file`. That was a test I added in #329 to handle new NYISO interconnection data. I think we just need to make sure the `"Interconnection Location"` column is handled correctly for NYISO.

jdangerx:
Ah, yeah, looks like you renamed `Points of Interconnection` in the active sheet, which propagated downstream in the old code but needs to be manually updated for the completed sheet now. I think everything should pass now 🤞 Thanks for your patience!

WillKoehrsen:
All the tests pass now, so I'll go ahead and merge. Thanks @jdangerx!

PR #312, opened 2024-01-10
Introduce ERCOT API Wrapper

djsensei:
Enables users to pull data from all of the endpoints of the ERCOT API. Provides three primary methods in ercot_api.py:
1) `hit_ercot_api` - prepares and executes a request to a specified endpoint
2) `list_all_endpoints` - enumerates all available endpoints with simple descriptions
3) `describe_one_endpoint` - lists the parameters of an endpoint with expected types

The latter two methods are conveniently available via CLI, but `hit_ercot_api` is not, because it returns a dataframe and is thus expected to be used within some outer workflow. 

*WIP - TODOs*
- [x] Design review
- [x] Unit tests
- [x] Develop CLI
- [x] Example notebook


gridstatus/ercot_api/README.md
```diff
@@ -0,0 +1,52 @@
+## ERCOT API
+
+A module to enable access and usage of [ERCOT's API](https://apiexplorer.ercot.com/)
+
+### Module Contents
+
+*pubapi-apim-api.json* was downloaded from https://apiexplorer.ercot.com/api-details#api=pubapi-apim-api by selecting Open API 3 (JSON) from the "API Definition" dropdown. It contains metadata about all of the available endpoints in the ERCOT API.
```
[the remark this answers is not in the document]
djsensei: I'm not actually sure how to programatically download it. I stumbled into it by a pleasant accident (as noted in the README). But I'm sure that would be easy enough to figure out. My thoughts on how to best maintain it:
1) The file lives in an awkward middle ground of size - if it were bigger, clearly makes sense to minimize re-downloading, but if it were smaller, could make sense to download each time to ensure freshness.
2) I don't know yet how frequently the API will be updated, and therefore how best to design around those updates. (this has implications on unit test design too). 
3) Therefore, I think it makes sense to leave it fairly simple for now, and be ready to make improvements if and when the API itself evolves

WillKoehrsen:
@djsensei Thanks for your contribution! It looks like there are some [linting issues with the build.](https://github.com/kmax12/gridstatus/actions/runs/7578020698/job/20796977271?pr=312) You should be able to fix these by running `make lint` and addressing the issues.

djsensei:
Thanks @WillKoehrsen, should be all set now pending CI!

kmax12:
Just triggered CI to run. Will do another review later today with a eye towards getting this merged. Thanks for the diligence and patience on this!

gridstatus/ercot_api/pubapi-apim-api.json
WillKoehrsen, line 1: I'm not sure what our general stance towards storing data in the repo is. Even if the file doesn't change, I think we still might want to download it rather than have the file in the repo.

WillKoehrsen: Ahh, I see there's already a conversation on this point.

docs/Examples/ercot_api/ERCOT API Examples.ipynb
WillKoehrsen, line 1: I'm running into an error when running this notebook locally. In cell 4:

```python
hit_ercot_api(agg_gen_summary_endpoint)
```

Output

```
Paginating results:   0%|                                 | 0/1 [00:00<?, ?it/s]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Input In [4], in <cell line: 0>()
----> 1 hit_ercot_api(agg_gen_summary_endpoint)

File ~/git/gridstatus/gridstatus/ercot_api/ercot_api.py:76, in hit_ercot_api(endpoint, page_size, max_pages, **api_params)
     73 if columns is None:
     74     # first request only: populate columns and update total pages
     75     columns = [f["name"] for f in response["fields"]]
---> 76     total_pages = response["_meta"]["totalPages"]
     77     # determine number-of-pages denominator for progress bar
     78     if max_pages is None:

KeyError: '_meta'
```

and the same error in cell 8:

```python
hit_ercot_api(spp_endpoint, max_pages=10)
```

djsensei: Hrmm, I'm getting the same thing. I made [a related change a few commits ago](https://github.com/kmax12/gridstatus/pull/312/commits/9aa227ce331c201b738ae818b97a7321396d2647) because what had worked before no longer did, so this is curious behavior. Investigating...

WillKoehrsen: The API specification says you _may_ get back a `"_meta"` property, but it's not required. 

Screenshot 2024-01-29 at 13 24 15

When I manually retrieve one of the reports, there's no `"_meta"` property and also no `"data"`.

```python
In [7]: r = requests.get('https://api.ercot.com/api/public-reports/np6-345-cd/act_sys_load_by_wzn')
Out[7]: <Response [200]>

In [8]: r.json().keys()
Out[8]: dict_keys(['report', 'fields', '_links'])
```

So maybe the API docs are incorrect/outdated?

djsensei: I think you may be right about `maybe the API docs are incorrect/outdated?`. I get those same three fields in a response to the first notebook query ("/np3-910-er/2d_agg_gen_summary", no params), but on their [api reference website ](https://apiexplorer.ercot.com/api-details#api=pubapi-apim-api&operation=getData_97) it's "links" not "_links"

If I add some parameters (like the second notebook example) the "_meta" and "data" fields return and the code works as-is. Perhaps it's reasonable to assume that those two fields come as a pair, and if either is missing they both will be missing (and the query returns zero results). Would be great if they would tell you _why_ there are zero results, but explanation or not that probably varies from endpoint to endpoint in a way that falls out of the scope of this PR.

djsensei: Alright, I have a safety valve fix for when no data is returned, but I'll need to update the notebook so the first example isn't just an empty result set...

djsensei: > That's interesting; I wondered if the problem had something to do with not setting parameters making the dataset empty. This behavior is not documented in the API's docs since none of the parameters are marked as "required". 🤷🏽

Not only that, but it worked fine a few weeks ago with no params, returning what I assume is the full set of results.

WillKoehrsen: Hmm, I wonder if there could also be a timing issue. Like the dataset is published at a certain time each day...

Anyway, if you want to commit your updates, I'll try to review them today.

djsensei: Commits inbound!

More curiosity: I wrote a little test script to try and see which endpoints returned values even when given no params, and ended up stumbling upon a weird observation on the actual system load by weather zone endpoint ("/np6-345-cd/act_sys_load_by_wzn"). `page_size` counts as an input, even though it doesn't have anything to do with the contents of the results. 

```
hit_ercot_api("/np6-345-cd/act_sys_load_by_wzn") -> no results
hit_ercot_api("/np6-345-cd/act_sys_load_by_wzn", max_pages=1) -> no results
hit_ercot_api("/np6-345-cd/act_sys_load_by_wzn", page_size=10) -> results!
hit_ercot_api("/np6-345-cd/act_sys_load_by_wzn", page_size=10, max_pages=1) -> results!
```

So I've updated the `page_size` argument to have a default value of 1000, and always be passed as a query parameter, and that seems to trigger the API to give us results again 🧐. Bonus feature: didn't have to change the notebook after all!

WillKoehrsen (commented):
Thanks for the contribution @djsensei! I'm good with this PR but will let @kmax12 make the final approval.

kmax12:
im seeing a test error when i run on github actions: https://github.com/kmax12/gridstatus/actions/runs/7703617024/job/20996079651?pr=312#step:7:2391

once that is cleaned up, im good to merge.

djsensei:
The test failure (at least the one on my new test suite) seems non-deterministic, haven't been able to replicate it. Maybe something to do with hitting the API at the exact time that data is refreshed, since it was supposed to return 48 rows (two days of hourly) and we only got 24?

In an ideal world we wouldn't be hitting an external API from a test suite, but since that's the whole point of this module...

WillKoehrsen:
> seems non-deterministic, haven't been able to replicate it. 

The test suite [re-runs failed tests up to 5 times](https://github.com/kmax12/gridstatus/blob/main/Makefile#L10) so if it was just a temporary issue, I think it should have passed. 

A common issue is tests that pass locally when run in the developer's time zone but then fail because GitHub actions run in UTC. That seems likely to be the issue in this case since the tests also pass locally for me.

I think the issue is this line:

```python
two_days_ago = datetime.datetime.now() - datetime.timedelta(days=2)
```

Locally, right now, this returns ` datetime.datetime(2024, 1, 27, 20, 12, 13, 36715)`. 

When run with a timezone of `UTC`:

```python
two_days_ago = datetime.datetime.now(tz=pytz.UTC) - datetime.timedelta(days=2)
```

The date changes: `datetime.datetime(2024, 1, 28, 1, 12, 30, 32344, tzinfo=<UTC>)`

WillKoehrsen:
Just leaving this here in case we want to take a programmatic approach to retrieving the API documentation in a follow-on PR. I figured out how to download the documentation by inspecting the network requests:

```python
import requests

# URL for the documentation
api_doc_url = "https://apiexplorer.ercot.com/developer/apis/pubapi-apim-api?export=true&api-version=2022-04-01-preview"

# Not sure how many of these are required, but they are the same
# headers as in a browser request
headers = {
    'authority': 'apiexplorer.ercot.com',
    'accept': 'application/vnd.oai.openapi+json',
    'accept-language': 'en-US,en;q=0.9',
    'cache-control': 'no-cache',
    'dnt': '1',
    'pragma': 'no-cache',
    'referer': 'https://apiexplorer.ercot.com/api-details',
    'sec-ch-ua': '"Not A(Brand";v="99", "Google Chrome";v="121", "Chromium";v="121"',
    'sec-ch-ua-mobile': '?0',
    'sec-ch-ua-platform': '"macOS"',
    'sec-fetch-dest': 'empty',
    'sec-fetch-mode': 'cors',
    'sec-fetch-site': 'same-origin',
    'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36',
    'x-ms-apim-client': 'dev-portal|apiexplorer.ercot.com|exportApi-unauthorized'
}

api_response = requests.get(api_doc_url, headers=headers)

api_documentation = api_response.json()
print(api_documentation)
```

It doesn't work without those headers.

WillKoehrsen (approved):
👍🏽 The only failing part of the build is code coverage for the changes, but I think the tests are adequate for the new functionality.

PR #313, opened 2024-01-14
Switch SPP weis to use interval data source

kmax12:


gridstatus/spp.py
```diff
@@ -672,20 +672,26 @@ def _finalize_spp_df(self, df, market, location_type, verbose=False):
...
+            date: date to get data for. if end is not provided, will get data for
+                5 minute interval that date is in.
+            end: end date
+            verbose: print url
         """
+        # if no end, find nearest 5 minute interval end
+        # to use
+        if end is None:
+            # round date up to nearest 5 minutes
+            end = date + pd.Timedelta(minutes=5) - pd.Timedelta(
```
WillKoehrsen, line 689: We might want to handle the case where `date="latest"`. Currently we get an error here.

gridstatus/decorators.py
```diff
@@ -387,3 +390,12 @@ def __ladd__(self, other):
 
     def __radd__(self, other):
         return self.__ladd__(other)
+
+class FiveMinOffset:
+    def __ladd__(self, other):
+        # round up to next 5 min interval
+        # remove seconds and microseconds
+        return other.replace(second=0, microsecond=0) + pd.DateOffset(minutes=5) - pd.DateOffset(minutes=other.minute % 5)
```
WillKoehrsen, line 398: You can round up a datetime/Timestamp to the nearest 5 min with `.ceil("5min")`

gridstatus/spp.py
WillKoehrsen: We could add the following before the `if end is None` conditional:

```py
if date == "latest":
    date = pd.Timestamp.now(tz=self.default_timezone).ceil("5min")
```

gridstatus/decorators.py
kmax12: Cool I'll do that in the next iteration just stepped away from my computer

gridstatus/decorators.py
```diff
@@ -387,3 +390,13 @@ def __ladd__(self, other):
 
     def __radd__(self, other):
         return self.__ladd__(other)
+
+class FiveMinOffset:
+    def __ladd__(self, other):
+        # round up to next 5 min interval
+        # add 1 microsecond to ensure we make it to the 
+        # next interval when already on a 5 min interval
+        return (other + pd.Timedelta(microseconds=1)).ceil("5min")
```
kmax12, line 399: @WillKoehrsen to use ceil i needed to add 1 micro second since 

```python
>>> pd.Timestamp('2024-01-14 12:05:00').ceil("5Min")
Timestamp('2024-01-14 12:05:00')
```

gridstatus/spp.py
kmax12: there is a specific url for latest, so i implemented that

WillKoehrsen (approved):
Looks good. There are a couple of linting issues that ruff is complaining about.

PR #355, opened 2024-03-10
Fix SPP DST Issues

kmax12:
Our usage of pandas ceiling wasn't DST friendly. Updated logic. This should resolve issues iterating over 5 minute files

### DIFF
--- a/gridstatus/decorators.py
+++ b/gridstatus/decorators.py
@@ -400,10 +400,21 @@ class MonthBeginOffset:
 
 class FiveMinOffset:
     def __ladd__(self, other):
-        # round up to next 5 min interval
-        # add 1 microsecond to ensure we make it to the
+        # Store the original timezone
+        original_tz = other.tz
+
+        # Convert to UTC to avoid DST issues
+        other_utc = other.tz_convert("UTC")
+
+        # Round up to the next 5 min interval
+        # Add 1 microsecond to ensure we make it to the
         # next interval when already on a 5 min interval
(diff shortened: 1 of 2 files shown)


gridstatus/decorators.py
```diff
@@ -400,10 +400,21 @@ def __radd__(self, other):
...
+        # Store the original timezone
+        original_tz = other.tz
+
+        # Convert to UTC to avoid DST issues
+        other_utc = other.tz_convert("UTC")
+
+        # Round up to the next 5 min interval
+        # Add 1 microsecond to ensure we make it to the
         # next interval when already on a 5 min interval
-        return (other + pd.Timedelta(microseconds=1)).ceil("5min")
```
kmax12, line 406: this would round up to a non existent time

gridstatus/tests/test_decorators.py
```diff
@@ -0,0 +1,11 @@
+import pandas as pd
+
+from gridstatus.decorators import FiveMinOffset
+
+# todo test other offsets
+
+
+def test_five_min_offset():
```
WillKoehrsen, line 8: Do we also want to add a test during the end of DST? I don't see why the logic would fail, but a test wouldn't hurt.

WillKoehrsen (approved):

gridstatus/tests/test_decorators.py
kmax12: good idea. added.

however, what this made me realize is that the code actually doesn't handle the repeated hour in the case of SPP. we need to add `d` to the file name in those cases. going to create follow up issue

image

PR #360, opened 2024-03-14
Enhances EIA functionality

samgdotson:
This PR enhances the EIA class functionality by 
* adding a `list_facets` function to help data discovery.
  * Does not duplicate existing functionality
  * Handles exceptions
* adds a `_facet_handler` function to ensure facets are passed in a `list` (helpful for cases where users desire a single facet type). 
* in `EIA.get_dataset`
  * Apply `_facet_handler`
  * Add `frequency` parameter with default value `hourly`, updated the docstring
  * updates docstring to reflect new handlers
* [Bug-fix] in `_handle_region_data` checks that the dataframe has a column, `col` before casting series values to float.
* Adds handler for `rto/region-sub-ba-data` 
* Adds test for new handlers.

Unfortunately, I have not run the tests because I'm working from a windows machine. But the new test follows the pattern of other tests. Please let me know if I need to change anything in order to merge this PR!

Edit (2024-03-15): I have installed `poetry` and run the new tests, which pass locally.

### DIFF
--- a/gridstatus/tests/test_eia.py
+++ b/gridstatus/tests/test_eia.py
@@ -37,8 +37,8 @@ def _check_region_data(df):
     assert df["Interval End"].dtype == "datetime64[ns, UTC]"
     assert df.shape[0] > 0
     assert df.columns.tolist() == columns
-    
-    
+
+
 def _check_region_subba_data(df):
     columns = [
         "Interval Start",
@@ -47,7 +47,7 @@ def _check_region_subba_data(df):
         "BA Name",
         "Subregion",
         "Subregion Name",
-        "MW"
+        "MW",
     ]
 
     assert df["Interval Start"].dtype == "datetime64[ns, UTC]"
@@ -90,13 +90,13 @@ def test_list_routes():
     routes = eia.list_routes("electricity/rto/")
 
     assert "interchange-data" in [r["id"] for r in routes["routes"]]
-    
+
 
 def test_list_facets():
     eia = gridstatus.EIA()
-    
+
     facets = eia.list_facets("electricity/rto/region-data")
-    
+
     assert "type" in facets.keys()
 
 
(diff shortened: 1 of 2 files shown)


WillKoehrsen:
Hey @samgdotson thanks for your contribution! It looks like the linting check is failing. Are you able to run linting, either with `make lint-fix` or `poetry run ruff gridstatus/ --fix` then `poetry run black gridstatus/`?

gridstatus/tests/test_eia.py
```diff
@@ -37,6 +37,23 @@ def _check_region_data(df):
     assert df["Interval End"].dtype == "datetime64[ns, UTC]"
     assert df.shape[0] > 0
     assert df.columns.tolist() == columns
+    
+    
+def _check_region_subba_data(df):
```
WillKoehrsen, line 42: In Pytest, only method names starting with `test_` are interpreted as tests. The other methods in this file that start with `_check` are called by `test_` methods. This method should be renamed to something like `test_region_sub_ba_data`. 

https://docs.pytest.org/en/6.2.x/goodpractices.html#test-discovery

WillKoehrsen: Btw, you said you couldn't run the tests locally because you're on Windows...have you tried using `poetry` to run the tests? You can try `poetry run pytest gridstatus/tests/test_eia.py ` to run just the `eia` tests.

samgdotson: Forgetting to actually add the `test_` was a total goof on my part! I was distracted yesterday. Also, this is my first time using `poetry` so it's a little unfamiliar! I was able to run the tests and they all pass locally.

samgdotson:
Thanks for the quick review @WillKoehrsen. My apologies for not fully reading the `CONTRIB` guide.

Updates to this PR:
* runs blackify 
* adds `test_region_subba_data` to *actually* test the new functions.
* adds a `test_list_facets` to test the *other* new function, which wasn't previously tested.
* changes the `start` and `end` variables for a few tests in `test_eia.py` to provide a fixed and consistent time period.

Regarding that last point, there were a few tests that depended on `pd.Timestamp.now()` AND expect a dataset with no null values. In my view, this is undesirable because a reporting entity that has no nulls today might have nulls tomorrow (evinced by currently failing tests!). Let me know if you'd like this change reverted.

WillKoehrsen (approved):
👍🏽 Everything works well locally, and the tests pass.

PR #361, opened 2024-03-16
ERCOT API Historical Data Retrieval

WillKoehrsen:
- Adds the ability to retrieve historical data (more than 90 days in the past) from the ERCOT API

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -36,6 +36,14 @@ SHADOW_PRICES_SCED_ENDPOINT = "/np6-86-cd/shdw_prices_bnd_trns_const"
 # How long a token lasts for before needing to be refreshed
 TOKEN_EXPIRATION_SECONDS = 3600
 
+# Number of historical links to fetch at once. The max is 1_000
+DEFAULT_HISTORICAL_SIZE = 1_000
+
+# Number of days in past when we should use the historical method
+# NOTE: this seems to vary per dataset. If you get an error about no data available and
+# your date is within this number of the current date, try decreasing this number.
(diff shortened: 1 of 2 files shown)


WillKoehrsen:
None of the test failures are related to the `ercot_api` because we don't run those tests in GitHub actions. They do pass running locally.

gridstatus/ercot_api/ercot_api.py
```diff
@@ -351,6 +395,108 @@ def _shadow_prices_column_name_mapper(self):
...
+                # Data comes back as a compressed zip file.
+                response = self.make_api_call(link, verbose=verbose, parse_json=False)
+
+                # Convert the bytes to a file-like object
+                bytes = pd.io.common.BytesIO(response)
+
+                # Decompress the zip file and read the csv
+                dfs.append(pd.read_csv(bytes, compression="zip"))
+
+                time.sleep(0.2)  # Necessary to avoid rate limiting
```
kmax12, line 443: not a big deal, but i like to make this a parameter `get_historical_data`, so caller can tweak themselves to avoid rate limit

kmax12 (approved):
nice work on this. looks great

gridstatus/ercot_api/ercot_api.py
```diff
@@ -351,6 +396,125 @@ def _shadow_prices_column_name_mapper(self):
             "ViolationAmount": "Violation Amount",
         }
 
+    def get_historical_data(
+        self,
+        endpoint,
+        start_date,
+        end_date,
+        sleep_seconds=0.1,
```
WillKoehrsen, line 404: Updated to make the sleep time a parameter.

PR #368, opened 2024-03-27
Return only latest version of pjm LMP

kmax12:


gridstatus/pjm.py
```diff
@@ -757,6 +757,9 @@ def get_lmp(
                 ),
             )
 
+        # returns on the latest version of the data
```
WillKoehrsen, line 760: Does PJM issue corrections for data?

WillKoehrsen (approved):

gridstatus/pjm.py
kmax12: i think that is what happened here. they do keep older versions so I think we can backfill corrections in the future if needed

PR #374, opened 2024-04-04
MISO Load Forecast Update

WillKoehrsen:
- Updates MISO `get_load_forecast` to use a new source that includes past dates
- Renames `Forecast Time` to `Publish Time`

### DIFF
--- a/gridstatus/tests/test_miso.py
+++ b/gridstatus/tests/test_miso.py
@@ -94,7 +94,7 @@ class TestMISO(BaseTestISO):
         "MISO MTLF (MWh)",
     ]
 
-    def test_get_load_forecast_today_or_latest(self):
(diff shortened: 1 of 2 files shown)


gridstatus/tests/test_miso.py
```diff
@@ -81,13 +81,75 @@ def test_get_load_historical_with_date_range(self):
 
     """get_load_forecast"""
 
+    load_forecast_cols = [
+        "Interval Start",
+        "Interval End",
+        "Publish Time",
+        "LRZ1 MTLF (MWh)",
```
kmax12, line 88: i think we can get rid of the `(MWh)`. we dont include that in any other dataset

kmax12 (approved):

PR #375, opened 2024-04-04
Add XLRD Dependency

WillKoehrsen:
- Adds `xlrd` as a dependency which is required for reading `.xls` files.


pyproject.toml
```diff
@@ -36,6 +36,7 @@ tabula-py = "^2.5.1"
 lxml = "^4.9.1"
 plotly = "^5.11.0"
 termcolor = "^2.2.0"
+xlrd = "^2.0.1"
```
WillKoehrsen, line 39: Forgot to commit this as part of #374.

kmax12 (approved):

kmax12:
don't we already read xls files in other places? why did we need to add?

kmax12:
there is also the `calamine` engine which is somewhat new to pandas and i saw somewhere it's much more performant 

https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#io-calamine. 

probably for another PR though

WillKoehrsen:
Prior to #374, the library only read `xlsx` files which use `openpyxl` as the `engine`. For `xls` files, the `engine` is `xlrd` (or maybe `calamine` which isn't installed).

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_excel.html#pandas.read_excel

PR #378, opened 2024-04-15
PJM Historical Load Forecast

WillKoehrsen:
- Adds historical load forecast for PJM using https://dataminer2.pjm.com/feed/load_frcstd_hist/definition
- This historical forecast includes all vintages (10 forecasts for each interval)

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -8,7 +8,7 @@ import requests
 import tqdm
 
 from gridstatus import utils
-from gridstatus.base import ISOBase, Markets, NoDataFoundException, NotSupported
+from gridstatus.base import ISOBase, Markets, NoDataFoundException
 from gridstatus.decorators import (
     _get_pjm_archive_date,
     pjm_update_dates,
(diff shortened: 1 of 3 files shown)


gridstatus/pjm.py
```diff
@@ -554,25 +554,49 @@ def get_load_forecast(self, date, verbose=False):
             ),
             "forecast_area": "RTO_COMBINED",
```
kmax12, line 555: while we're in this method, i think we should update it to get all forecast areas. to be consistent with how we do load, I think we should pivot on `forecast_area` make them each unique columns

kmax12: im noticing tht the 7 day forecast has different regions than the historical

image

vs 

image

WillKoehrsen: That's a good callout. I only checked that the load forecast for `RTO` in the historical data from the most recent vintage was ≈ `RTO_COMBINED` in the seven day forecast.

@cwaldoch Do you know why PJM would have different forecast areas in the two datasets?

WillKoehrsen: > I think we should pivot on forecast_area make them each unique columns

IMO it's better to have a `forecast_area` column instead of making each area a column if you are storing the data because otherwise you have to update your table every time the forecast areas change. It depends on the use case though.

kmax12: You're right about it being more flexible format.

I don't think forecast areas change very often, which I why I did it this for load. I think for end users the columns is an easier to use format too. I mainly was trying to be consistent with how we do the load. 

I'm okay if we want a forecast_area column in this case.

kmax12: The same areas also named differently across the load forecast datasets

WillKoehrsen:
This has been updated to separate `get_load_forecast_historical` into its own function and `get_load_forecast` only supports the current forecast.

WillKoehrsen:
@kmax12 Any further changes needed?

kmax12 (approved):
LGTM

PR #381, opened 2024-05-03
add gen_outages_by_type dataset

vel9:


WillKoehrsen:
This looks good! Running `test_pjm` locally, there is one test failure, `test_get_interconnection_queue`, but I think that was failing before.

gridstatus/pjm.py
```diff
@@ -899,6 +899,9 @@ def _get_pjm_json(
                 end = utils._handle_date(end)
             else:
                 end = start + pd.DateOffset(days=1)
+                # the end is inclusive, pulls records
+                # for next day without below adjustment
+                end = end - pd.DateOffset(seconds=1)
```
WillKoehrsen, line 904: Something I'm not sure about is when to use `DateOffset` vs `Timedelta` (used later in this PR). From [this comment](https://github.com/pandas-dev/pandas/issues/32026#issuecomment-586631368) by a Pandas maintainer, I think we should be using `DateOffset` but I'm interested in hearing other opinions.

WillKoehrsen: We don't need to change anything in this PR, just something to think about.

WillKoehrsen (approved):
Looks good! Could you add a bullet in `CHANGELOG.md` about this addition?

gridstatus/pjm.py
vel9: Good point and thank you for the comment link. Seeing this for our use case, so I do think we should switch to Timedelta:
```
ts = pd.Timestamp('2024-11-03 00:00:00', tz='US/Eastern')
print(ts + pd.Timedelta(days=1)) # 2024-11-03 23:00:00-05:00
print(ts + pd.DateOffset(days=1)) # 2024-11-04 00:00:00-05:00
```

PR #382, opened 2024-05-24
Add Short Term System Adequacy

WillKoehrsen:
- Adds short term system adequacy from https://www.ercot.com/mp/data-products/data-product-details?id=NP3-763-CD


kmax12 (approved):
LGTM

PR #384, opened 2024-06-03
Add ERCOT Temperature Forecasts by Weather Zone

WillKoehrsen:
- Add hourly temperature forecasts by weather zone for ERCOT
- Data from https://www.ercot.com/mp/data-products/data-product-details?id=NP4-722-CD

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -101,6 +101,9 @@ SYSTEM_WIDE_ACTUALS_RTID = 12340
 # https://www.ercot.com/mp/data-products/data-product-details?id=NP3-763-CD
 SHORT_TERM_SYSTEM_ADEQUACY_REPORT_RTID = 12315
 
+# https://www.ercot.com/mp/data-products/data-product-details?id=NP4-722-CD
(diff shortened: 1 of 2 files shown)


gridstatus/tests/test_ercot.py
```diff
@@ -1142,7 +1142,7 @@ def _check_short_term_system_adequacy(self, df):
             df["Interval End"] - df["Interval Start"] == pd.Timedelta(hours=1)
         ).all()
 
-    def test_get_short_term_system_adequacy_today_or_latest(self):
```
WillKoehrsen, line 1145: Method name was incorrect because this only tests for "today".

gridstatus/ercot.py
```diff
@@ -2904,6 +2956,13 @@ def parse_doc(
 
         return doc
 
+    def _weather_zone_column_name_mapping(self):
```
WillKoehrsen, line 2959: Using the same column naming for `load_forecast_by_weather_zone` and `temperature_forecast_by_weather_zone`.

kmax12 (approved):

PR #390, opened 2024-06-17
MISO Weekly LMP 5 Minute Data

WillKoehrsen:
- Adds `MISO().get_lmp_weekly` to get the weekly LMP data that includes corrections to the real-time data
- Retrieves data from https://www.misoenergy.org/markets-and-operations/real-time--market-data/market-reports/#nt=%2FMarketReportType%3AHistorical%20LMP%2FMarketReportName%3AWeekly%20Real-Time%205-Min%20LMP%20(zip)&t=500&p=0&s=MarketReportPublished&sd=desc


gridstatus/tests/test_miso.py
WillKoehrsen, line 1: Run the specific tests with ` pytest gridstatus/tests/test_miso.py -k test_get_lmp_weekly_historical`

gridstatus/miso.py
```diff
@@ -261,6 +261,74 @@ def _get_solar_and_wind_forecast_data(self, date, fuel, verbose=False):
...
+
+        df = df.merge(
+            node_to_type,
+            left_on="Location",
+            right_on="Node",
+            how="left",
+        )
+
+        df["Energy"] = df["LMP"] - df["Loss"] - df["Congestion"]
+        df["Market"] = Markets.REAL_TIME_5_MIN.value
```
WillKoehrsen, line 313: Should we add a new `REAL_TIME_5_MIN_WEEKLY` value to differentiate?

kmax12: yes, we should. 

in general I wish every lmp dataset was it's own function like we do for other datasets.

WillKoehrsen: Updated. Any further changes before this can be merged?

kmax12 (approved):

PR #393, opened 2024-06-20
ISONE LMP Real Time 5 Min Handle Current Date

WillKoehrsen:
- Makes querying for the current date for ISONE lmp real time 5 min more efficient by filtering the intervals queried for

### DIFF
--- a/gridstatus/isone.py
+++ b/gridstatus/isone.py
@@ -1,5 +1,4 @@
 import io
-import math
 from typing import BinaryIO
 
 import pandas as pd
@@ -34,6 +33,8 @@ class ISONE(ISOBase):
         Markets.REAL_TIME_HOURLY,
         Markets.DAY_AHEAD_HOURLY,
     ]
+
+    lmp_real_time_intervals = ["00-04", "04-08", "08-12", "12-16", "16-20", "20-24"]
 
     hubs = {"H.INTERNAL_HUB": 4000}
     zones = {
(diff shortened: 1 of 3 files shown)


gridstatus/isone.py
```diff
@@ -406,40 +408,37 @@ def get_lmp(
...
-                dfs.append(
-                    pd.read_csv(
-                        u,
-                        skiprows=[0, 1, 2, 3, 5],
-                        skipfooter=1,
-                        engine="python",
-                    ),
-                )
+                try:
+                    dfs.append(
```
WillKoehrsen, line 426: This needs to be a `try...except` because the interval file may not be published for the provided `date` and `end`.

gridstatus/isone.py
```diff
@@ -817,6 +811,51 @@ def _create_interval_start_from_hour_start(self, data):
...
+                raise ValueError("Date and end must be on the same day")
+
+        # Convert intervals to hour ranges
+        interval_hours = [
+            (int(interval.split("-")[0]), int(interval.split("-")[1]))
+            for interval in intervals
+        ]
+
+        if not end:
+            end = date.replace(hour=23, minute=59, second=59, microsecond=0)
```
WillKoehrsen, line 840: If this method is only called on th current date, then I think this should be instead set to the current time.

gridstatus/isone.py
```diff
@@ -406,40 +408,37 @@ def get_lmp(
...
-            # todo handle intervals for current day
-            intervals = ["00-04", "04-08", "08-12", "12-16", "16-20", "20-24"]
+            intervals = self.lmp_real_time_intervals[:]
 
-            # optimze for current day
             if now.date() == date.date():
-                hour = now.hour
-                # select completed 4 hour intervals based on current hour
-                intervals = intervals[: math.ceil((hour + 1) / 4) - 1]
+                intervals = self._filter_intervals_in_range(
```
kmax12, line 414: i sort of prefer the current logic that converts the requested time into the index of the interval array rather than take the names of the intervals and converting them to datetime objects.

i think the logic would be something like, which strikes me as a lot easier to maintain. i didn't run this so there could be off by one errors. 

```python
# converts hour into the index 
# in 4 hour interval array
def _hour_to_interval_index(hour):
    if not 0 <= hour <= 23:
        raise ValueError("Hour must be between 0 and 23")
    return hour // 4

# no need to get data for future intervals
last_interval_hour = min(now.hour, end.hour) if end else now.hour
intervals = intervals[_hour_to_interval_index(date.hour) : _hour_to_interval_index(last_interval_hour)+1]
```

kmax12 (approved):

gridstatus/isone.py
kmax12: you might need to change my logic to only include the completed intervals. I think that is why we didn't previously need the `try...except` you added

WillKoehrsen: Do you know how long after the interval is completed the files are available? I think we may want to keep the `try...except`.

kmax12: i dont not. im find with that approach

WillKoehrsen: Updated to use your simpler logic. The `try` has been kept.

PR #394, opened 2024-06-24
Fixes ISONE LMP Real Time 5 Min When No Intervals are Selected

WillKoehrsen:
- Fixes the ISONE LMP Real Time 5 Minute data fetching when no intervals are selected by still getting current data
- This can happen when the `date` and `end` are < 4 hours apart

### DIFF
--- a/gridstatus/isone.py
+++ b/gridstatus/isone.py
@@ -392,7 +392,6 @@ class ISONE(ISOBase):
         Find Node ID mapping:
             https://www.iso-ne.com/markets-operations/settlements/pricing-node-tables/
         """  # noqa
-
         if date == "latest":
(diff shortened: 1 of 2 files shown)


gridstatus/isone.py
```diff
@@ -451,10 +457,15 @@ def get_lmp(
...
-                ]
-                data = pd.concat([data, data_current])
+
+                if data:
+                    data_current = data_current[
+                        data_current["Local Time"] > data["Local Time"].max()
+                    ]
+
+                    data = pd.concat([data, data_current])
+                else:
```
WillKoehrsen, line 467: Need to write a test for this case (no intervals selected).

kmax12 (approved):

PR #395, opened 2024-06-26
Fix IESO Load Forecast by Not Verifying

WillKoehrsen:
- Fixes IESO `get_load_forecast` by not verifying SSL

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -822,7 +822,7 @@ class IESO(ISOBase):
         sleep = 5
 
         while retry_num < max_retries:
-            r = requests.get(url)
+            r = requests.get(url, verify=False)
 
             if r.ok:
                 break


gridstatus/ieso.py
```diff
@@ -822,7 +822,7 @@ def _request(self, url, verbose):
         sleep = 5
 
         while retry_num < max_retries:
-            r = requests.get(url)
+            r = requests.get(url, verify=False)
```
WillKoehrsen, line 825: The problem started on June 4 when the site updated its certificate. This solution is not acceptable for a production environment so I'm working through other options.

Screenshot 2024-06-26 at 15 14 54

gridstatus/ieso.py
```diff
@@ -822,7 +823,7 @@ def _request(self, url, verbose):
         sleep = 5
 
         while retry_num < max_retries:
-            r = requests.get(url)
+            r = requests.get(url, verify=certifi.where())
```
WillKoehrsen, line 826: This is the accepted solution but does not work locally.

kmax12 (approved):

PR #396, opened 2024-06-28
Fix Forecast Time for NYISO Load Forecast

WillKoehrsen:
- Fixes the Forecast Time for `NYISO().get_load_forecast`
- The correct forecast time is the modified date of the file

### DIFF
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -1,7 +1,9 @@
+from datetime import datetime
 from typing import BinaryIO
 
 import pandas as pd
 import requests
+from zoneinfo import ZoneInfo
 
 import gridstatus
 from gridstatus import utils
@@ -870,7 +872,13 @@ class NYISO(ISOBase):
                     log(msg, verbose)
                     continue
                 df = pd.read_csv(z.open(csv_filename))
-                df["File Date"] = d.normalize()
+
+                # Get the modified date of the file
+
+                df["File Date"] = datetime(
+                    *z.getinfo(csv_filename).date_time,
+                    tzinfo=ZoneInfo(self.default_timezone),
+                )
 
                 df = _handle_time(df, dataset_name)
                 all_dfs.append(df)
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -838,7 +838,7 @@ class NYISO(ISOBase):
 
             df = pd.read_csv(csv_url)
             df = _handle_time(df, dataset_name)
-            df["File Date"] = date.normalize()
+            df["File Date"] = self._get_load_forecast_file_date(date, verbose)
         else:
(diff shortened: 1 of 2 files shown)


gridstatus/nyiso.py
```diff
@@ -870,7 +872,13 @@ def _download_nyiso_archive(
                     log(msg, verbose)
                     continue
                 df = pd.read_csv(z.open(csv_filename))
-                df["File Date"] = d.normalize()
+
+                # Get the modified date of the file
+
+                df["File Date"] = datetime(
```
kmax12, line 878: why not use pandas here like we do elsewhere?

kmax12:
```python
pd.read_html("http://mis.nyiso.com/public/P-7list.htm", skiprows=2)
```

image

WillKoehrsen: Yeah, I haven't got the non-zip file publish time yet.

WillKoehrsen: Thanks, updated to use that file list.

gridstatus/nyiso.py
```diff
@@ -836,7 +838,7 @@ def _download_nyiso_archive(
 
             df = pd.read_csv(csv_url)
             df = _handle_time(df, dataset_name)
-            df["File Date"] = date.normalize()
+            df["File Date"] = self._get_load_forecast_file_date(date, verbose)
```
kmax12, line 841: this is inside of `def _download_nyiso_archive`, which isn't specific to load forecast

WillKoehrsen: Good point. I added a conditional so the `File Date` is only added when the caller requests it.

gridstatus/nyiso.py
```diff
@@ -836,7 +839,8 @@ def _download_nyiso_archive(
 
             df = pd.read_csv(csv_url)
             df = _handle_time(df, dataset_name)
-            df["File Date"] = date.normalize()
+            if add_file_date:
```
kmax12, line 842: maybe since this only works when `dataset_name == 'isolf'`, you just check for that and always add in that case. no need for the `add_file_date` parameter

kmax12 (approved):

gridstatus/nyiso.py
WillKoehrsen: Updated.

gridstatus/nyiso.py
```diff
@@ -879,6 +890,21 @@ def _download_nyiso_archive(
 
         return df.sort_values("Time").reset_index(drop=True)
 
+    def _get_load_forecast_file_date(self, date, verbose=False):
+        """Retrieves the last updated time for load forecast file from the archive"""
+        data = pd.read_html(
+            "http://mis.nyiso.com/public/P-7list.htm",
```
WillKoehrsen, line 896: This is the archive of load forecast files.

PR #398, opened 2024-07-11
Fix MISO LMP Hourly Data

WillKoehrsen:
- Fixes the MISO LMP hourly datasets (prelim, day ahead, and final)
- The error was introduced in https://github.com/kmax12/gridstatus/pull/390

### DIFF
--- a/gridstatus/tests/test_miso.py
+++ b/gridstatus/tests/test_miso.py
@@ -108,6 +108,7 @@ class TestMISO(BaseTestISO):
         Markets.REAL_TIME_HOURLY_PRELIM,
     )
     def test_get_lmp_historical(self, market):
+        # Prelim data only goes back 4 days
         if market == Markets.REAL_TIME_HOURLY_PRELIM:
             date = self.local_today() - pd.Timedelta(days=2)
         else:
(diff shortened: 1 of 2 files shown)


gridstatus/tests/test_miso.py
```diff
@@ -108,6 +108,7 @@ def test_lmp_date_range(self, market):
         Markets.REAL_TIME_HOURLY_PRELIM,
     )
     def test_get_lmp_historical(self, market):
+        # Prelim data only goes back 4 days
```
WillKoehrsen, line 111: These tests [did fail](https://github.com/kmax12/gridstatus/actions/runs/9784853404/job/27016653393) when the error was introduced but since it constantly fails now, I didn't wait for it to finish before merging. This specific test passes now although there are still others that do not in `test_miso`.

gridstatus/tests/test_miso.py
```diff
@@ -252,7 +253,7 @@ def test_get_solar_forecast_historical(self):
         assert df["Publish Time"].dt.date.unique() == pd.to_datetime(past_date).date()
 
     def test_get_solar_forecast_historical_date_range(self):
-        past_date = self.local_today() - pd.Timedelta(days=250)
+        past_date = self.local_today() - pd.Timedelta(days=200)
```
WillKoehrsen, line 256: These tests were failing because there was no forecast with the publish date '2023-11-06'. I looked through the source data and confirmed this. There were two forecasts with the publish date '2023-11-07'.

kmax12 (approved):

PR #399, opened 2024-07-11
Add PJM 5 minute solar generation and instantaneous wind generation

thehomebrewnerd:
Adds:
- https://dataminer2.pjm.com/feed/five_min_solar_generation/definition
- https://dataminer2.pjm.com/feed/instantaneous_wind_gen/definition

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -1396,7 +1396,7 @@ class PJM(ISOBase):
         )
 
     @support_date_range(frequency=None)
-    def get_five_min_solar_generation(
+    def get_solar_generation_5_min(
         self,
         date: str | pd.Timestamp,
         end: Optional[str | pd.Timestamp] = None,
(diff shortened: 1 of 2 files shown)


gridstatus/pjm.py
WillKoehrsen, line 1: When I run `iso.get_five_min_solar_generation('today')`, the data starts at "03:35:00-04:00"

Screenshot 2024-07-11 at 14 14 03

but when I run `iso.get_instantaneous_wind_generation('today')` the data starts at the beginning of the local day "00:00:01-04:00".

Screenshot 2024-07-11 at 14 14 52

I think the latter is the behavior we want.

thehomebrewnerd: @WillKoehrsen There is no data for solar gen before that time stamp being returned from the API. See image below from the data explorer: 
image

WillKoehrsen: I see the same behavior on the PJM data explorer, so maybe this is fine. 

Screenshot 2024-07-11 at 14 17 17


Screenshot 2024-07-11 at 14 17 08

thehomebrewnerd: Seems like there is some missing data on the PJM side...

WillKoehrsen: Oh, yeah, you're right. The behavior is correct on other days (the data starts at the same time for both datasets).

WillKoehrsen (commented):
The returned results are as expected and the added tests pass locally 👍🏽. We generally wait to merge these until we've verified they work with our internal libraries though.

kmax12 (approved):

WillKoehrsen (approved):

PR #404, opened 2024-07-19
Add scrapers for PJM wind and solar generation zonal data

thehomebrewnerd:


gridstatus/pjm.py
```diff
@@ -1693,3 +1693,134 @@ def _parse_transmission_limits(self, df: pd.DataFrame):
...
+
+    def _parse_wind_generation_zonal(self, df: pd.DataFrame):
+        df = df.pivot_table(
+            index=["Time", "Interval Start", "Interval End"],
+            columns="area",
+            values="wind_generation_mw",
+            aggfunc="first",
+        ).reset_index()
+
+        # SOUTH and OTHER columns did not exist at the start of the
```
thehomebrewnerd, line 1806: @kmax12 @cwaldoch These areas are not present at the start of the data - they don't show up until around 2022. Do we always want these columns to be present in the results?

gridstatus/pjm.py
```diff
@@ -1693,3 +1693,134 @@ def _parse_transmission_limits(self, df: pd.DataFrame):
...
+        df = df[
+            [
+                "Interval Start",
+                "Interval End",
+                "MIDATL",
+                "OTHER",
+                "RFC",
+                "RTO",
+                "SOUTH",
+                "WEST",
```
WillKoehrsen, line 1753: Does solar have the same issue with some areas not present in the historical data as wind?

thehomebrewnerd: No, it does not. It also does not go back as far, only to 2019 instead of 2011.

WillKoehrsen (approved):
Looks good. Should wait for confirmation on those extra areas though.

gridstatus/pjm.py
WillKoehrsen: Examples:

1. https://github.com/kmax12/gridstatus/blob/main/gridstatus/spp.py#L527-L530
2. https://github.com/kmax12/gridstatus/blob/main/gridstatus/miso.py#L264-L266
3. https://github.com/kmax12/gridstatus/blob/main/gridstatus/ercot.py#L1438-L1440

kmax12: yes, our convention has been to include all column ever offered in the dataset and use nulls when they are missing. adding the info to description is important to avoid confusion by users

kmax12 (approved):

PR #405, opened 2024-07-24
ERCOT Unplanned Resource Outages Updates

WillKoehrsen:
- Adds `Publish Time` to ERCOT unplanned resource outages
- Renames `Report Time` to `Current As Of` to better reflect what this column represents

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2084,11 +2084,10 @@ class Ercot(ISOBase):
 
         xls = utils.get_zip_file(doc.url, verbose=verbose)
 
-        df = self._handle_unplanned_resource_outages_file(xls)
-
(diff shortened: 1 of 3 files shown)


gridstatus/ercot.py
```diff
@@ -2106,7 +2105,10 @@ def _handle_unplanned_resource_outages_file(self, xls):
             skipfooter=1,
         )
 
-        df.insert(0, "Report Time", as_of)
+        # Already has local timezone
+        df["Publish Time"] = pd.to_datetime(doc.publish_date)
+
+        df.insert(0, "Current As Of", as_of)
```
WillKoehrsen, line 2111: The reasoning behind renaming this from `Report Time` to `Current As Of` is ERCOT calls this "Current as of" in the data. `Report Time` seems ambiguous compared to `Current As Of`.

PR #406, opened 2024-07-24
Add spp hourly load dataset scraper

thehomebrewnerd:
- https://portal.spp.org/pages/hourly-load

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -1240,7 +1240,6 @@ class SPP(ISOBase):
         msg = f"Downloading {url}"
         log(msg, verbose)
         df = pd.read_csv(url)
-        df = df.rename(columns=lambda x: x.strip())
 
         return self._process_hourly_load(df)
 
@@ -1265,10 +1264,13 @@ class SPP(ISOBase):
         return df
 
     def _process_hourly_load(self, df):
-        # Some historical files contain null rows. Drop them.
+        # Some column names contain leading whitespace in some files - remove it
+        df = df.rename(columns=lambda x: x.strip())
+
+        # Some files contain null rows. Drop them.
         df = df.dropna(how="all")
 
-        # Some historical files don't have time 00:00 for the first interval in a day
+        # Some files don't have time 00:00 for the first interval in a day
         # for example 12/2/2016 instead of 12/2/2016 00:00. This causes datetime
         # conversion problems. This fixes it.
         def clean_market_hour(val):
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -1251,7 +1251,19 @@ class SPP(ISOBase):
         """
         url = f"{FILE_BROWSER_DOWNLOAD_URL}/hourly-load?path=/{year}/{year}.zip"  # noqa
         df = utils.download_csvs_from_zip_url(url, verbose=verbose)
+
+        # Some data files have leading whitespace in header - remove it
         df = df.rename(columns=lambda x: x.strip())
+        
+        # Some historical files don't have time 00:00 for the first interval in a day
+        # for example 12/2/2016 instead of 12/2/2016 00:00. This causes datetime
+        # conversion problems. This fixes it.
+        def clean_market_hour(val):
+            if val.endswith(":00"):
+                return val
+            return val + " 00:00"
+        
+        df["MarketHour"] = df["MarketHour"].apply(clean_market_hour)
 
         df = self._process_hourly_load(df)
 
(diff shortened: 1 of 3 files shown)


gridstatus/spp.py
```diff
@@ -1216,6 +1216,102 @@ def _match(
             )
         ]
 
+    ### NEW
+    @support_date_range("DAY_START")
+    def get_hourly_load(self, date, end=None, verbose=False):
+        """Get Hourly Load
+
+        Supports recent data. For historical annual data use get_hourly_load_annual
```
WillKoehrsen, line 1224: It looks like this method doesn't support "today" or "latest" since the data is released on a delay of 2 days. We should raise a `NotSupport` exception in these cases.

thehomebrewnerd: Updated: [172f553](https://github.com/kmax12/gridstatus/pull/406/commits/172f5538cf5508b655392af2cde08cb95ebc4b1f)

gridstatus/utils.py
```diff
@@ -219,14 +219,19 @@ def download_csvs_from_zip_url(url, process_csv=None, verbose=False):
...
     all_dfs = []
+
     for f in z.filelist:
         if f.filename.endswith(".csv"):
             df = pd.read_csv(z.open(f.filename))
             if process_csv:
                 df = process_csv(df, f.filename)
+
+            # Some data files have leading whitespace in header - remove it
+            df = df.rename(columns=lambda x: x.strip())
```
WillKoehrsen, line 230: Does this have effects on other places that use this util?

thehomebrewnerd: That's a good question and I don't know. Maybe I should add a param to control this so it defaults to the current behavior instead? Without this here the concat gets messed up and we don't get a clean dataframe, so we somehow need to handle this new wrinkle.

WillKoehrsen: Yeah, adding another param to not change the existing behavior seems like the right approach.

gridstatus/spp.py
```diff
@@ -1216,6 +1218,109 @@ def _match(
...
+        if date in ["today", "latest"] or utils.is_today(
+            date, tz=self.default_timezone
+        ):
+            raise NotSupported("Only historical data is available for hourly load data")
+
+        url = f"{FILE_BROWSER_DOWNLOAD_URL}/hourly-load?path=/{date.strftime('%Y')}/DAILY_HOURLY_LOAD-{date.strftime('%Y%m%d')}.csv"  # noqa
+        msg = f"Downloading {url}"
+        log(msg, verbose)
+        df = pd.read_csv(url)
+        df = df.rename(columns=lambda x: x.strip())
```
kmax12, line 1243: i think the goal is to have `_process_hourly_load` do everything besides download the data, so i think move this line in there.

gridstatus/spp.py
```diff
@@ -1216,6 +1218,109 @@ def _match(
...
+            "WFEC",
+            "WR",
+        ]
+
+        # historical data doesn't have all columns
+        for c in cols:
+            if c not in df.columns:
+                df[c] = pd.NA
+
+        df = df[cols]
```
kmax12, line 1320: let's add a `System Total` column at the end that sums everything up. makes this data bit easier to work with if you want to use it as a source for all of SPP.

gridstatus/utils.py
```diff
@@ -215,18 +215,26 @@ def get_response_blob(resp: requests.Response) -> io.BytesIO:
...
 
     all_dfs = []
+
     for f in z.filelist:
         if f.filename.endswith(".csv"):
             df = pd.read_csv(z.open(f.filename))
             if process_csv:
                 df = process_csv(df, f.filename)
+
+            if strip_whitespace_from_cols:
```
kmax12, line 231: maybe this can be removed if you move the stripping to `_process_hourly_load`?

kmax12 (approved):
looks good once everything is working downstream

gridstatus/utils.py
thehomebrewnerd: It needs to happen before the concat here based on the data. The data for some years has some files with whitespace in the column names and some without. If we leave the whitespace pandas thinks they are different columns and messes up the concat for the year. Unfortunately, I think we have to leave it here.

gridstatus/spp.py
thehomebrewnerd: Sounds good. I'll see if I can also move some of the clean up logic from `get_hourly_load_annual` into `_process_hourly_load` as well.

WillKoehrsen (approved):

kmax12 (approved):
still looks good to me

PR #411, opened 2024-08-02
Fix SPP Cookie Issue

WillKoehrsen:
- Fixes an issue with requesting and setting cookies + header for SPP file listing
- https://github.com/kmax12/gridstatus/issues/410

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -3,7 +3,6 @@ from typing import BinaryIO
 import pandas as pd
 import requests
 import tqdm
-from bs4 import BeautifulSoup
 
 from gridstatus import utils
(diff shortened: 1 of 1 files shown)


gridstatus/spp.py
```diff
@@ -1161,17 +1166,12 @@ def _get_marketplace_session(self) -> dict:
         """
         html = requests.get(FILE_BROWSER_API_URL)
```
WillKoehrsen, line 1167: This request returns a 404, but it still has a `JSESSIONID` and `ZSRF-TOKEN` in the cookies of the response.

kmax12 (approved):

PR #412, opened 2024-08-02
SPP LMP by Bus

WillKoehrsen:
- Adds SPP LMP by bus
- https://portal.spp.org/pages/rtbm-lmp-by-bus

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -10,6 +10,7 @@ from gridstatus.decorators import FiveMinOffset, support_date_range
 from gridstatus.gs_logging import log
 from gridstatus.lmp_config import lmp_config
 
+RTBM_LMP_BY_BUS = "rtbm-lmp-by-bus"
 FS_RTBM_LMP_BY_LOCATION = "rtbm-lmp-by-location"
 FS_DAM_LMP_BY_LOCATION = "da-lmp-by-location"
 MARKETPLACE_BASE_URL = "https://portal.spp.org"
@@ -29,6 +30,7 @@ BASE_LOAD_FORECAST_MID_TERM_URL = f"{FILE_BROWSER_DOWNLOAD_URL}/mtlf-vs-actual?p
 
 
 LOCATION_TYPE_ALL = "ALL"
+LOCATION_TYPE_BUS = "Bus"
 LOCATION_TYPE_HUB = "Hub"
 LOCATION_TYPE_INTERFACE = "Interface"
 LOCATION_TYPE_SETTLEMENT_LOCATION = "Settlement Location"
@@ -86,6 +88,7 @@ class SPP(ISOBase):
 
     location_types = [
         LOCATION_TYPE_ALL,
+        LOCATION_TYPE_BUS,
         LOCATION_TYPE_HUB,
         LOCATION_TYPE_INTERFACE,
         LOCATION_TYPE_SETTLEMENT_LOCATION,
@@ -807,6 +810,7 @@ class SPP(ISOBase):
             - ``Hub``
             - ``Interface``
             - ``ALL``
+            - ``Bus``
         """
         if market not in self.markets:
             raise NotSupported(f"Market {market} not supported")
@@ -818,6 +822,7 @@ class SPP(ISOBase):
             df = self._get_rtm5_lmp(
                 date,
                 end,
+                location_type,
                 verbose,
             )
         elif market == Markets.DAY_AHEAD_HOURLY:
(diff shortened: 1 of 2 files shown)


gridstatus/spp.py
```diff
@@ -855,18 +860,29 @@ def _get_rtm5_lmp(
...
             urls = [
                 FILE_BROWSER_DOWNLOAD_URL
                 + "/"
-                + FS_RTBM_LMP_BY_LOCATION
-                + "?path=%2FRTBM-LMP-SL-latestInterval.csv",
+                + fs_name
+                + f"?path=%2F{latest_file_id}-latestInterval.csv",
             ]
         else:
             urls = self._file_browser_list(
```
kmax12, line 884: a big downside to our current approach here of using `_file_browser_list` is it only supports querying for one day at a time.

the files we are downloading have a known url structure, so there is no need to this operation where we list things. we can just directly construct the URL for the interval we want to load. 

this approach is currently taken by [`get_lmp_real_time_weis`](https://github.com/kmax12/gridstatus/blob/affd7e8d306c5a8f65b784dc2069a3e0a3fdda5d/gridstatus/spp.py#L1055)

we should update this function to look more like that and refactor this to not use `_file_browser_list`. 

in doing this, i think it make sense to break apart the current `get_lmp` function in the three functions for each of separate datasets. this will break our current api, but i think we should eventually move this library away from having the `get_lmp` function that takes `market` or `location_type` as a parameter to decide which dataset to grab

kmax12 (approved):

gridstatus/spp.py
```diff
@@ -10,6 +10,7 @@
 from gridstatus.gs_logging import log
```
kmax12, line 10: accidentally hit approve.

kmax12 (changes requested):

gridstatus/spp.py
```diff
@@ -782,6 +786,60 @@ def get_interconnection_queue(self, verbose=False):
...
+            verbose: print url
+        """
+        year = date.strftime("%Y")
+        month = date.strftime("%m")
+        day = date.strftime("%d")
+
+        day_url = f"https://portal.spp.org/file-browser-api/download/rtbm-lmp-by-bus?path=%2F{year}%2F{month}%2FBy_Day%2FRTBM-LMP-DAILY-BUS-{year}{month}{day}.csv"
+
+        # First, check if there are daily files for the date. If not, go to
+        # 5-minute files.
```
WillKoehrsen, line 805: @kmax12 I haven't finished this yet, but what do you think of the approach of first looking for daily files, then switching to 5-min interval files if the daily file isn't available? This cuts down the number of requests by a factor of 288 for historical dates.

There are also monthly rollup files but those are fairly large so I stuck with the daily files.

gridstatus/spp.py
WillKoehrsen: I've updated the PR to separate the `get_lmp` method into three functions:

* `get_lmp_real_time_5_min_by_location`
* `get_lmp_real_time_5_min_by_bus`
* `get_lmp_day_ahead_hourly`

I've also removed `_file_browser_list` in favor of constructing the URLs through string interpolation.

gridstatus/spp.py
```diff
@@ -782,55 +785,178 @@ def get_interconnection_queue(self, verbose=False):
 
         return queue
 
-    @lmp_config(
-        supports={
-            Markets.REAL_TIME_5_MIN: ["latest", "today", "historical"],
-            Markets.DAY_AHEAD_HOURLY: ["latest", "today", "historical"],
-        },
-    )
+    @support_date_range(frequency=None)
```
WillKoehrsen, line 788: I set the frequency to `None` because with `DAY_START` you can't pass in two times on the same date. In that case, the decorator would set `date` to the beginning of the day and set `end` to `None` which is not what you want.

gridstatus/spp.py
WillKoehrsen: This approach proved too complicated. Now this only uses the 5-minute files.

gridstatus/spp.py
WillKoehrsen: Nevermind this, I updated to use `"5_MINUTE"` frequency.

gridstatus/spp.py
```diff
@@ -782,55 +784,130 @@ def get_interconnection_queue(self, verbose=False):
...
-        supports={
-            Markets.REAL_TIME_5_MIN: ["latest", "today", "historical"],
-            Markets.DAY_AHEAD_HOURLY: ["latest", "today", "historical"],
-        },
-    )
+    def get_lmp_real_time_5_min_by_location(
+        self,
+        date,
+        end=None,
+        location_type=LOCATION_TYPE_ALL,
```
kmax12, line 791: we don't need to change this for this PR since it's preexisting, but just an overall comment to keep in mind. 

 I think we should eventually move away from doing any `location_type` filtering in `gridstatus` in the LMP method. we should focus library on simply querying the ISOs API and returning cleaned data as they provide it. no benefit of maintaining this extra logic to filter down what the ISOs return.

we can do this type of filtering much better on the hosted API.

kmax12 (approved):
LGTM!

gridstatus/spp.py
WillKoehrsen: I agree. We already fetch all the data, and we should return it all to the caller who can filter if necessary.

PR #414, opened 2024-08-07
Update How ERCOT Unplanned Resource Outages Handles Date Parameter

WillKoehrsen:
- Updates `Ercot().get_unplanned_resource_outages()` to treat the `date` parameter as a filter on the publish date instead of the data date which required adding 3 days to given date
- This change gives more consistency across methods for datasets with a publish date

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2067,10 +2067,12 @@ class Ercot(ISOBase):
     def get_unplanned_resource_outages(self, date, verbose=False):
         """Get Unplanned Resource Outages.
 
-        Data published at ~5am central on the 3rd day after the day of interest.
+        Data published at ~5am central on the 3rd day after the day of interest. Since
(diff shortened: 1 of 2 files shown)


gridstatus/ercot.py
```diff
@@ -2079,7 +2081,7 @@ def get_unplanned_resource_outages(self, date, verbose=False):
         """
         doc = self._get_document(
             report_type_id=UNPLANNED_RESOURCE_OUTAGES_REPORT_RTID,
-            date=date.normalize() + pd.DateOffset(days=3),
+            date=date.normalize(),
```
kmax12, line 2084: would it make sense to switch to `@support_date_range(frequency=None)` and then use `published_before` and `published_after` like we do for `get_load_forecast`? that will more reliable iterate over all the publish times i think

WillKoehrsen: Okay, that makes sense. I've updated the code.

PR #416, opened 2024-08-09
Fix Parsing Issue for ISONE LMP

WillKoehrsen:
- Fixes a parsing issue for ISONE introduced in https://github.com/kmax12/gridstatus/pull/409

### DIFF
--- a/gridstatus/tests/test_isone.py
+++ b/gridstatus/tests/test_isone.py
@@ -122,21 +122,24 @@ class TestISONE(BaseTestISO):
         )
 
     def test_get_lmp_real_time_no_intervals_gets_current_data(self):
-        # Make sure when there are no intervals to fetch, the method still
-        # fetches the current data.
-        date = self.local_now().normalize() + pd.DateOffset(hours=2)
+        date = self.local_now() - pd.DateOffset(hours=2)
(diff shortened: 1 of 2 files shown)


WillKoehrsen:
None of the `get_lmp` tests currently fail on `main`, which is how this error got through. I will write a test that fails on `main` but passes with these changes once I figure out the circumstances under which the error occurs.

kmax12 (approved):

gridstatus/tests/test_isone.py
```diff
@@ -122,21 +122,24 @@ def test_get_lmp_dst_boundaries(self, date, market):
...
-        date = self.local_now().normalize() + pd.DateOffset(hours=2)
+        date = self.local_now() - pd.DateOffset(hours=2)
         end = date + pd.DateOffset(hours=1)
 
-        df = self.iso.get_lmp(
-            (date, end),
-            market=Markets.REAL_TIME_5_MIN,
-            verbose=VERBOSE,
-        )
+        # Mock the method _select_intervals_for_data_request in the class to return []
```
WillKoehrsen, line 128: I verified that the updated version of this test fails on main and passes with the changes in this PR.

PR #419, opened 2024-08-12
Fix Makefile pytest Command

WillKoehrsen:
- Fixes the Makefile `pytest` command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ clean:
 	find . -name '*~' -delete
 	find . -name '.coverage.*' -delete
 
-PYTEST_CMD := poetry run pytest -s -vv gridstatus/ -n auto --reruns 5 --reruns-delay 3 -durations=25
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ clean:
 	find . -name '*~' -delete
 	find . -name '.coverage.*' -delete
 
-PYTEST_CMD := poetry run pytest -s -vv gridstatus/ -n auto --reruns 5 --reruns-delay 3 -durations=25
+PYTEST_CMD := poetry run pytest -s -vv gridstatus/ -n auto --reruns 5 --reruns-delay 3 --durations=25
```
WillKoehrsen, line 9: Sorry, I introduced this in #418 to print out the 25 longest-running tests, but incorrectly only used one `-`.

kmax12 (approved):

PR #420, opened 2024-08-19
Adds PJM Ancillary Service Scrapers

thehomebrewnerd:
Adds scrapers for:
- https://dataminer2.pjm.com/feed/reserve_market_results/definition
- https://dataminer2.pjm.com/feed/da_reserve_market_results/definition


gridstatus/pjm.py
```diff
@@ -1824,3 +1836,172 @@ def _parse_wind_generation_by_area(self, df: pd.DataFrame):
...
+            "reserve_market_results",
+            start=date,
+            params={
+                "fields": "datetime_beginning_ept,datetime_beginning_utc,locale,"
+                "service,mcp,mcp_capped,reg_ccp,reg_pcp,as_req_mw,total_mw,as_mw,"
+                "ss_mw,tier1_mw,ircmwt2,dsr_as_mw,nsr_mw,regd_mw"
+            },
+            end=end,
+            filter_timestamp_name="datetime_beginning",
+            interval_duration_min=5,
```
WillKoehrsen, line 1953: Should this be `60 if date < pd.Timestamp('2022-10-01') else 5`?

> Data posted prior to October 1, 2022 will be hourly granularity. Data posted after October 1st will be five-minute granularity as part of PJM Comprehensive Pricing Reform.

thehomebrewnerd: Maybe, I'm definitely not clear how to handle this. What if that Oct 1, 2022 line falls between the specified start and end dates?

WillKoehrsen: Good point. It doesn't seem like we have resolved this issue yet. My preference would be for two methods, one for data before the frequency change and one for data after to avoid mixing frequencies.

gridstatus/decorators.py
```diff
@@ -56,7 +56,7 @@ def wrapped_f(*args, **kwargs):
                 save_to = args_dict.pop("save_to")
                 os.makedirs(save_to, exist_ok=True)
 
-            error = "ignore"
+            error = "raise"
```
thehomebrewnerd, line 59: Is this change problematic? Without this the `ValueError` I added if date/end combos are invalid gets swallowed.

kmax12: swallowed in what situation? 

the point of this parameter is so the the caller of `get_real_time_as_market_results` can override this setting.

thehomebrewnerd: > swallowed in what situation?
> 
> the point of this parameter is so the the caller of `get_real_time_as_market_results` can override this setting.

NVM. I wasn't following the logic after this line correctly.

gridstatus/pjm.py
thehomebrewnerd: Updated so an error is raised if start/end dates that include the granularity change date are specified. As a side note, the granularity actually changes on Sep 1, 2022 when you look at the data, and not the published date of Oct 1, 2022.

image

WillKoehrsen (approved):

PR #421, opened 2024-08-22
Fix URL for ERCOT API registration instructions

matthewkennedy5:
The current URL in the comment is a dead link (404), updated it to the new URL for ERCOT's API registration and authentication page


WillKoehrsen (approved):
@matthewkennedy5 Thank you for the fix!

PR #423, opened 2024-08-24
Address Issue with ERCOT LMP Timestamp

WillKoehrsen:
- Addresses an issue where ERCOT creates files with invalid timestamps
- Fix is to coerce to a NaT when the timestamp is invalid then drop the rows with NaT

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2388,10 +2388,16 @@ class Ercot(ISOBase):
             },
         )
 
-        df["SCED Timestamp"] = pd.to_datetime(df["SCED Timestamp"]).dt.tz_localize(
+        # Some files have errors with the timestamp
+        df["SCED Timestamp"] = pd.to_datetime(
+            df["SCED Timestamp"],
+            errors="coerce",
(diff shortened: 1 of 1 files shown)


WillKoehrsen:
In [this file](https://www.ercot.com/misdownload/servlets/mirDownload?doclookupId=1030016183), ERCOT has added a row of headers in the middle of the file, causing the data retrieval to fail when it tries to convert "SCEDTimestamp" to a pandas timestamp. 

Screenshot 2024-08-23 at 21 17 09

The error can be reproduced on `main` with:

```pandas
from pandas import Timestamp
from gridstatus.ercot import Ercot, ELECTRICAL_BUS_LOCATION_TYPE

Ercot().get_lmp(
    date=(
        Timestamp("2024-08-23 00:00:00", tz="US/Central"),
        Timestamp("2024-08-23 00:10:00", tz="US/Central"),
    ),
    location_type=ELECTRICAL_BUS_LOCATION_TYPE,
)
```

kmax12 (approved):

PR #424, opened 2024-08-25
Fix Tests August 2024

WillKoehrsen:
- Attempts to fix the failing tests

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -6,7 +6,7 @@ on: # Run on all pushes and pull requests
   pull_request:
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-m
     strategy:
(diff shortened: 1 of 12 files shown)


kmax12:
@WillKoehrsen im good to merge these in when you are

WillKoehrsen:
I think we might be getting rate-limited by the ERCOT API because the tests for Python 3.10 and 3.11 run at the same time with the same creds.

WillKoehrsen:
> I think we might be getting rate-limited by the ERCOT API because the tests for Python 3.10 and 3.11 run at the same time with the same creds.

This doesn't seem to be the case but the ERCOT API tests are failing.

Kladar (approved):
Thanks for fixing! Would love to get to a point where we basically just rely on good, comprehensive testing such that PRs that pass everything can be merged with as little human intervention as possible.

PR #426, opened 2024-09-04
Add CAISO Fuel Regions

WillKoehrsen:
- Adds `CAISO().get_fuel_regions()` method
- Retrieves the file available at https://www.caiso.com/library/iso-fuel-region-and-electric-region-definitions

### DIFF
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -647,6 +647,21 @@ class CAISO(ISOBase):
         ]
         return df
 
+    def get_fuel_regions(self, verbose=False):
+        """Retrieves the (mostly static) list of fuel regions with associated data.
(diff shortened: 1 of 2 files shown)


gridstatus/caiso.py
```diff
@@ -647,6 +647,21 @@ def get_gas_prices(
         ]
         return df
 
+    def get_fuel_regions(self, verbose=False):
```
WillKoehrsen, line 650: Is it accurate to call this `get_fuel_regions`, or should it be something like `get_gas_pricing_regions`?

WillKoehrsen: CAISO calls these "fuel regions" so I will leave the name. At some point, we might want to rename `get_gas_prices` to `get_fuel_prices` but I'll leave as is for now.

PR #427, opened 2024-09-06
CAISO Tie Flows and Tie Schedule

WillKoehrsen:
- Adds real time tie flows for CAISO
- Adds day ahead tie schedule for CAISO


gridstatus/caiso.py
```diff
@@ -1193,30 +1193,54 @@ def _process_tie_flows_data(self, df):
...
+
+        df.columns = df.columns.map(
+            lambda x: x.title()
+            .replace("_", " ")
+            .replace("Baa", "BAA")
+            .replace("Mw", "MW")
+            .replace("Id", "ID"),
         )
 
+        df["Interface ID"] = df["Tie Name"] + "_" + df["From BAA"] + "_" + df["To BAA"]
```
kmax12, line 1225: what do you think about using double underscore to separate each part? some of the values have underscores in them, so I think this will help the readability of these ids.

WillKoehrsen: That's a good idea though I'm not sure if double underscores or another separate, like a hyphen, would be better.

kmax12: im good with hyphen!

WillKoehrsen:
I've removed the day ahead tie schedule data because of the issues we need to sort out with that dataset. When we want to add it in a separate PR, we can use the code removed in [this commit](https://github.com/gridstatus/gridstatus/pull/427/commits/62ce9adcb07a62c10fd69ef7f8266c4229fa63f8).

kmax12 (approved):

PR #428, opened 2024-09-07
EIA Henry Hub Natural Gas Spot Prices

WillKoehrsen:
- Adds EIA Henry Hub Natural Gas Spot Price data

### DIFF
--- a/gridstatus/eia.py
+++ b/gridstatus/eia.py
@@ -16,6 +16,8 @@ import gridstatus
 from gridstatus import utils
 from gridstatus.gs_logging import log
 
+HENRY_HUB_NATURAL_GAS_SPOT_PRICES_PATH = "natural-gas/pri/fut"
+
 
 class EIA:
     BASE_URL = "https://api.eia.gov/v2/"
@@ -219,7 +221,6 @@ class EIA:
         df = raw_df.copy()
 
         if dataset in DATASET_CONFIG:
-
             df = DATASET_CONFIG[dataset]["handler"](df)
 
         return df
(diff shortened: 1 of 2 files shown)


gridstatus/eia.py
```diff
@@ -727,6 +752,28 @@ def _handle_fuel_type_data(df):
     return df
 
 
+def _handle_henry_hub_natural_gas_spot_prices(df):
+    df["Interval Start"] = pd.to_datetime(df["period"], utc=True)
```
WillKoehrsen, line 756: The other EIA datasets use "period" as the "Interval End" but that is not correct for this dataset because that would put the data one day ahead of how the EIA shows it [here](https://www.eia.gov/dnav/ng/hist/rngwhhdD.htm).

kmax12: maybe add this as a comment in the code too?

kmax12 (approved):

gridstatus/eia.py
kmax12, line 756: are you sure this is UTC time?

WillKoehrsen: The `period` isn't a time, it's only a date. But, we don't really support datasets that only have a date column so I parsed it into UTC.

gridstatus/eia.py
```diff
@@ -727,6 +754,35 @@ def _handle_fuel_type_data(df):
...
 
+def _handle_henry_hub_natural_gas_spot_prices(df):
+    # The other EIA datasets use "period" as the "Interval End" but that is not correct
+    # for this dataset because that would put the data one day ahead of how the EIA
+    # shows it here: https://www.eia.gov/dnav/ng/hist/rngwhhdD.htm
+    # We use the HENRY_HUB_TIMEZONE because the spot prices are based on delivery
+    # at Henry Hub
+    df["Interval Start"] = pd.to_datetime(df["period"]).dt.tz_localize(
+        HENRY_HUB_TIMEZONE,
+    )
```
WillKoehrsen, line 765: Updated to use US/Central time for interval start and end.

WillKoehrsen: I'm not sure this is correct because this data series also includes futures prices which are based on the NYMEX. 

> Official daily closing prices at 2:30 p.m. from the trading floor of the New York Mercantile Exchange (NYMEX) for a specific delivery month.

Should we update this series to only include the spot prices?

WillKoehrsen: Based on conversations, the answer was we should also include the futures prices (which end on April 5, 2024) and still use Central Time. We should note this in the dataset description.

PR #432, opened 2024-09-28
Handle naming changes in ercot solar and wind hourly reports

kmax12:
ERCOT Changes
image


WillKoehrsen (approved):

Kladar:
Chiming in that I'm following along here!

PR #431, opened 2024-09-23
[ENG-229] - Change CAISO Load file URL

Kladar:
## Summary 
This changes the approach to CAISO load file getting slightly. Pulls from the `latest` file first, given that it could be more up to date than the history file if today's data is desired. Also adds type annotations to the function and variables, switches to using f-strings, some other python best practice stuff.

## Details
@kmax12 Mostly gauging how much latitude there is to alter the functionality of this code (for example, removing the `try - except` block, with the desire in the future to handle exceptions separately, given that there are some other failure modes I could see this running into)

## Tests
Tried running just the `tests/test_caiso.py` and it seems to have hung a bit, as well as running into a certificate error that might be straightforwardly remedied. Details:

CleanShot 2024-09-23 at 15 35 36@2x

PR #442, opened 2024-09-30
[ENG 787] Fix CAISO file dating

Kladar:
## Summary
Updates the CAISO file-puller logic to check the latest valid value `Time` and checks it against the current CAISO time. If the latest value is later than the current time, it means the file is from yesterday, so update the datetime column accordingly.

Also adds some quality of life (which I have a bit of a habit of doing on PRs, hopefully nothing controversial).

## Relevant Issues (Optional)
Bug introduced in PR #431 

## Details
Quality of Life Stuff like 
- docstrings
- slight import renaming (import utils functions individually rather than as a module), 
- a tweak to ruff (since we are using black formatter for line-length, the ruff-linter is annoying right now). Recommend switching to ruff-format 

## Checklist

- [x] I have read and followed the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added or updated necessary documentation
- [x] I have run the pre-commit hooks and addressed any issues
- [x] I have updated the CHANGELOG.md file if necessary

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,22 +1,7 @@
 ## Summary
 <!-- Provide a brief summary of the changes in this pull request -->
 
-## Relevant Issues (Optional)
-<!-- List any related issues that this pull request addresses or closes -->
-<!-- Use the GitHub syntax to link issues: Closes #123, Addresses #456 -->
+**Relevant Issues**: 
 
-## Details
+### (*Optional*) Details
 <!-- Provide a more detailed explanation of the changes made -->
-<!-- Include any important implementation details or design decisions -->
-
-## Resources (Optional)
-<!-- Include any relevant links, documentation, or screenshots that may be helpful for reviewers -->
-
-## Checklist
-<!-- Mark the following checklist items as completed by replacing [ ] with [x] -->
-- [ ] I have read and followed the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
-- [ ] I have added tests that prove my fix is effective or that my feature works
-- [ ] I have added or updated necessary documentation
-- [ ] I have run the pre-commit hooks and addressed any issues
-- [ ] I have updated the CHANGELOG.md file if necessary
-
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,0 +1,22 @@
+## Summary
+<!-- Provide a brief summary of the changes in this pull request -->
+
+## Relevant Issues (Optional)
+<!-- List any related issues that this pull request addresses or closes -->
+<!-- Use the GitHub syntax to link issues: Closes #123, Addresses #456 -->
+
+## Details
+<!-- Provide a more detailed explanation of the changes made -->
+<!-- Include any important implementation details or design decisions -->
+
+## Resources (Optional)
+<!-- Include any relevant links, documentation, or screenshots that may be helpful for reviewers -->
+
+## Checklist
+<!-- Mark the following checklist items as completed by replacing [ ] with [x] -->
+- [ ] I have read and followed the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
+- [ ] I have added tests that prove my fix is effective or that my feature works
+- [ ] I have added or updated necessary documentation
+- [ ] I have run the pre-commit hooks and addressed any issues
+- [ ] I have updated the CHANGELOG.md file if necessary
+
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -1497,7 +1497,7 @@ def _get_historical(
 
     # sometimes this data is from the previous day, in which case we want to label it as such
     latest_file_time = utils.check_latest_value_time(df, file)
-    current_caiso_time = pd.Timestamp.now(tz="US/Pacific")
+    current_caiso_time = pd.Timestamp.now(tz=CAISO.default_timezone)
 
     if latest_file_time > current_caiso_time:
         date = date - pd.Timedelta(days=1)
@@ -1505,7 +1505,7 @@ def _get_historical(
     df["Time"] = df["Time"].apply(
         _make_timestamp,
         today=date,
-        timezone="US/Pacific",
+        timezone=CAISO.default_timezone,
     )
 
     # sometimes returns midnight, which is technically the next day
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -12,7 +12,7 @@ import tabula
 from tabulate import tabulate
 from termcolor import colored
 
-from gridstatus import utils
+from gridstatus import caiso_utils, utils
 from gridstatus.base import (
     GridStatus,
     ISOBase,
@@ -1501,7 +1501,7 @@ def _get_historical(
 
     # for the latest data, we want to check if the data is actually from the previous day and update the date accordingly
     if latest:
-        latest_file_time = utils.check_latest_value_time(df, column)
+        latest_file_time = caiso_utils.check_latest_value_time(df, column)
         current_caiso_time = pd.Timestamp.now(tz=CAISO.default_timezone)
 
         if latest_file_time > current_caiso_time:
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -12,6 +12,7 @@ import tabula
 from tabulate import tabulate
 from termcolor import colored
 
+from gridstatus import utils
 from gridstatus.base import (
     GridStatus,
     ISOBase,
@@ -22,14 +23,6 @@ from gridstatus.base import (
 from gridstatus.decorators import support_date_range
 from gridstatus.gs_logging import log
 from gridstatus.lmp_config import lmp_config
-from gridstatus.utils import (
-    _handle_date,
-    check_latest_value_time,
-    format_interconnection_df,
-    get_response_blob,
-    is_today,
-    move_cols_to_front,
-)
 
 _BASE = "https://www.caiso.com/outlook/SP"
 _HISTORY_BASE = "https://www.caiso.com/outlook/SP/History"
@@ -306,7 +299,7 @@ class CAISO(ISOBase):
             publish_time_offset_from_day_start=publish_time_offset_from_day_start,
         )
 
-        df = move_cols_to_front(
+        df = utils.move_cols_to_front(
             df.rename(
                 columns={
                     "TRADING_HUB": "Location",
@@ -372,7 +365,7 @@ class CAISO(ISOBase):
         return data
 
     def get_pnodes(self, verbose=False):
-        start = _handle_date("today")
+        start = utils._handle_date("today")
 
         df = self.get_oasis_dataset(
             dataset="pnode_map",
@@ -720,7 +713,7 @@ class CAISO(ISOBase):
         msg = f"Downloading interconnection queue from {url}"
         log(msg, verbose)
         response = requests.get(url)
-        return get_response_blob(response)
+        return utils.get_response_blob(response)
 
     def get_interconnection_queue(self, verbose=False):
         raw_data = self.get_raw_interconnection_queue(verbose)
@@ -813,7 +806,7 @@ class CAISO(ISOBase):
             "Winter Capacity (MW)",
         ]
 
-        queue = format_interconnection_df(
+        queue = utils.format_interconnection_df(
             queue=queue,
             rename=rename,
             extra=extra_columns,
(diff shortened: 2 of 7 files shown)


gridstatus/caiso.py
```diff
@@ -23,6 +22,14 @@
 from gridstatus.decorators import support_date_range
 from gridstatus.gs_logging import log
 from gridstatus.lmp_config import lmp_config
+from gridstatus.utils import (
```
kmax12, line 25: why this change?

I don't feel strongly, but I like when I'm reading the code later in the file to be able to see that im using a repo-wide `util` function. 

also, this change broke the code: https://github.com/gridstatus/gridstatus/actions/runs/11113242953/job/30877128940?pr=442#step:7:31

pyproject.toml
```diff
@@ -82,10 +82,9 @@ line-length = 88
 src = ["gridstatus"]
 
 [tool.ruff.lint]
-ignore = []
+ignore = ["E501"]
```
kmax12, line 85: why add this to ignore?

Kladar: because we are using black, so black will try the formatting to the best of it's ability, but it falls down on comments and things

> Since you're using Black, you can just disable the Ruff E501 rule; Black will keep your lines under its line-length (as well as it can).

From [first answer here.](https://stackoverflow.com/questions/77396198/how-to-auto-format-e501-line-too-long-89-88-characters-in-pre-commit) 

Mostly I'd like to move to ruff-format (especially if we are using ruff-lint), but that feels like a separate PR. This is a  temporary quality of life don't-make-me-count-characters-even-if-black-can't-figure-out-a-docstring addition.

Kladar:
I don't think it likes the hidden method `_handle_date` imported from utils. Fixing now.

gridstatus/caiso.py
```diff
@@ -1484,6 +1502,13 @@ def _get_historical(
     # but has nulls for all other columns
     df = df.dropna(subset=df.columns[1:], how="all")
 
+    # sometimes this data is from the previous day, in which case we want to label it as such
+    latest_file_time = check_latest_value_time(df, file)
+    current_caiso_time = pd.Timestamp.now(tz="US/Pacific")
```
WillKoehrsen, line 1507: Nit: `tz=self.default_timezone` to avoid using the string.

gridstatus/caiso.py
Kladar: yes saw that. Change was because the other imports were by function and that tends to be my preference, but will alter it back to avoid the code break.

Kladar: done.

pyproject.toml
WillKoehrsen: Should we move to only using `ruff`? It doesn't seem like you need both ruff and black.

gridstatus/caiso.py
Kladar: makes sense, though `_get_historical()` is not a class method with a self param and is not currently being passed the timezone. 

see "timezone" params just below it doing a similar str reference. 

```
df["Time"] = df["Time"].apply(
        _make_timestamp,
        today=date,
        timezone="US/Pacific",
    )
```

Kladar: Ah, should I do `CAISO.default_timezone`?

pyproject.toml
thehomebrewnerd: +1 to using only `ruff`. In a past life I've struggled with `ruff` and `black` wanting different things, so sticking with one tool seems good to me.

gridstatus/caiso.py
Kladar: Pushed, lmk what you think.

docs/pull_request_template.md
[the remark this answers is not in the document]
Kladar: Totally fair, I kind of the let copilot go ham on "make me a PR template" haha, can definitely pare it back. 

I kind of like Summary and Details for the reason I used it, like mention quality of life improvements in the summary and details them in details, but can definitely make it optional and lose the Resources and Checklist portions. I'll push another and you can let me know what you think.

gridstatus/utils.py
```diff
@@ -345,3 +345,9 @@ def move_cols_to_front(df, cols_to_move):
     for c in cols_to_move:
         cols.remove(c)
     return df[cols_to_move + cols]
+
+
+def check_latest_value_time(df: pd.DataFrame, value_col: str):
+    """Check the latest value time in the DataFrame"""
+    latest_time = df.loc[df[value_col].last_valid_index(), "Time"]
```
WillKoehrsen, line 352: Can this be simplified to `df['Time'].max()`?

pyproject.toml
WillKoehrsen: > I can make that change in the next PR.

That would be great, thanks!

gridstatus/caiso.py
```diff
@@ -1466,6 +1466,17 @@ def _get_historical(
     verbose: bool = False,
 ) -> pd.DataFrame:
 
+    """Get the historical data file from CAISO given a data series name, formats, and returns a pandas dataframe.
+
+    :param file: The name of the data we are wanting, which is equivalent to the file to get from CAISO
+    :type file: str
```
WillKoehrsen, line 1472: I don't have strong feelings about which to use, but I think we should standardize on function doc strings. [Here are a few options](https://stackoverflow.com/a/24385103/5755357) of which I slightly prefer Google's.

gridstatus/caiso.py
```diff
@@ -1484,10 +1495,17 @@ def _get_historical(
     # but has nulls for all other columns
     df = df.dropna(subset=df.columns[1:], how="all")
 
+    # sometimes this data is from the previous day, in which case we want to label it as such
+    latest_file_time = utils.check_latest_value_time(df, file)
+    current_caiso_time = pd.Timestamp.now(tz=CAISO.default_timezone)
+
+    if latest_file_time > current_caiso_time:
+        date = date - pd.Timedelta(days=1)
```
WillKoehrsen, line 1503: Do we need this logic for all datasets that use `_get_historical`, or only for the two

gridstatus/caiso.py
WillKoehrsen: > though _get_historical() is not a class method

My fault. Looks like you got it.

docs/pull_request_template.md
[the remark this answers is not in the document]
WillKoehrsen: The updated version is good. I like checklists provided all the items are actually necessary.

gridstatus/utils.py
Kladar: I don't think so (that was my initial approach), but typically there is a timestamp for every time in the day because it also has the day ahead forecast for the value. The file updates the value column incrementally for what we care about. So we need the time at which we have the latest value. `max()` would return the 23:55 timestamp each time. 

Example 
CleanShot 2024-09-30 at 15 09 35@2x

I haven't checked this comprehensively for all methods that use `_get_historical()` though (`fuelsource`, `demand`, `storage`), so I will do that now. `max()` may work for some, but conceivably this implementation handles both cases.

gridstatus/caiso.py
Kladar: Oop good catch, I also prefer google's (with types) and forgot to update my autodocstring extension. 

CleanShot 2024-09-30 at 15 24 15@2x

Highly recommend this though. https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

Kladar:
Also going to investigate these relevant tests a bit 
CleanShot 2024-09-30 at 15 39 37@2x

gridstatus/utils.py
WillKoehrsen: Thanks for the image, that's helpful. I couldn't figure out what `value_col` was being passed from the caller because we call this with `utils.check_latest_value_time(df, file)` where evidently `file` is a `str`.

Kladar: Okay, I feel this works now as a general solution for `storage`, `fuelsource`, and `demand`, which are the 3 files that take this data in. Hard to test outside of production workflows given the tendency to not work in the ~mornings and to know for sure, but methods are working.

CleanShot 2024-09-30 at 18 23 01@2x

@kmax12

Kladar: @WillKoehrsen yeah also took me a bit to realize `file` is meaning the file from caiso sans the .csv, in the request URL. 

Also I was assuming that the file and the value_col were going to be the same, but for instance the `demand` file has the column `Current demand`, so I wrote a small mapping dictionary in that util method to get the correct column name depending on the file. That feels incredibly brittle, but at least gets us off performing the morning data surgery.

gridstatus/utils.py
```diff
@@ -345,3 +345,16 @@ def move_cols_to_front(df, cols_to_move):
     for c in cols_to_move:
         cols.remove(c)
     return df[cols_to_move + cols]
+
+
+def check_latest_value_time(df: pd.DataFrame, file: str):
+    """Check the latest value time in the DataFrame"""
+    value_col_map = {
```
kmax12, line 352: i think it's more maintainable to add another parameter to `_get_historical` that is the column to use for this calculation.

we would then require all the callers of `_get_historical` (e.g `_get_historical_fuel_mix`, `_get_historical_load`) to provide the column they want to use. 

by doing this, if we were to add another endpoint like https://www.caiso.com/outlook/sp/history/20240929/rtm_forecast.csv, we wouldn't have to touch any of these lower level util functions

gridstatus/caiso.py
```diff
@@ -1484,10 +1496,17 @@ def _get_historical(
     # but has nulls for all other columns
     df = df.dropna(subset=df.columns[1:], how="all")
 
+    # sometimes this data is from the previous day, in which case we want to label it as such
+    latest_file_time = utils.check_latest_value_time(df, file)
+    current_caiso_time = pd.Timestamp.now(tz=CAISO.default_timezone)
+
+    if latest_file_time > current_caiso_time:
```
kmax12, line 1503: the tests are failing because the logic here is flawed. 

we only want to do this check if we are querying for the non-dated file.

gridstatus/utils.py
```diff
@@ -345,3 +345,16 @@ def move_cols_to_front(df, cols_to_move):
     for c in cols_to_move:
         cols.remove(c)
     return df[cols_to_move + cols]
+
+
+def check_latest_value_time(df: pd.DataFrame, file: str):
```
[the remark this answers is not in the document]
Kladar: done

gridstatus/caiso.py
Kladar: added logic to check only if it's the non-dated.

pyproject.toml
Kladar: Tracked in ENG-797

kmax12:
> 1 hr for checks is too long to be a useful iteration to test for code regressions in PRs.

a good release valve right now is to run the tests on the datasets your PR touches locally. that is pretty good for iterating quickly. 

getting tests running faster can be goal of future iteration. the tests take a long time because they depend on fetching data. any fast tests will have to bypass that as you suggest.

gridstatus/caiso.py
```diff
@@ -1463,14 +1464,28 @@ def _make_timestamp(time_str, today, timezone="US/Pacific"):
...
     date: str | pd.Timestamp,
+    column: str,
     verbose: bool = False,
 ) -> pd.DataFrame:
 
+    """Get the historical data file from CAISO given a data series name, formats, and returns a pandas dataframe.
+
+    Args:
+        file (str): The name of the data we are wanting, which is equivalent to the file to get from CAISO
+        date (str | pd.Timestamp): The date of the data to get from CAISO
```
kmax12, line 1475: missing the docstring for column

kmax12 (approved):
things are looking good. given that all the caiso tests are passing locally, im good to merge this in without waiting for github CI. then, we can update our internals systems to use this version of `gridstatus`. thanks for fixing!

WillKoehrsen (approved):

gridstatus/utils.py
WillKoehrsen: I see this was addressed by adding a `column` parameter to the function which makes sense.

WillKoehrsen:
For future PRs, can we Squash and Merge?

kmax12:
I updated repo settings to only allow squash merging now. dont know why it wasnt previously set up like that

image

Kladar:
I will tend to delete my branches once they are merged, which can be an automatic setting as well.

PR #446, opened 2024-10-03
feat[ENG 256] Add 5 Minute PJM Solar and Wind Forecast

Kladar:
## Summary
This adds the PJM 5 Minute Wind and PJM 5 Minute Solar Forecasts to the `PJM()` `get_solar_forecast()` and `get_wind_forecast()` methods by introducing the `resolution` kwarg, which can take either `5min` or `hourly`, with the default being `hourly` to maintain backward compatibility. 

It also dives into testing as it relates to PJM, slowly starting to label and separate unit tests from integration tests.

#### Data
This data is nearly identical to the hourly data, such that the addition was fairly straightforward and only required adding a few lines of code. You can see a JSON example of it in the added fixtures. This felt quite straightforward to implement.

#### Testing
This was a bit more of a rabbit hole that I wanted to understand. A lot of this is an initial pass at thinking out loud about the topic. The PJM tests run in about 3-8 minutes locally when ignoring the slow tests, which is manageable for local development. However, I still think it'll be a good idea to thoughtfully start to separate out categories of tests to run at specific times for clarity, developer velocity, comprehensiveness, etc. Namely

1. Integration tests - these are tests that query an external service, such as the PJM 5 minute solar forecast. These are ideal for making sure services are up (or not) and returned data is in the correct format (or not). Integration tests will see if an external service's data schema or column headers have changed, for instance. They are typically slow, something that you might not want to run on every commit. 

2. Unit tests - these test an internal unit of our code, the primary purpose being to avoid regressions and not allow runtime errors in production. These go hand-in-hand with type safety and static checking. They also recursively push us to write code that is easy to test. Can two numbers be multiplied together? Can that column be renamed? Can I parse this file into an in-memory object? etc. They are typically very fast, and can be run automatically on every commit without too much trouble. They can also rely on static fixtures of data, removing the external API dependency.

These are the two I start to label and develop here, offering a fast unit-test-only `make` command and marking the tests that query an external service as `integration`. This will be an incremental process, but this gives us something to aim at I think.  

The third test not touched on in this PR as much are
3. Data tests - these test the format and comprehensiveness of the data itself, and are a bit fuzzier across integration and unit testing. Data testing can happen against any store of data, not just externally returned data, and also plays a role further downstream. But it can be useful to check for gaps in incoming data, to catch such a thing as far upstream as possible. Data testing can also be compute intensive though, which means it potentially slows down data that is checked in-flight. Here they are mostly coupled to the integration tests. 

#### Fixtures
These fixtures are probably a bit large to be checking into source control, but that is also a bit in the eye of the beholder. These are simply the output of a simple call to the `get_forecast()` methods, like:

```
df = pjm.get_wind_forecast(date="2024-10-01", resolution="5min")
json_data = df.to_json(orient='records', date_format='iso')
with open('gridstatus/tests/fixtures/pjm/pjm_5min_wind_forecast_sample.json', 'w') as f:
    f.write(json_data)
```

I imagine we update the fixtures periodically, either when running integration tests (they could update on passing) or on a periodic schedule. Also we'll want to update them whenever there is a schema change coming form the API of course. 

### Conclusion
Obviously the lines are fuzzy, and there is a balance between speed and comprehensiveness of a test. One could definitely argue a different way, like "I want to know everything about the API and the data coming out of it before storing anything" which is more similar to the structure now. Nothing in this PR is definitive, but will be nice to move in a direction that tests the things we want to test when we want to test them, to avoid regressions and keep developer velocity in mind. Keen for all of your thoughts on the data querying and the future testing direction!

Local tests are passing. 
CleanShot 2024-10-02 at 16 27 24@2x


gridstatus/pjm.py
```diff
@@ -1133,17 +1133,41 @@ def get_interconnection_queue(self, verbose=False):
         return queue
 
     @support_date_range(frequency=None)
-    def get_solar_forecast(self, date, end=None, verbose=False):
+    def get_solar_forecast(
```
WillKoehrsen, line 1136: We aren't consistent in this across the codebase, but there's a question here whether we should use one method with an extra parameter or separate `get_solar_forecast_hourly` and `get_solar_forecast_5_min` methods. I'm in favor of separate methods to maintain a consistent interface with our `get_` methods.

gridstatus/tests/test_pjm.py
```diff
@@ -457,6 +526,109 @@ def _check_solar_forecast(self, df):
...
+        assert isinstance(df, pd.DataFrame)
+        assert not df.empty
+        self._check_solar_forecast(df)
+
+    def test_get_solar_forecast_unit_5min(self, sample_5min_solar_forecast_data):
+        with mock.patch.object(
+            self.iso,
+            "get_solar_forecast",
+            return_value=sample_5min_solar_forecast_data,
+        ):
```
WillKoehrsen, line 576: It's not clear to me what this is testing. Since the function is mocked with the fixture, any changes to the function are not tested.

gridstatus/pjm.py
kmax12: i agree. i think one method per source endpoint is highly preferable for maintenance. 

it's nice to know exactly what endpoint/dataset is going to hit without having to understand parameters. 

I wish we never combined endpoints for things like the `get_lmp` method

gridstatus/tests/test_pjm.py
Kladar: Yeah true. It's poorly named for what it does now, since it was adapted from the prior integration-style test that did query the external service into simply not doing that. But this is more testing the fixture properties are what they say they are. Not super useful, but I guess not entirely useless. Mostly wanted this is to start introducing fixtures to test_pjm.py more generally and adapting the existing tests to them, with the assumption that they will evolve considerably from here.

We could remove it if we aren't interested in testing the fixture properties (but the test in only a few milliseconds).

gridstatus/pjm.py
Kladar: sounds good, updated.

gridstatus/tests/test_pjm.py
WillKoehrsen: It would make more sense if we mocked `_get_pjm_json` with a fixture and checked that `get_solar_forecast` correctly handles the output. I think if we use mocking, we should get as close to the raw output from the external API as we can so our unit tests are checking the data parsing.

My main issue with only testing the fixture properties is it could give a false sense of security. If the source code changes but the developer doesn't update the fixture, the unit test will still pass even if the source code introduces an error.

Kladar: yes agreed, you read my mind on `_get_pjm_json()`, which is what I really want to unit test (as well as `_parse_solar_forecast` etc).  Was having some trouble getting at it yesterday and wanted to get the new forecast query logic out to you before too long and not spend any more time in the testing trenches, but next pass was going to be to get at `_get_pjm_json`. I can get there for this PR though, if we don't mind waiting a few more hours.

gridstatus/pjm.py
```diff
@@ -1158,7 +1171,46 @@ def get_solar_forecast(self, date, end=None, verbose=False):
 
         return self._parse_solar_forecast(df)
 
-    def _parse_solar_forecast(self, df):
+    @support_date_range(frequency=None)
+    def get_solar_forecast_5min(
```
kmax12, line 1175: in the rest of the code base we typically add a space to `5_min`, so it is `get_solar_forecast_5_min`

gridstatus/tests/test_pjm.py
```diff
@@ -13,6 +13,51 @@
 from gridstatus.tests.decorators import with_markets
 
 
+# Fixtures for sample data
+@pytest.fixture
+def sample_5min_solar_forecast_data():
+    fixture_path = "gridstatus/tests/fixtures/pjm/pjm_5min_solar_forecast_sample.json"
```
kmax12, line 19: > These fixtures are probably a bit large to be checking into source control, but that is also a bit in the eye of the beholder.

what would be the other easy option? a public s3 bucket?

that seems like it would be pretty straightforward to set up.

gridstatus/tests/test_pjm.py
kmax12: I think we all our in agreement, but I'll just repeat it....we should fixtures as close as possible to the exact data that comes back from the source. anything else is embedding our transformation logic into the fixture. and our goal is to test the transformation logic. 

what I'll add is, maybe we should want to think about structuring each dataset's scraper a bit differently, so we can automatically reason more about each step/mock different steps

```python

class DatasetBase:
    def run(self):
        data = self.fetch_raw_data()
        data = self.process_raw_data(data)
        data = self.finalize_data(data)
        return data
    
class PJMBase(DatasetBase):
    """"Reusable pjm methods""""
    def query_pjm(self, url):
        pass

    def process_pjm_response(self, response):
        pass

class PJMSolarForecastHourly(PJMBase):
    def fetch_raw_data(self):
        """returns data as raw as possible from source ie json, file, etc"""
        return self.query_pjm(...)

    def process_raw_data(self, data):
        """standardized processing of raw data across many datasets"""
        return self.process_pjm_response(data)
        
    def finalize_data(self, data):
        """dataset specific processing"""
        return data
```

I dont know if this is right approach but just want to through something out there


practically speaking, i think this is a bigger discussion, so no need to implement too much yet. i think we merge this dataset in and then set up time to discuss the ideal end state across repo

kmax12 (approved):
code for scraper itself seems good, my approval is here once all comments are addressed

gridstatus/tests/fixtures/pjm/pjm_5min_solar_forecast_sample.json
WillKoehrsen, line 1: Can we compress these fixtures? If we are going to keep them in git, we should keep them as small as possible to minimize the repo size.

WillKoehrsen: Or at least reduce the size to where they can be viewed on GitHub.

gridstatus/tests/test_pjm.py
Kladar: Yeah I agree with all of the above, and gave it a day of trying to test things in a way that worked with the existing structure in case it was straightforward and extensible across the repo, and struggled to come up with anything that was compelling. Found it hard to create fixtures for the _general_ methods (`_get_json`, `_get_pjm_json`) that can then be reused in the _specific_ methods (`get_solar_forecast_5_minute`) without heavily duplicating/repeating the build up methods or heavily parameterizing the fixtures and having conditional logic to the specific method tests. 

Tried 3-4 approaches and didn't find anything I'd put my stamp on. I built up some new tests for the general methods with the fixtures and left it at that, and we can revisit.

gridstatus/tests/fixtures/pjm/pjm_5min_solar_forecast_sample.json
Kladar: Yep agreed - commenting here from all-hands so we have it: plan is to leave behind a greatly truncated json to keep the `_get_pjm_json` test and then we can revisit.

gridstatus/tests/test_pjm.py
Kladar: yep I think that's the move. Probably in the future the test harness setup would check if you have the latest fixture zipped file locally, if so use it, if not download it from s3. (And we add that directory to the gitignore).

gridstatus/pjm.py
Kladar: done

gridstatus/tests/fixtures/pjm/pjm_5min_solar_forecast_response.json
Kladar, line 1: @WillKoehrsen I'll leave this truncated PJM API response output here to test `_get_pjm_json` if that's alright, until we get a file-based fixture system up and running.

gridstatus/tests/test_pjm.py
```diff
@@ -440,8 +464,64 @@ def test_pjm_update_dates(self):
...
+                pd.DatetimeTZDtype,
+            ), f"{col} is not a timezone-aware datetime column"
+            assert str(result[col].dt.tz) == str(
+                self.iso.default_timezone,
+            ), f"{col} timezone doesn't match the default timezone"
+
+    # TODO(kladar) Finding it very difficult to build up the tests and fixtures to test the logic beyond _get_pjm_json.
+    # I've tried 3 or 4 approaches and nothing seemed not terrible, so I don't want to spend too much more time on it at this stage.
+    # I'm sure it'll become clear with time what the best way to do it is.
+    # NOTE(kladar): The following are the old tests, which are more integration tests pulling actual data from the API"""
```
Kladar, line 524: Here's the new test, with a fixture developed from the raw PJM API response and the `_get_json` output mocked for testing `_get_pjm_json`, which is the only new unit test.

gridstatus/tests/test_pjm.py
Kladar: Captured this in ETL v3, resolving.

Kladar:
Local Passing, merging 
CleanShot 2024-10-04 at 09 58 19@2x

gridstatus/tests/fixtures/pjm/pjm_5min_solar_forecast_response.json
WillKoehrsen: Yep, this works.

WillKoehrsen:
> The PJM tests run in about 3-8 minutes locally when ignoring the slow tests, which is manageable for local development.

Just wondering, were you running the tests in parallel? This command takes 90 seconds locally for me:

```
poetry run pytest gridstatus/tests/test_pjm.py -m "not slow" -n 8
```

PR #448, opened 2024-10-04
[ENG-803] Exclude Tests from Package

WillKoehrsen:
## Summary
- Excludes `gridstatus.tests` from the package

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,13 @@ maintainers = [
 keywords = ["energy", "independent system operator"]
 license = "LICENSE"
 
+[tool.setuptools]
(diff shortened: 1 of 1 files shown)


WillKoehrsen:
I'm not sure if including tests in the package is a best practice or not. Both `numpy` and `pandas` do include tests, while `requests` does not.

[This](https://stackoverflow.com/a/17722381/5755357) is a decent argument for including the tests.

WillKoehrsen:
Closing because we decided this was unnecessary work. The main reason for excluding the test directory was to avoid distributing fixtures needed for tests, but, we are not planning on storing these in the repo.

PR #449, opened 2024-10-05
Fix ERCOT API Hourly Solar and Wind Tests

WillKoehrsen:
## Summary
- Fixes the hourly solar and wind report tests for `ErcotAPI`

### (*Optional*) Details
- Tests were failing after the changes from https://github.com/gridstatus/gridstatus/pull/432 and due to changes in the data

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -295,10 +295,6 @@ class ErcotAPI:
 
         data = Ercot()._rename_hourly_wind_or_solar_report(data)
 
-        # Add HSL SYSTEM WIDE if it is not in the data (older data may not have it)
-        if "HSL SYSTEM WIDE" not in data:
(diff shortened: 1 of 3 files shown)


gridstatus/ercot_api/ercot_api.py
```diff
@@ -295,6 +295,10 @@ def _handle_hourly_wind_report(self, data, verbose=False):
 
         data = Ercot()._rename_hourly_wind_or_solar_report(data)
 
+        # Add HSL SYSTEM WIDE if it is not in the data (older data may not have it)
```
kmax12, line 298: do we want to put this in `_rename_hourly_wind_or_solar_report` so we've centralized all this standardization logic?

kmax12 (approved):

gridstatus/ercot_api/ercot_api.py
WillKoehrsen: Yes, good idea, updated.

PR #451, opened 2024-10-08
fix[ENG 810] Resolve undated times correctly and bust cloudfront cache

Kladar:
## Summary
Makes sure we're getting todays date in the CAISO timezone. Before it was returning the UTC date and pairing it with the local time, causing a silent failure.

Also adds the epoch current epoch time to the URL to bust the cloudfront cache on file return.

### (*Optional*) Details
Moves _make_timestamp to caiso_utils, updates check_latest_value_time, etc. Hard to test for sure without replicating the system where pandas is running as well as running it at the times where it becomes an issue, but makes sense from a first principles approach.

Local tests passing.
CleanShot 2024-10-08 at 10 30 39@2x

### DIFF
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -1466,13 +1466,14 @@ def _get_historical(
     Returns:
         pd.DataFrame: A pandas dataframe of the data
     """
-
+    # NOTE: The cache buster is necessary because CAISO will serve cached data from cloudfront on the same url if the url has not changed.
+    cache_buster = int(pd.Timestamp.now(tz=CAISO.default_timezone).timestamp())
     if utils.is_today(date, CAISO.default_timezone):
-        url: str = f"{_BASE}/{file}.csv"
+        url: str = f"{_BASE}/{file}.csv?_={cache_buster}"
         latest = True
     else:
         date_str: str = date.strftime("%Y%m%d")
-        url: str = f"{_HISTORY_BASE}/{date_str}/{file}.csv"
+        url: str = f"{_HISTORY_BASE}/{date_str}/{file}.csv?_={cache_buster}"
         latest = False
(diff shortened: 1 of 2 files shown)


kmax12 (approved):
just one comment otherwise, looks good

gridstatus/caiso.py
```diff
@@ -1479,13 +1466,14 @@ def _get_historical(
...
+    # NOTE: The cache buster is necessary because CAISO will serve cached data from cloudfront on the same url if the url has not changed.
+    cache_buster = int(pd.Timestamp.now(tz=CAISO.default_timezone).timestamp())
     if utils.is_today(date, CAISO.default_timezone):
-        url: str = f"{_BASE}/{file}.csv"
+        url: str = f"{_BASE}/{file}.csv?_={cache_buster}"
         latest = True
     else:
         date_str: str = date.strftime("%Y%m%d")
-        url: str = f"{_HISTORY_BASE}/{date_str}/{file}.csv"
+        url: str = f"{_HISTORY_BASE}/{date_str}/{file}.csv?_={cache_buster}"
```
Kladar, line 1476: @kmax12

kmax12: missed this. thanks!

gridstatus/caiso_utils.py
```diff
@@ -1,6 +1,21 @@
...
+        month=today.month,
+        day=today.day,
+        hour=hour,
+        minute=minute,
+    )
+    ts = ts.tz_localize(timezone, ambiguous=True)
+    return ts
+
+
 def check_latest_value_time(df: pd.DataFrame, column: str):
```
WillKoehrsen, line 19: It seems like the name of this function no longer represents what the function does. Something like `get_latest_time_value_on_today` is more accurate since it returns a timestamp from today.

PR #452, opened 2024-10-09
[ENG 809] Add ISO New England API Integration and Initial Datasets

Kladar:
## Summary
This adds an integration to the ISONE API, based on the existing work for the ERCOT API. It also adds the datasets for `realtimehourlydemand` and `dayaheadhourlydemand`, as well as continuing quality-of-life improvements and tests with fixtures mocking API responses. 

Subject to change, so I welcome your comments on data column names, dataframe formats, test approach, etc! 

#### API
The API mostly acts as expected, but has some quirks that I will talk about here and that we will continue to explore more as we add more datasets. Hopefully this puts us on a good footing to add new ones quickly. 

First, there are some methods for getting the `current` and `current_static` dataset, which in both cases is just the current or latest interval available. Since most applications with want more than this single interval, not a huge emphasis will be put on this. One thing to note is that the reason I call it `static` when no dates or locations are given is that it will still return the interval for all of the regional zones in ISONE, like below.

`dayahead`
CleanShot 2024-10-09 at 15 38 15@2x

Not exactly sure when `dayahead` gets published, but it is currently just before 7PM ET, October 9th and the endpoint is still returning October 9th results and in fact they are currently in the past. It is likely that day ahead results publish in the late evening for the following day, but need more observations to know. 

`realtime`
CleanShot 2024-10-09 at 15 39 25@2x

Also noting that the `realtime` data has always been 3-5 days behind since I've been querying it.  

Data goes back to 2017-07-01 for both endpoints it seems. I'm not sure if that is a rolling window, but my guess is no because of the niceness of the date and the API taking over from the raw file downloads from ISONE's perspective. So my guess is this data will be available from that date onward, and potentially for other datasets as well. 

Second, the `locations` endpoints are included since the API takes their `locationId`, which is opaque unless you know it. In the future, getting the `locations_all` and then filtering for the ID you want might make sense, for larger datasets. For now there is a dictionary included with known zones and their IDs, since we are focusing on the zones for load as opposed to nodal information etc. `.H.INTERNALHUB` and `NEPOOL` are not available from the realtime dataset and are excluded when looking up IDs.

#### Tests
I created fixtures for the likely endpoints that hopefully are big enough to capture API output behavior but small enough to check in to source control (for now). In the future it is likely that the test fixtures will be moved off to s3 and downloaded at test runtime if needed.

Testing approach is still evolving in the repository, but as we can see these tests are unit and run very quickly 
CleanShot 2024-10-09 at 15 49 07@2x


gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,439 @@
...
+    "INTERNALHUB": 4000,
+    "MAINE": 4001,
+    "NEWHAMPSHIRE": 4002,
+    "VERMONT": 4003,
+    "CONNECTICUT": 4004,
+    "RHODEISLAND": 4005,
+    "SEMASS": 4006,
+    "WCMASS": 4007,
+    "NEMASSBOST": 4008,
+}
```
Kladar, line 29: These were taken from the original POC code. We could have them match the API response verbatim (e.g. `.Z.MAINE`), leave them as is in this current cleaned up state, or some third option. The API really only cares about the `int` `Id` though.

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,439 @@
...
+                response.raise_for_status()
+
+        if parse_json:
+            return response.json()
+        else:
+            return response.content
+
+    def get_locations(self) -> pd.DataFrame:
+        """
+        Get a list of all locations.
```
WillKoehrsen, line 142: Is this all locations or is it only "core" locations? 

> Get list of core locations (Hub, Load Zones, External Nodes)

https://webservices.iso-ne.com/docs/v1.1/rest.locations.html

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,439 @@
...
+        """
+        df["BeginDate"] = pd.to_datetime(df["BeginDate"])
+        df["Load"] = pd.to_numeric(df["Load"], errors="coerce")
+        df["LocId"] = pd.to_numeric(df["LocId"], errors="coerce")
+        df = df.rename(columns={"BeginDate": "Interval Start"})
+        return df[["Interval Start", "Location", "LocId", "Load"]]
+
+    def get_realtime_hourly_demand_current_static(self) -> pd.DataFrame:
+        """
+        Get the most recent real-time hourly demand data for default locations.
```
WillKoehrsen, line 212: What are the "default locations"?

gridstatus/gs_logging.py
```diff
@@ -1,3 +1,24 @@
+import logging
+
+
+def setup_gs_etl_logger():
+    logger = logging.getLogger("gs_etl")
+    logger.setLevel(logging.INFO)  # Set the logger level to DEBUG
```
WillKoehrsen, line 6: Nit: comment needs updating.

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,439 @@
...
+            data = response["HourlyRtDemand"]
+            data["Location"] = location
+            data["LocId"] = location_id
+            all_data.append(data)
+
+        df = pd.DataFrame(all_data)
+        return self._handle_demand(df)
+
+    @support_date_range(frequency="D")
+    def get_realtime_hourly_demand_historical_range(
```
WillKoehrsen, line 280: Instead of separate methods for `current` and `historical_range` data, I think we should have one method, `get_realtime_hourly_demand`. If `date == "latest"` then we use the current endpoint, otherwise the historical endpoint. That matches what we do in our other ISOs.

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,439 @@
...
+            df (pd.DataFrame): Input DataFrame with demand data.
+
+        Returns:
+            pd.DataFrame: Processed DataFrame.
+        """
+        df["BeginDate"] = pd.to_datetime(df["BeginDate"])
+        df["Load"] = pd.to_numeric(df["Load"], errors="coerce")
+        df["LocId"] = pd.to_numeric(df["LocId"], errors="coerce")
+        df = df.rename(columns={"BeginDate": "Interval Start"})
+        return df[["Interval Start", "Location", "LocId", "Load"]]
```
kmax12, line 208: we also want to add "Interval End". we do that across all datasets.

we also should rename `LocId` to `Location Id`. we like to avoid abbreviations and by adding space, the column name works better with case-insentive formatting

gridstatus/isone_api/isone_api.py
Kladar: Ah yeah this is a typo, gets a list of core locations. the locations_all endpoint I believe gets "everything", meaning nodal locations since it returns about 1300 locations. 

Updated.

gridstatus/isone_api/isone_api.py
Kladar: default locations for `dayahead` are in the screenshots in the PR summary, and captured in the `ZONE_LOCATIONID_MAP` (without their prefixes). Subtract the `EXCLUDE_FROM_REALTIME_HOURLY_DEMAND` to get the realtime defaults. I suppose that could have been clearer. 

Not sure what all we are going to want from them, this `static` method, or really the `current` methods in general, since they only return one interval.

gridstatus/isone_api/isone_api.py
Kladar: I see what you mean. The `current` methods only return a single hourly interval though, rather than a full day. I was thinking we would probably refrain from using them entirely, since it's tricky to time cron jobs or upsert or otherwise not miss intervals if a downstream system fails. 

I could definitely see removing the `current` methods entirely (but they might be useful). But heard about having a `latest` functionality that uses the `day/{date}/` endpoints, I can work on that.

gridstatus/gs_logging.py
```diff
@@ -1,3 +1,24 @@
...
+        ),
+    )
+    logger.addHandler(handler)
+    return logger
+
+
+logger = setup_gs_etl_logger()
+
+
 def log(msg, verbose=False):
```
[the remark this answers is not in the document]
Kladar: Definitely can, though it might be a separate PR. There are about 200+ places where the existing `log` is called, mostly with two arguments `(msg, verbose)`, so I didn't want to get too far afield from the task or break a bunch of other things not related to ISO NE API by doing so. 

[I made a ticket for that change here](https://www.notion.so/gridstatus/Update-gridstatus-to-use-detailed-logger-11be835f42aa808f82c3d16d7365abd1?pvs=4). 

Could also make an argument for a larger change, going to a structured logging system like `structlog` but I'm fine without that tbh.

gridstatus/gs_logging.py
```diff
@@ -1,3 +1,24 @@
+import logging
+
+
+def setup_gs_etl_logger():
```
WillKoehrsen, line 4: Function name?

gridstatus/isone_api/isone_api.py
WillKoehrsen: I don't think we need separate `_static` methods. We can add a `locations` argument to `get_realtime_hourly_demand`.

gridstatus/isone_api/isone_api.py
WillKoehrsen: > The current methods only return a single hourly interval

Multiple of our methods only return a single interval when you pass `date="latest"`. For example, `MISO().get_fuel_mix("latest")` returns a single interval.

gridstatus/gs_logging.py
Kladar: made a ticket to address the logging separately than this PR. https://www.notion.so/gridstatus/Update-gridstatus-to-use-detailed-logger-11be835f42aa808f82c3d16d7365abd1?pvs=4

gridstatus/isone_api/isone_api.py

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,439 @@
...
+                url,
+                params=api_params,
+                auth=(self.username, self.password),
+                headers=headers,
+            )
+
+            retries += 1
+            if response.status_code == 200:
+                break
+            elif response.status_code == 429 and retries <= self.max_retries:
```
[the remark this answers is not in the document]
Kladar: There is pretty good information about the datasets but not much that describes the behavior of the API itself. Couldn't find a rate limit anywhere https://www.iso-ne.com/participate/support/web-services-data

gridstatus/gs_logging.py
WillKoehrsen: Sorry, I meant the function name should be updated for this repository.

Kladar: oh yeah, got it, done.

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,381 @@
...
+            end_date (str | None): The end date for the data request. Only used if date is not "latest".
+            locations (list[str], optional): List of specific location names to request data for.
+                                             If None, data for all locations will be retrieved.
+
+        Returns:
+            pandas.DataFrame: A DataFrame containing the real-time hourly demand data for all requested locations.
+        """
+        all_data = []
+
+        match (date, locations):
```
WillKoehrsen, line 240: Interesting approach with the match case statement. It's concise; my only concern is it's not consistent with the rest of the codebase.

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,381 @@
...
+                        not location_id
+                        or location in EXCLUDE_FROM_REALTIME_HOURLY_DEMAND
+                    ):
+                        raise ValueError(
+                            f"{location}: Not a known ISO NE Hub or Zone for this data",
+                        )
+
+                    url = f"{BASE_URL}/realtimehourlydemand/day/{date.strftime('%Y%m%d')}/location/{location_id}"
+                    response = self.make_api_call(url)
+                    print(response)
```
WillKoehrsen, line 287: Should we switch this to a log? Or maybe remove if this was for debugging.

Kladar: yeah debugging. Maybe I will re-request review once ready, I was going to add a few more test cases to cover more of the code paths in the methods.

gridstatus/isone_api/isone_api.py
Kladar: It's kind of the endpoints upstream that cause lead me to this approach, since they offer the endpoints like

```
/dayaheadhourlydemand/current # this gets all core location latest, seems most useful
/dayaheadhourlydemand/current/location/{locationId}] # this is the above but requires an explicit location
/dayaheadhourlydemand/day/{day}/location/{locationId} # this is for historical, also requires explicit location
```

Conceivably I could go down to just two cases and use the latter two endpoints, since we would emulate that first endpoint but just use the explicit locations and loop through them ourselves. Essentially treat the first two match cases the same and consolidate. But that first endpoint seems the most useful for what we are probably wanting to do with the data so it felt strange to not use it.

Kladar:
Going to add some more tests to cover the other code paths, but those notwithstanding ready for another review.

gridstatus/isone_api/isone_api.py
Kladar: Ah yeah, either works. Once I find myself putting `and` and conditional statements in my ifs

```
if date="latest" and not locations:
``` 
or start nesting 
```
if date = latest: 
          if locations:
``` 
I tend to switch to match-case, but this one could probably go either way for sure.

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,439 @@
...
+        all_data = []
+
+        for location in locations:
+            location_id = ZONE_LOCATIONID_MAP.get(location)
+            if not location_id:
+                log.warning(f"{location}: Not a known ISO NE Hub or Zone for this data")
+                continue
+
+            url = f"{BASE_URL}/dayaheadhourlydemand/day/{date.strftime('%Y%m%d')}/location/{location_id}"
+            response = self.make_api_call(url)
```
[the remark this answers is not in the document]
Kladar: I believe I cover this now

gridstatus/isone_api/isone_api.py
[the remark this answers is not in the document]
kmax12: we're you actually hitting issues with rate limit, or did you just include this logic to be safe?

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,379 @@
...
+        url: str,
+        api_params: dict = None,
+        parse_json: bool = True,
+        verbose: bool = False,
+    ):
+        log.debug(f"Requesting url: {url} with params: {api_params}")
+        retries = 0
+        delay = self.initial_delay
+        headers = {"Accept": "application/json"}
+        while retries <= self.max_retries:
```
kmax12, line 104: is this retry logic mostly copied from `ercot_api.py`? if so, we probably should abstract it out.

im okay if we want to leave that as a TODO./ followup task

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,379 @@
...
+        Process demand DataFrame: convert types, rename columns, and add Interval End.
+
+        Args:
+            df (pd.DataFrame): Input DataFrame with demand data.
+            interval_minutes (int): Duration of each interval in minutes. Default is 60.
+
+        Returns:
+            pd.DataFrame: Processed DataFrame.
+        """
+        df["Interval Start"] = pd.to_datetime(df["BeginDate"])
```
kmax12, line 211: i think we want to do `pd.to_datetime(df["BeginDate"]).dt.tz_convert(self.default_timezone)` to make sure this gets but into the right timezone.

it's minor difference but it takes it from the absolute `-4` offset to the named `US/Eastern` that respects DST.

The situation where this matter is when the interval end is on the other side of a DST change, so having the fixed offset is confusion. 

image

gridstatus/isone_api/isone_api.py
kmax12: im fine with match-case. it's clear what's going on, even if we don't typically use it.

gridstatus/isone_api/isone_api.py
```diff
@@ -0,0 +1,379 @@
...
+        if not all([self.username, self.password]):
+            raise ValueError(
+                "Username and password must be provided or set as environment variables",
+            )
+
+        self.sleep_seconds = sleep_seconds
+        self.initial_delay = min(max(0.1, sleep_seconds), 60.0)
+        self.max_retries = min(max(0, max_retries), 10)
+
+    def _handle_end_date(
```
kmax12, line 67: is `_handle_end_date` used anywhere?

kmax12 (approved):

gridstatus/isone_api/isone_api.py
Kladar: you're right, not anymore, removed.

gridstatus/isone_api/isone_api.py
Kladar: yeah copied over from ERCOT, marked in there with a TODO to consolidate and [made a ticket](https://www.notion.so/gridstatus/Consolidate-API-Handling-Methods-11ce835f42aa8043bfa0def723b4a45d?pvs=4).

gridstatus/isone_api/isone_api.py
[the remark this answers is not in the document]
Kladar: yeah I started with the `ercot_api` logic and took some of the things that might be useful when I started out. DIdn't hit any limits that I could tell (but also didn't stress test it very much).

gridstatus/tests/test_isone_api.py
WillKoehrsen, line 1: Are we not including any integration tests (that hit the API)?

Kladar: I didn't include any here just because I think it's worth think about them separately and running them in a separate way from the unit testing. I think of unit tests running almost as a pre-commit hook that runs super fast, and integration tests as something that happens continually for external services (and before deployment for internal ones). I will write up my thoughts more concretely in Notion I think.

I have been thinking about it some, and think there might be two main elements that I would want to test about external services: 
1. an API status page that sends a healthcheck request to a simple endpoint (and we can alert when one is down or unhealthy/latent)
2.  a top-to-bottom dataset integration test that runs occasionally (and possibly updates the fixtures, runs units on them, units tell us about schema changes/downstream breakages).

Kladar:
Added one more test to handle the multiple location call, which was more trouble than expected. But should be pretty good now. Testing approach will evolve.

PR #454, opened 2024-10-14
fix[ENG 809] Fix ISONE API methods

Kladar:
## Summary
Needs to be `end` not `end_date` to be picked up by the decorator. 

### (*Optional*) Details
Need to support the `@support_date_range` decorator.


kmax12:
this normally something we'd add a test for. I thought [this test](https://github.com/gridstatus/gridstatus/blob/a22b64c232517e26ebc9a91a20bffd9e9e16e5b7/gridstatus/tests/test_isone_api.py#L204-L205) covered it, but i see that actually only check one day. 

I would add a test here

gridstatus/isone_api/isone_api.py
```diff
@@ -195,8 +195,8 @@ def _handle_demand(
     @support_date_range("DAY_START")
     def get_realtime_hourly_demand(
         self,
-        date,
-        end_date: str | None = None,
+        date: str | pd.Timestamp = "latest",
+        end: str | pd.Timestamp | None = None,
```
WillKoehrsen, line 199: Can we add a `verbose` parameter to these methods to be used for logging like in our other `get_` methods?

Kladar: yes, but I'll state that I'm skeptical of the addition of arguments that are only touched by decorators and not in the function itself. If there is some precedent/resource supporting this approach in python I would love to read it. IMO those should be parameters of the decorator rather than the function. These will likely fail static type checking and have to be manually ignored.

WillKoehrsen:
> I thought [this test](https://github.com/gridstatus/gridstatus/blob/a22b64c232517e26ebc9a91a20bffd9e9e16e5b7/gridstatus/tests/test_isone_api.py#L204-L205) covered it, but i see that actually only check one day.

I always forget if `support_date_range` is inclusive of the end or not. I find [this logic](https://github.com/gridstatus/gridstatus/blob/a22b64c232517e26ebc9a91a20bffd9e9e16e5b7/gridstatus/decorators.py#L166-L211) hard to reason through without just calling the method and checking.

Kladar:
> this normally something we'd add a test for. I thought [this test](https://github.com/gridstatus/gridstatus/blob/a22b64c232517e26ebc9a91a20bffd9e9e16e5b7/gridstatus/tests/test_isone_api.py#L204-L205) covered it, but i see that actually only check one day.
> 
> I would add a test now that we know we missed it once

Yep good call. I've added some generic fixtures for these endpoints that hopefully support both unit testing of the multiday functionality and easy updating from a semi-regular integration test. I can feel us getting closer to a more generic test harness through this as well.

gridstatus/isone_api/isone_api.py
```diff
@@ -231,9 +232,6 @@ def get_realtime_hourly_demand(
                         )
                     url = f"{BASE_URL}/realtimehourlydemand/current/location/{location_id}"
                     response = self.make_api_call(url)
-                    from pprint import pprint
-
-                    pprint(response)
```
Kladar, line 236: accidently left this in last time. I have an idea of a better way to create fixtures in-situ now as well.

gridstatus/isone_api/isone_api.py
WillKoehrsen: `verbose` isn't used by the `support_date_range` decorator. It should be used within the function itself to show information about the URLs being retrieved. Take a look for example at any of the `get_` methods in `isone.py`.

Kladar: ah got it. Yeah to me that's what logging levels should be used for.

WillKoehrsen: I agree there's a better way, but I'm not sure what it looks like with logging levels. If I'm in the interpreter running this function, how would I turn on/off logging if we used logging levels? Right now, I do:

```
# Logging
ISONEAPI().get_realtime_hourly_demand(date, end, verbose=True)
# No logging
ISONEAPI().get_realtime_hourly_demand(date, end, verbose=False)
```

Kladar: You import the logger and change the level in the interpreter/REPL/script
```
from gridstatus.gs_logging import logger
```

Example `INFO`
```
logger.setLevel(logging.INFO)
df = api.get_realtime_hourly_demand(date="2024-10-06", end="2024-10-08", locations=[".Z.MAINE", ".Z.NEWHAMPSHIRE"])
```
CleanShot 2024-10-14 at 18 09 27@2x

Example `DEBUG`
```
logger.setLevel(logging.DEBUG)
df = api.get_realtime_hourly_demand(date="2024-10-06", end="2024-10-08", locations=[".Z.MAINE", ".Z.NEWHAMPSHIRE"])
```
CleanShot 2024-10-14 at 18 11 01@2x

Kladar: I pushed a commit to make the logger a bit more dynamic in setting a level at a script level.

I'm still getting double output on certain log calls, can't quite figure out why that's occurring only some of the time, but it is likely a less than ideal import somewhere I haven't tracked down yet, but will once I get to the logging ticket.

CleanShot 2024-10-14 at 18 14 43@2x

Kladar: Going to remove those example logs from the method. It appears that the DEBUG setting already outputs the URL.

WillKoehrsen: Thanks, this makes sense.

WillKoehrsen (approved):

gridstatus/isone_api/isone_api.py
```diff
@@ -355,4 +377,8 @@ def get_dayahead_hourly_demand(
             )
 
         df = pd.DataFrame(all_data)
+        # NOTE(kladar): 2017-07-01 to 2018-06-01 causes an issue
+        # as there are duplicates of the .H.INTERNALHUB location. Deduping them here
+        df = df.drop_duplicates(subset=["BeginDate", "Location"], keep="first")
```
Kladar, line 382: @WillKoehrsen added this to de-dupe the verified duplicates and any more that may appear. Could make sense to move to `_handle_demand` but don't want to belabor this too much.

poetry.lock
```diff
@@ -4901,4 +4901,4 @@ type = ["pytest-mypy"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.9"
-content-hash = "b4ecb3559a0758888aaf816046088c21200bd61962f4b2c90141ab19c4e91ac3"
+content-hash = "4e2bf2a93e90d4cd39c6f42e010c1469c9ea9c4274fadc55cadfce55cbdbbec9"
```
WillKoehrsen, line 4904: Should this be updated if there haven't been any packages added/removed?

gridstatus/isone_api/isone_api.py
WillKoehrsen: Dropping duplicates is fine if we verified the duplicates occur in the source data (and aren't an artifact of our data).

PR #455, opened 2024-10-15
chore[ENG 776] ruff format over black

Kladar:
## Summary
Move to ruff formatter vs using black, for speed, simplicity, modernity, consistency.

### (*Optional*) Details
Feel free to add or suggest any other configurations for the ruff formatter. 

This shouldn't cause files that are touched to be reformatted drastically, but I've seen it happen before too. If it 1. formats touched files and 2. that gets annoying for reviewers, then we can run a format-only PR over the entire repo.


WillKoehrsen:
Do we also need to update the `Makefile` to remove references to `black`?

WillKoehrsen:
> we can run a format-only PR over the entire repo.

I'm fine with doing this as part of this PR.

Kladar:
> > we can run a format-only PR over the entire repo.
> 
> I'm fine with doing this as part of this PR.

Sounds good, done (only 7 files touched in the difference formatter, seems like minor changes)

Kladar:
> Do we also need to update the `Makefile` to remove references to `black`?

Yep good call, didn't realize those were in there. Updated.

pyproject.toml
WillKoehrsen, line 1: We should also be able to remove `black` from the dev dependencies in this file. 

https://github.com/gridstatus/gridstatus/pull/455/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L54

WillKoehrsen (approved):
Thanks, looks good!

pyproject.toml
Kladar: ah yeah, done.

poetry.lock
```diff
@@ -245,43 +245,6 @@ charset-normalizer = ["charset-normalizer"]
...
-platformdirs = ">=2"
-tokenize-rt = {version = ">=3.2.0", optional = true, markers = "extra == \"jupyter\""}
-tomli = {version = ">=1.1.0", markers = "python_full_version < \"3.11.0a7\""}
-typing-extensions = {version = ">=3.10.0.0", markers = "python_version < \"3.10\""}
-
-[package.extras]
-colorama = ["colorama (>=0.4.3)"]
-d = ["aiohttp (>=3.7.4)"]
-jupyter = ["ipython (>=7.8.0)", "tokenize-rt (>=3.2.0)"]
-uvloop = ["uvloop (>=0.15.2)"]
```
Kladar, line 283: hmm, why did these get removed on a ```poetry lock --no-update``` ?

PR #456, opened 2024-10-15
chore[ENG 776] Add mypy to work toward type safety

Kladar:
## Summary
This adds mypy as a CLI package and pre-commit hook, defines its settings in the pyproject.toml, and adds a coverage report command to the makefile that you can run yourself to dive into type annotations and coverage.

```
make mypy-coverage
```

CleanShot 2024-10-14 at 21 33 41@2x

Not starting from zero by any means and I don't think it'll be too hard to get full-ish coverage as we go.

### Details
Type-safety is critical for large python projects to be readable by newcomers, understood at a glance by people and programs, and avoids the largest class of runtime errors in python ahead of time with static checking. 

Since this is a large, existing repository with ~26k LOC (and maybe about 16k of which we want to check), there are some tips to make it go more smoothly we can employ if we are getting bogged down in type minutae. The goal is to _increase velocity_ through readability and catching errors before production, so if mypy slows us down too much it defeats the purpose. [Recommend this page from the mypy docs, it's very good.](https://mypy.readthedocs.io/en/stable/existing_code.html). We will dial it in as we go, and ideally are happy with our coverage by the end of the year. 

The default behavior of the mypy pre-commit hook is that it only runs on files that have been modified and thus have changes trying to be committed. For those files it then evaluates the entire file (and will block noncompliant files from being committed). If this happens while you are working, you have several options:
1. Update the file to be type-safe. Typically this is adding type hint annotation to functions (input args and output objects). 
2. Sometimes there is an import that doesn't have types or a edgecase that is hard to type, in which case you can ignore a single line or a block of code with a `# type: ignore`

```
import frobnicate # type: ignore
result = some_function()  # type: ignore
```
3. We can add a file or directory to the `exclude` list in the pyproject.toml. This is definitely useful if we are going to take a while to get around to adding types to certain sections of code. That said, sometimes its easier to just bite the bullet on a file and get it over the hump.

If there is a category of errors we want to ignore we can also add that to the `pyproject.toml` - I'll keep an eye out for this as I develop in here over the next week. We can also apply different rules to different directories, but that's overkill I think. 

I'd like to clear as much of the deck as possible in regards to this ticky tack python stuff before the start of the next quarter, such that we can hit the ground running on the work that moves the needle. A lot of this python level-up stuff is one-and-done, set-it-and-forget-it, but IMO impedes development at least somewhat until its knocked out.


Kladar:
Looks like the `types-all` [is deprecated](https://stackoverflow.com/questions/78833765/no-matching-distribution-found-for-types-pkg-resources). Not that I worry too much about external package type safety, but yeah, don't run `--install-types` or similar I guess.

.pre-commit-config.yaml
```diff
@@ -35,3 +35,7 @@ repos:
     hooks:
       - id: ruff
         args: ["--fix"]
+  - repo: https://github.com/pre-commit/mirrors-mypy
```
WillKoehrsen, line 38: I'm not sure about adding this to the pre-commit because pretty much every existing file is going to fail the check. Eventually, we'll get to the point where they pass, but that could be far in the future.

pyproject.toml
```diff
@@ -113,3 +114,14 @@ exclude_lines = [
     "raise NotImplementedError",
     "if __name__ == .__main__.:",
 ]
+
+[tool.mypy]
+ignore_missing_imports = true
```
WillKoehrsen, line 119: It seems like highlighting missing imports would be useful. Why ignore missing imports?

.pre-commit-config.yaml
Kladar: Fair. Middle ground is that I added each file to the `exclude` list in pre-commit so it shouldn't fail anything. Then we can slowly opt in files as they level up but not get blocked in the meantime. I've tested this and it correctly ignores the files I've changed, but let me know if you run into an issue.

WillKoehrsen: That works. I [also see you can run](https://github.com/pre-commit/pre-commit.com/blob/main/sections/advanced.md#temporarily-disabling-hooks) `SKIP=mypy git commit -m "message"` to skip a specific pre-commit (unlike `--no-verify` which skips all pre-commits).

pyproject.toml
WillKoehrsen: Makes sense, thanks for the explanation.

WillKoehrsen (approved):

PR #460, opened 2024-10-17
feat[ENG 849] ISONE Load Forecasts

Kladar:
## Summary
Adds new methods for getting the `reliabilityregionloadforecast/` and `hourlyloadforecast/` from ISO NE API. Also implements some `vcrpy` cassette fixtures for the existing and new methods. I like it a lot.

#### On Data
`/current` and `/day` endpoints return 8 locations, 192 rows for a “day”, 24 intervals per location. Current returns the whole latest day, which at time of writing (October 17th, 2:01PM ET) is October 19th, all intervals. It was published October 17th, ~1:30pm (half hour ago). There is more or less a new forecast every half hour, but it can vary, and they are not published at clean times (e.g. `2019-07-04 09:37:43-04:00`). Just whenever there is a new forecast.

the **`/reliabilityregionloadforecast/day/{day}`** returns the latest forecast data for that day’s intervals. Typically this means the forecast published the day before and the day of, collated together to give the shortest horizon forecast for each interval.

image

the **`/reliabilityregionloadforecast/day/{day}/all`** endpoint on the other hand returns all of the published forecasts for that day’s intervals. Typically this means that it goes back several days beforehand, returning a lot more data for a given day. Often it is 5x to 20x the number of intervals than the simple `day` endpoint, so while I default to give back `all` intervals, I give the option to return just the `latest` available as well. 

CleanShot 2024-10-17 at 14 39 41@2x

They are doing more frequent forecasting as time goes on, such that dates in 2024 return more data than dates in 2017. But rough back of the envelope is that there is a new forecast every half hourish, projected ~2 days out. 

CleanShot 2024-10-17 at 14 38 59@2x

I could stand to find more patterns here but I haven't looked very closely yet.

#### On Testing
I really like the VCR library. It worked pretty much out of the box. I think it is the future of this repository.

```
pytest -v gridstatus/tests/test_isone_api.py 
```

I did run into a similar issue as I did previously with `end_date` vs `end` in the `@support_date_range` decorator in that sometimes we are sneakily making multiple API calls with the single call of a `get_()` method. VCR assumes whatever comes back first is the thing it should use for the fixture, but with adding a little bit of `before_record` callback functionality that VCR supports, I was able to get it to concatenate the responses from the API into a single fixture (and clear the previous cassettes before re-writing them, if we are recreating them in a `record_mode=all`). This seemed to satisfy the tests and set us up nicely for switching between record modes/occasional fixture updating. 

You can add the environment variable `VCR_RECORD_MODE=all` to recreate fixtures, which is essentially running in integration mode (it defaults to "once" in the code).

We could probably stand to zip up the fixtures before expanding too much, but still need to figure out the s3 part which is probably a bigger priority.


gridstatus/isone_api/isone_api.py
```diff
@@ -65,7 +66,10 @@ def __init__(
         self.initial_delay = min(sleep_seconds, 60.0)
         self.max_retries = min(max(0, max_retries), 10)
 
-    # TODO(kladar) abstract this out to a base class since it is shared with ERCOT API logic
+    def parse_problematic_datetime(self, date_string: str) -> pd.Timestamp:
+        dt = datetime.strptime(date_string, "%Y-%m-%dT%H:%M:%S.%f%z")
```
WillKoehrsen, line 70: Can we just pass this in as `format` to `pd.to_datetime`? Or, is the format for some time strings different?

gridstatus/isone_api/isone_api.py
```diff
@@ -382,3 +339,193 @@ def get_dayahead_hourly_demand(
...
+        except AttributeError:
+            # NOTE(kladar) consistently the above logic fails for DST conversion days.
+            # This catches those and fixes it.
+            # Data coming out looks good, no missed intervals and no duplicates.
+            log.warning("Standard datetime conversion failed. Using custom parsing.")
+            df["Interval Start"] = df["BeginDate"].apply(
+                self.parse_problematic_datetime,
+            )
+
+        df = df.sort_values("Interval Start")
```
WillKoehrsen, line 373: Could we also add `Location` to the sort for consistency?

gridstatus/isone_api/isone_api.py
```diff
@@ -382,3 +339,193 @@ def get_dayahead_hourly_demand(
...
+        horizons: Literal["latest", "all"] = "all",
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Get the hourly load forecast data for specified locations and date range.
+
+        Args:
+            date (str): The start date for the data request. Use "latest" for most recent data.
+            end_date (str | None): The end date for the data request. Only used if date is not "latest".
+            horizons (Literal["latest", "all"]): The horizons for the data request. Options are "latest" or "all", defaults to "all".
```
WillKoehrsen, line 400: Could we add a little more explanation of what horizons does? Maybe just move the comment from the function body into the docstring.

gridstatus/isone_api/isone_api.py
Kladar: Yeah, I tried that (and a few other approaches that are similar) but the issue was with the way the pandas `dt` accessor handled dates in the object, all throwing this common `dt` error. So had to come up with a way to avoid the pandas accessor methods and fall back to `datetime` https://stackoverflow.com/questions/68734962/pandas-dataframe-column-can-only-use-dt-accessor-with-datetimelike-values

So my current hypothesis it that it's less of a string format issue and more of a duck-typing failure on the pandas end. Only happens on DST conversion days, and happens on every one of them. 

But also points out that the return type is wrong here, just updated that.

gridstatus/isone_api/isone_api.py
Kladar: Yep sounds good, I moved that to the docstring and expanded the explanation. Hopefully a bit more clear now.

gridstatus/isone_api/isone_api.py

gridstatus/isone_api/isone_api.py
WillKoehrsen: Got it, thanks for the explanation.

gridstatus/isone_api/isone_api.py
```diff
@@ -382,3 +339,200 @@ def get_dayahead_hourly_demand(
...
+            f"Processed demand data: {len(df)} entries from {df['Interval Start'].min()} to {df['Interval Start'].max()}",
+        )
+        return df[["Interval Start", "Interval End", "Location", "Location Id", "Load"]]
+
+    @support_date_range("DAY_START")
+    def get_hourly_load_forecast(
+        self,
+        date: str | pd.Timestamp = "latest",
+        end: str | pd.Timestamp | None = None,
+        horizons: Literal["latest", "all"] = "all",
```
WillKoehrsen, line 391: Given the explanation for `horizons`, I wonder if it makes sense to set the default to `"latest"`. My intuition is most people would only want the most recent forecast.

gridstatus/isone_api/isone_api.py
WillKoehrsen: Thanks.

gridstatus/isone_api/isone_api.py
```diff
@@ -65,7 +66,10 @@ def __init__(
         self.initial_delay = min(sleep_seconds, 60.0)
         self.max_retries = min(max(0, max_retries), 10)
 
-    # TODO(kladar) abstract this out to a base class since it is shared with ERCOT API logic
+    def parse_problematic_datetime(self, date_string: str) -> datetime:
+        dt = datetime.strptime(date_string, "%Y-%m-%dT%H:%M:%S.%f%z")
```
WillKoehrsen, line 70: When running `ISONEAPI().get_hourly_load_forecast('2024-03-11')`, I get the error `TypeError: strptime() argument 1 must be str, not Timestamp` which originates from this function.

gridstatus/tests/vcr_utils.py
```diff
@@ -0,0 +1,77 @@
+import json
+import os
+import shutil
+from urllib.parse import parse_qs, urlparse
+
+import vcr
```
WillKoehrsen, line 6: Do we need to add this library to the dev dependencies in `pyproject.toml`?

Kladar: ah yeah we do, still learning poetry a bit, will definitely add.

gridstatus/isone_api/isone_api.py
Kladar: Mmm, yeah I seem to get both errors actually (dt accessor and string parse), probably because I'm trying to do this operation on the whole column series, even though certain datetimes want one and others want another method. I think it is the actual DST fall back/spring forward intervals don't get recognized as datetimes, and then the other intervals are timestamps and thus fail the string parsing. One second.

CleanShot 2024-10-18 at 13 15 56@2x

Kladar: Just confirming that I get the .dt error when attempting to format with 

```
df["BeginDate"] = pd.to_datetime(df["BeginDate"], format="%Y-%m-%dT%H:%M:%S.%f%z")
df["CreationDate"] = pd.to_datetime(df["CreationDate"], format="%Y-%m-%dT%H:%M:%S.%f%z")
```

CleanShot 2024-10-18 at 13 24 02@2x

But a quick instance check on the problematic seems to fix it. I could just parse all the datetimes like that instead of trying the pd.to_datetime at all, but kind of want to avoid the parse to string and then call strptime wherever possible, only using it on the intervals/days that need it. 

I'm not sure that we are actually missing intervals in the database, but we must be because of the strptime failure, but my dupe SQL isn't picking it up.

gridstatus/tests/vcr_utils.py
Kladar: Looks like there is a conflict (though I definitely didn't run into any usage issues that I could tell), so I will add some dev dep logic and we'll see where we stand.

Makes me a little nervous that VCR hasn't upgraded to 2.0+ of urllib3. Might explore if there are other packages that do something similar.

Kladar: Seems like there is some motion, but not a ton. https://github.com/kevin1024/vcrpy/pull/777 And it is a much more active project than [betamax](https://github.com/kevin1024/vcrpy/pull/777), which is similar.

But ultimately it seems there is no resolution for existing `urllib3` and `vcrpy` dependencies. Very frustrating. 
CleanShot 2024-10-18 at 14 14 52@2x

gridstatus/isone_api/isone_api.py
Kladar: I actually had it that way and thought I wouldn't hide the fact that you can get all of them, given that `latest` is a subset of `all`. I see `all` as the critical functionality and `latest` is nice-to-have.

gridstatus/tests/vcr_utils.py
Kladar: Yeah, unfortunately with [no ability to specify conditional package installation](https://stackoverflow.com/questions/76102868/how-to-specify-2-diff%C3%A9rents-source-for-a-same-package-when-in-prod-or-in-dev-whe) in poetry and no resolution between urllib3 and vcr, I can't put vcr in the pyproject.toml and have it resolve.

Right now the options are
1. forgo vcr entirely (sad)
2. downgrade urllib3 to 1.25.0 (which might have other effects)
3. put the installation recommendation for vcr in the readme only, and open an issue on vcr to support 2.0+ urllib3. (path of least resistance, and again I didn't have any issues with urllib3==2.2.2 and vcrpy==6.0.2 when developing, so is likely the best option right now)
4. switch to a more flexible, modern package manager like `uv` [which allows conditional multi-source dependencies.](https://docs.astral.sh/uv/concepts/dependencies/#multiple-sources) (this would be my long term desire, but probably out of scope for this PR)

gridstatus/isone_api/isone_api.py
WillKoehrsen: Okay, I tried the new logic with both DST start and DST end and it seems to correctly parse the datetimes. Maybe we should add tests for these dates? (We can do in a follow-up PR).

gridstatus/isone_api/isone_api.py
```diff
@@ -382,3 +341,201 @@ def get_dayahead_hourly_demand(
...
+            date_columns = ["BeginDate", "CreationDate"]
+            df[date_columns] = df[date_columns].apply(pd.to_datetime)
+            df[date_columns] = df[date_columns].apply(
+                lambda x: x.dt.tz_convert(self.default_timezone),
+            )
+
+        except AttributeError:
+            # NOTE(kladar) consistently the above logic fails for DST conversion days.
+            # This catches those and fixes it.
+            # Data coming out looks good, no missed intervals and no duplicates.
```
WillKoehrsen, line 494: There should be a missed interval when DST starts right? Because the 2am local hour doesn't exist.

WillKoehrsen (approved):

gridstatus/isone_api/isone_api.py
WillKoehrsen: Anyway, this looks right

```python
ISONEAPI().get_hourly_load_forecast('2024-03-10')['Interval Start'].unique()[:4]

['2024-03-10 00:00:00-05:00', '2024-03-10 01:00:00-05:00', '2024-03-10 03:00:00-04:00', '2024-03-10 04:00:00-04:00']
Length: 4, dtype: datetime64[ns, US/Eastern]
```

gridstatus/isone_api/isone_api.py
```diff
@@ -382,3 +341,201 @@ def get_dayahead_hourly_demand(
...
+            f"Processed demand data: {len(df)} entries from {df['Interval Start'].min()} to {df['Interval Start'].max()}",
+        )
+        return df[["Interval Start", "Interval End", "Location", "Location Id", "Load"]]
+
+    @support_date_range("DAY_START")
+    def get_hourly_load_forecast(
+        self,
+        date: str | pd.Timestamp = "latest",
+        end: str | pd.Timestamp | None = None,
+        horizons: Literal["latest", "all"] = "all",
```
WillKoehrsen, line 393: Sorry, should have mentioned this earlier, but we call different versions of a forecast "vintages" so maybe we should use that here. I've only heard "horizon" used in the context of how far out a forecast goes ([example](https://help.anaplan.com/forecast-horizon-calculations--b7c8bc9e-ff41-4705-92fa-c94ec1cf7bdf)).

@kmax12 Do you have an opinion?

Kladar: Yep makes sense, I think of horizon and vintage as two sides of the same coin so happy to go with either where it makes sense.

`vintage` ~= `publish_time` aka, when was a forecast made? (like wine vintages. If a vintage is "2021", it was made in 2021)
`horizon` ~= `interval - publish_time` aka, how far out is a "forecast" (or collated series of forecasted intervals) 

In this case the `latest` actually does return a `horizon`, rather than a `vintage`. Typically, for relevant forecast performance statistics, we want the same horizon, basically "how good are we at predicting 30 minutes out? 1 day out?" It's collated across publish times or combining multiple vintages.

Comparing across vintages typically just shows how your forecast production system is improving over time. Which can be useful, but is technically not what is going on here, from my perspective.  

I talked about this in my case study briefly. 
CleanShot 2024-10-21 at 08 35 53@2x

Kladar: Neither here nor there, but I think we should add a horizon column to all forecast data, because it is an incredibly useful index and a bit of an expensive operation to create it sometimes.

PR #465, opened 2024-10-31
Update to all IESO public report links

BrianWeiHaoMa:
## Summary
Changed all http://reports.ieso.ca/public/ URL parts to https://reports-public.ieso.ca/public/ to reflect the IESO links update.

### Details
The IESO changed public report URLs to start with https://reports-public.ieso.ca/public/ since October 23. Source: https://www.ieso.ca/Sector-Participants/IESO-News/2024/10/New-url-for-Accessing-IESO-Public-Reports-Site.

PR #466, opened 2024-11-01
ERCOT DAM 60 Day Gen/Load Resources AS Offers Historical API Data

WillKoehrsen:
## Summary

- Adds ERCOT DAM 60 Day Gen Resources AS Offers to the historical ERCOT API
- Adds ERCOT DAM 60 Day Load Resources as Offers to the historical ERCOT API

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -3,6 +3,7 @@ import os
 import random
 import time
 from typing import Optional
+from zipfile import ZipFile
 
 import numpy as np
 import pandas as pd
@@ -88,6 +89,14 @@ SETTLEMENT_POINT_PRICE_REAL_TIME_15_MIN = "/np6-905-cd/spp_node_zone_hub"
 # Day ahead settlement point prices
 # https://data.ercot.com/data-product-archive/NP4-190-CD
 SPP_DAY_AHEAD_HOURLY = "/np4-190-cd/dam_stlmnt_pnt_prices"
+
+# DAM 60 Day Load Resource AS Offers
+# https://data.ercot.com/data-product-archive/NP3-966-ER
+DAM_60_DAY_LOAD_RESOURCES_AS_OFFERS_ENDPOINT = "/np3-966-er/60_dam_load_res_as_offers"
+
+# DAM 60 Day Gen Resource AS Offers
+# https://data.ercot.com/data-product-archive/NP3-966-ER
+DAM_60_DAY_GEN_RESOURCES_AS_OFFERS_ENDPOINT = "/np3-966-er/60_dam_gen_res_as_offers"
 
 
 class ErcotAPI:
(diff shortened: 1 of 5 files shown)


WillKoehrsen:
This now works back to 2011-07-01.

gridstatus/ercot_api/ercot_api.py
```diff
@@ -978,6 +990,62 @@ def get_spp_day_ahead_hourly(self, date, end=None, verbose=False):
 
         return data.sort_values(["Interval Start", "Location"]).reset_index(drop=True)
 
+    @support_date_range(frequency=None)
+    def get_dam_load_60_day_resources_as_offers(self, date, end=None, verbose=False):
```
WillKoehrsen, line 994: Should these two functions be combined into one so you can get both the load and gen data without having to download the same file twice?

gridstatus/tests/test_ercot_api.py
```diff
@@ -975,6 +976,32 @@ def test_get_spp_day_ahead_hourly_historical_date_range(self):
         assert df["Interval Start"].min() == self.local_start_of_day(start_date)
         assert df["Interval End"].max() == self.local_start_of_day(end_date)
 
+    """get_dam_gen_60_day_resources_as_offers"""
+
+    def test_get_dam_gen_60_day_resources_as_offers_historical_date_range(self):
```
Kladar, line 981: I'll work to understand the different elements here, don't quite know enough about the ERCOT API to remark intelligently on everything yet. Getting there!

First thing that comes up though, is do we want to start using vcrpy to develop fixtures for API data and unit tests going forward? Running vcrpy with the setting `VCR_RECORD_MODE=all` has the effect of querying the API fresh each time, doing what we do here with the integration style, but also allows for the quick unit-test style testing in development by switching to `VCR_RECORD_MODE=once`. I think you should be able to keep everything you have here but also import the short boilerplate at the top of the test file and add the VCR decorator to the test methods. Then that's it, should be good to go!

the `gridstatus/tests/test_isone_api.py` has the bits, I think it's ready for primetime but I have only used it the one time so far.

```
from gridstatus.tests.vcr_utils import RECORD_MODE, setup_vcr

api_vcr = setup_vcr(
    source="ercot_api",
    record_mode=RECORD_MODE,
)
```
```suggestion
    @api_vcr.use_cassette("test_get_dam_gen_60_day_resources_as_offers_historical_data_range.yaml")
    def test_get_dam_gen_60_day_resources_as_offers_historical_date_range(self):
```
___
Only other thing worth noting is that for date range methods, adding them to the `vcr_utils.py` list might be required, depending on if the method has to make multiple calls to the API (thus returning multiple files that have to be concatenated together to create the correct multi-day fixture). Open question on how to organize that which I call out in a `NOTE` in the code, but for now just have to toss the dataset ID in the list and it hopefully handles it for you. 
```
DATE_RANGE_METHODS = [
    "realtimehourlydemand",
    "dayaheadhourlydemand",
    "hourlyloadforecast",
    "reliabilityregionloadforecast",
]
```

WillKoehrsen: Will the VCR tests fail when the data returned from the API changes? Currently, I do not have confidence in pre-recorded tests to catch API changes compared to integration tests that retrieve the data from the API every time they are run. For example, I do not think pre-recorded unit tests would have caught the issue solved by https://github.com/gridstatus/gridstatus/pull/465.

Kladar: For `RECORD_MODE=all`, yes I believe they would fail, since it pulls new data from the API (rather than loading the fixture) and runs the assert statements against it. It is effectively the same as what we are doing now, since it ignores the existing fixture if present and freshly hits the API. Which we would want to run on things that test the actual, current API response. I definitely don't think we should do away with that functionality. It is vital in certain contexts.

What I'm hoping to achieve is additional structure to toggle on and off the "pull data from the external API" part of a test, since that is a part of a test that is slow and dynamic and not what I'm aiming to test with a unit test. A kind of test I want to run in integration (to catch things like in #465) but not at commit-level development.

WillKoehrsen: Is the main benefit of pre-recorded tests speed? When I'm doing commit-level development I only run affected tests. For this PR, that takes about 30 seconds and it's rare for the subset of tests affected by any one commit to take more than 60 seconds to run.

gridstatus/ercot_api/ercot_api.py
```diff
@@ -978,6 +990,81 @@ def get_spp_day_ahead_hourly(self, date, end=None, verbose=False):
 
         return data.sort_values(["Interval Start", "Location"]).reset_index(drop=True)
 
+    @support_date_range(frequency=None)
+    def get_dam_load_and_gen_60_day_resources_as_offers(
```
WillKoehrsen, line 994: I think it makes sense to have this one method return two dataframes in a dictionary because:

* Both files are in the same zip file so it's more efficient to get both files at once
* The datasets are closely related
* This is the same approach as in `Ercot().get_60_day_dam_disclosure`

gridstatus/tests/test_ercot_api.py
WillKoehrsen: We should continue this conversation but I'm going to merge this to close out this ticket.

Kladar: There are [many,](https://stackoverflow.com/questions/10752/what-is-the-difference-between-integration-and-unit-tests) many benefits to [separating integration and unit tests](https://stackoverflow.com/questions/5357601/whats-the-difference-between-unit-tests-and-integration-tests) and is a [long standing software engineering best practice](https://stackoverflow.com/questions/520064/what-are-the-differences-between-unit-tests-integration-tests-smoke-tests-and) for very good reasons. So much so that it was one of the first things I noticed when working on this repository as a major issue to effective, high-velocity development.

I would argue that
1. "affected tests" is a misnomer, as within an increasingly complex system you shouldn't have to track for which tests are or aren't affected, and it might be hard to tell at first or second glance. For unit tests, we should be able to run them all and not care.
2. 30 seconds is _exceedingly_ slow for unit tests. For this size of repository I would assume the entire unit testing suite (if done as true unit tests) should take less than a minute. Current that takes over an hour.
3. Running tests locally, manually, after every commit, is very slow and tedious. For myself, I want them running and giving me automatic feedback via CI on every commit so I can move faster and worry less about breaking things. We currently can't do this given the "only run certain logic tests at certain times" approach. 

I also ran into a great example of what can happen when unit-level testing relies on external dependencies today, when I was running ISONE fuel mix tests. Ran the test file:

image

Ran the test file again without touching anything but the two tests failing and got a third test failing!

image

Did I break something? Or did the dependency fail? 🤷🏼 

Touched the two original tests again, ran them, and got the two failures again. 

image

I assume it was the dependency, but I can't know and have no way to test if it will happen again. Any developer might find something like this frustrating. It tangles what I actually want to test (the code logic) from what I don't (the external dependency). I can only know for sure what I am testing by mocking the external dependency.  

I will set up a live conversation.

PR #467, opened 2024-11-02
ERCOT DAM LMP By Bus DST Fix

WillKoehrsen:
## Summary

- Fixes DST end issue in `ErcotAPI().get_lmp_by_bus_dam()` caused by `DSTFlag` column being a bool instead of a string
- Adds tests for both DST start and end for LMP by bus DAM

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -3171,7 +3171,11 @@ class Ercot(ISOBase):
                 # So, it's N during DST And Y during Standard Time
                 # Pandas wants True for DST and False for Standard Time
                 # during ambiguous times
-                ambiguous = doc["DSTFlag"] == "N"
+                # Some ERCOT datasets use a boolean, some use a string
+                if doc["DSTFlag"].dtype == bool:
+                    ambiguous = ~doc["DSTFlag"]
+                else:
+                    ambiguous = doc["DSTFlag"] == "N"
 
             try:
                 doc["Interval Start"] = doc["Interval Start"].dt.tz_localize(
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -3121,6 +3121,23 @@ class Ercot(ISOBase):
 
         ending_time_col_name = "HourEnding"
 
+        def ambiguous_based_on_dstflag(df):
+            # DST Flag is Y during the repeated hour so it's N during DST And Y during
+            # Standard Time
+            # For the ambiguous arg, Pandas wants True for DST and False for Standard
+            # Time during ambiguous times
+            # Some ERCOT datasets use a boolean, some use a string
+            if df["DSTFlag"].dtype == bool:
+                return ~df["DSTFlag"]
+            # Assume that if the DSTFlag column is a string, it's "Y" or "N"
+            else:
+                assert set(df["DSTFlag"].unique()).issubset({"Y", "N"})
+                return df["DSTFlag"] == "N"
+
+        ambiguous = dst_ambiguous_default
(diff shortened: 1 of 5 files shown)


gridstatus/tests/test_ercot_api.py
```diff
@@ -704,6 +704,37 @@ def test_get_lmp_by_bus_dam_historical_range(self):
         assert df["Interval Start"].min() == past_date.normalize()
         assert df["Interval End"].max() == past_end_date.normalize()
 
+    def test_get_lmp_by_bus_dam_dst_end(self):
```
WillKoehrsen, line 707: This test fails on `main` as expected.

gridstatus/ieso.py
WillKoehrsen, line 1: This file was failing the lint check.

https://github.com/gridstatus/gridstatus/actions/runs/11645125118/job/32427807367?pr=467

gridstatus/ercot.py
```diff
@@ -3171,7 +3171,11 @@ def parse_doc(
                 # So, it's N during DST And Y during Standard Time
                 # Pandas wants True for DST and False for Standard Time
                 # during ambiguous times
-                ambiguous = doc["DSTFlag"] == "N"
+                # Some ERCOT datasets use a boolean, some use a string
+                if doc["DSTFlag"].dtype == bool:
```
kmax12, line 3175: there is another usage of DSTFlag in this function above. https://github.com/gridstatus/gridstatus/blob/cb77142c119651029847fc5af4be905141c9c2c7/gridstatus/ercot.py#L3147

perhaps we can define this `ambiguous` earlier in the function and reuse it multiple times

gridstatus/ercot.py
```diff
@@ -3171,7 +3171,11 @@ def parse_doc(
...
+                    ambiguous = ~doc["DSTFlag"]
+                else:
+                    ambiguous = doc["DSTFlag"] == "N"
```
kmax12, line 3178: maybe we can add some defensive programing and assert that unique values are `Y` and `N` here?

i dont feel strongly about this, just trying to think what would have caught this exact error

kmax12 (changes requested):

gridstatus/ercot.py
WillKoehrsen: Good idea. Added a separate method with an assertion.

https://github.com/gridstatus/gridstatus/pull/467/commits/017ca892eab4094d714a89895e3823b54ed97e7d

gridstatus/ercot.py
WillKoehrsen: Added a check for 'Y' and 'N' exclusively. Any other changes before merging?

kmax12 (approved):

WillKoehrsen:
Something to look out for: the `DSTFlag` column used to be "N" or "Y" for the API but recently switched to True/False. The same holds for the `RepeatHourFlag` column which is not addressed in this PR. 

We need a follow-up PR to handle the inconsistency in `RepeatHourFlag` (which we rename to `RepeatedHourFlag` and use in downstream processing).

gridstatus/ercot_api/ercot_api.py
```diff
@@ -1068,7 +1068,7 @@ def get_historical_data(
                     break  # Exit the loop if the operation is successful
 
                 except Exception as e:
-                    if "Rate limit is exceeded" in response.decode("utf-8"):
```
WillKoehrsen, line 1071: I ran into an issue when running tests that the `response` couldn't be decoded. The exception message can be used to determine if we are being rate-limited.

PR #468, opened 2024-11-03
ERCOT Fix Real Time AS Monitor, Real Time System Conditions, and Forecasts Publish Dates DST Issue

WillKoehrsen:
## Summary

- Fixes `Ercot().get_as_monitor()` and `Ercot().get_real_time_system_conditions()` by specifying whether a time is ambiguous (True during DST)
- Fixes forecasts that use the `PublishDate` using the "xhr" that ERCOT adds to the second set of file names generated during the repeated hour at DST end. The fix is to ignore the provided offset and use one based on whether the file is the second set.

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2963,17 +2963,18 @@ class Ercot(ISOBase):
 
             friendly_name = doc["Document"]["FriendlyName"]
 
-            # ERCOT adds xhr to the file name during the repeated hour that occurs
-            # during DST end. However, ERCOT may get the timezone offset wrong.
-            # Therefore, we remove the ERCOT provided timezone offset then re-add the
-            # offset taking into account the repeated hour.
+            # ERCOT adds xhr to the second set of file names during the repeated hour
+            # for DST end. However, ERCOT may get the timezone offset wrong in the
+            # PublishDate. Therefore, we remove the ERCOT provided timezone offset then
+            # re-add the offset accounting for the repeated hour.
+            # https://lists.ercot.com/cgi-bin/wa?A3=1111&L=NOTICE_TRAINING&E=quoted-printable&P=4519&B=--_000_B117FDA9B7BC68479362C1197F77D8790950ADCPW0005ercotcom_&T=text%2Fhtml;%20charset=us-ascii&XSS=3&header=1
             publish_date = (
                 pd.Timestamp(doc["Document"]["PublishDate"])
                 .tz_localize(None)
                 .tz_localize(
                     self.default_timezone,
-                    # Pandas wants ambiguous to be True when DST is True during the
-                    # repeated hour (doesn't care otherwise). The "xhr" file occurs
+                    # Pandas wants ambiguous to be True when DST is True (Pandas only
+                    # uses ambiguous during the repeated hour) The "xhr" file occurs
                     # after the clock has been set back an hour so is not in DST.
                     ambiguous="xhr" not in friendly_name,
                 )
@@ -3160,10 +3161,12 @@ class Ercot(ISOBase):
         ending_time_col_name = "HourEnding"
 
         def ambiguous_based_on_dstflag(df):
-            # DST Flag is Y during the repeated hour so it's N during DST And Y during
-            # Standard Time
-            # For the ambiguous arg, Pandas wants True for DST and False for Standard
-            # Time during ambiguous times
+            # DSTFlag is Y during the repeated hour (after the clock has been set back)
+            # so it's False/N during DST And True/Y during Standard Time.
+            # For ambiguous, Pandas wants True for DST and False for Standard Time
+            # during repeated hours. Therefore, ambgiuous should be True when
+            # DSTFlag is False/N
+
             # Some ERCOT datasets use a boolean, some use a string
             if df["DSTFlag"].dtype == bool:
                 return ~df["DSTFlag"]
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2961,13 +2961,29 @@ class Ercot(ISOBase):
             except Exception:
                 friendly_name_timestamp = None
 
+            friendly_name = doc["Document"]["FriendlyName"]
+
+            # ERCOT adds xhr to the file name during the repeated hour that occurs
+            # during DST end. However, ERCOT may get the timezone offset wrong.
+            # Therefore, we remove the ERCOT provided timezone offset then re-add the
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -1851,10 +1851,20 @@ def _download_html_table(self, url, verbose=False):
             1
         ]  # Split the string on ': ' to get just the time part
 
+        now = pd.Timestamp.now(tz=self.default_timezone)
+
+        # Determine if during the repeated DST hour. Pandas wants ambiguous=True if
+        # it is during the repeated hour. US/Central is UTC-6 during standard time
+        # and UTC-5 during DST.
+        ambiguous = (now.utcoffset().total_seconds() / 3600) == -5.0
```
WillKoehrsen, line 1859: I think this works because Pandas only cares about ambiguous during the repeated hour.

gridstatus/ercot.py
```diff
@@ -2939,13 +2950,29 @@ def _get_documents(
             except Exception:
                 friendly_name_timestamp = None
 
+            friendly_name = doc["Document"]["FriendlyName"]
+
+            # ERCOT adds xhr to the file name during the repeated hour that occurs
+            # during DST end. However, ERCOT may get the timezone offset wrong.
```
WillKoehrsen: I believe the `PublishDate` offset from ERCOT is wrong during DST and should be ignored (as implemented in this PR). This would mean that our publish time during DST has been off by one hour for ERCOT.

gridstatus/ercot.py
kmax12: "Pandas wants ambiguous=True if the time is DST during the repeated hour. US/Central is UTC-6 during standard time"

doesn't this imply we want 

```
ambiguous = (now.utcoffset().total_seconds() / 3600) == -6.0
```

gridstatus/ercot.py
WillKoehrsen, line 2956: > ERCOT may get the timezone offset wrong

I'm confused here. ERCOT is showing a timezone offset of 6 hours for the publish time even when DST is in effect. This seems wrong. The offset should be 5 hours during DST and 6 hours outside of DST.

Screenshot 2024-11-03 at 03 40 55

gridstatus/ercot.py
WillKoehrsen: I don't think so because 6 hours is the offset outside of DST. We want ambiguous to be True when the timezone is in DST (offset 5 hours). 

From the Pandas docs:

> bool-ndarray where True signifies a DST time, False designates a non-DST time (note that this flag is only applicable for ambiguous times)

Example for US/Central

* 0130-0500 in DST, ambiguous=True
* 0130-0600 not in DST, ambiguous=False

gridstatus/ercot.py
WillKoehrsen: The Posted Time is showing as 2024-11-02 9:30 PM which should have an offset of 5 hours. We've just been parsing the `PublishDate` as given.

gridstatus/ercot.py
kmax12: i think you have those switched

image

kmax12: oh wait im mixing things up. CST is "standard time" not "savings time" 🤦

kmax12: i think your comment is wrong/ambiguous you say: "Pandas wants ambiguous=True if it is during the repeated hour"

the non-DST time is when the hour repeats itself, so i would think a better might just be to copy pandas documentation: "True signifies a DST time, False designates a non-DST time"

WillKoehrsen: The comment currently reads:

> Pandas wants ambiguous=True if the time is DST during the repeated hour.

kmax12: looks good. sorry for my confusion here.

gridstatus/ercot.py
kmax12: > This would mean that our publish time during DST has been off by one hour for ERCOT.

does this apply to our history or just on DST transition days?

WillKoehrsen: I'm not sure. Looking at past reports shows the same incorrect `PublishDate` offset but ERCOT may be applying the non-DST offset to all timestamps it displays. now that DST is over. In other words, if we looked at those publish times yesterday, it's possible we would have seen the correct offset.

Screenshot 2024-11-03 at 06 33 48

kmax12: confusing. Let me look at this a bit more closely before we merge

kmax12: in reviewing the [list documents from the report](https://www.ercot.com/misapp/servlets/IceDocListJsonWS?reportTypeId=12311) in your screenshot and then also the [Real Time Settlement Point](https://www.ercot.com/misapp/servlets/IceDocListJsonWS?reportTypeId=12301&_=1730641843741) list, I reached the same conclusion with you and agree with this fix

Just to confirm my logic:
1. Ercot is using the UTC offset at the time document list is returned and adds it to the all publish times 
2. However, ERCOT doesn't adjust the time of timestamp, it just add the offset. 
3. This leads to situations where the publish time indicates a `-6` offset, but it should be interpreted as `-5`
4. Therefore, we should ignore the ERCOT offset and use `xhr` in the file name to differentiate when the time stamp is ambiguous 

My only concern here is does this change work across the datasets we've implemented that call `_get_documents`

gridstatus/ercot.py
WillKoehrsen: No worries, ERCOT makes this really confusing.

gridstatus/ercot.py
WillKoehrsen: Your logic matches my conclusion. We will have to wait until DST starts again though to confirm.

> My only concern here is does this change work across the datasets we've implemented that call _get_documents

ERCOT says that "xhr" in a file name represents the second file published in a repeated hour.

> ERCOT has appended an xhr indicator in the file naming convention to create file uniqueness across the repeated hour files generated. The first file generated will post with the expected naming convention. The second file generated will appear with the _xhr designation.

https://lists.ercot.com/cgi-bin/wa?A3=1111&L=NOTICE_TRAINING&E=quoted-printable&P=4519&B=--_000_B117FDA9B7BC68479362C1197F77D8790950ADCPW0005ercotcom_&T=text%2Fhtml;%20charset=us-ascii&XSS=3&header=1

kmax12 (approved):
Lgtm

PR #469, opened 2024-11-03
DST Fix for energy_storage_resources

kmax12:
The problematic intervals
 
image (91)

image (92)

and the missing interval

image

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -335,10 +335,6 @@ class Ercot(ISOBase):
         )
 
         df = df.sort_values("Time").reset_index(drop=True)
-        import pdb
-
-        pdb.set_trace()
-
         return df
 
     def get_fuel_mix(self, date, verbose=False):
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -303,6 +303,18 @@ class Ercot(ISOBase):
 
         df = pd.DataFrame(data["previousDay"]["data"] + data["currentDay"]["data"])
 
+        # todo(kanter): fix this for future DST dates
+        if "2024-11-03 02:00:00-0600" in df["timestamp"].values:
+            # ERCOT publishes two intervals with 2am timestamp
+            # during CDT to CST transition
+            # but skips the repeated 1am timestamp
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -323,6 +335,9 @@ def get_energy_storage_resources(self, date="latest", verbose=False):
         )
 
         df = df.sort_values("Time").reset_index(drop=True)
+        import pdb
```
WillKoehrsen, line 338: Remove before merging.

gridstatus/ercot.py
```diff
@@ -303,6 +303,18 @@ def get_energy_storage_resources(self, date="latest", verbose=False):
...
 
+        # todo(kanter): fix this for future DST dates
+        if "2024-11-03 02:00:00-0600" in df["timestamp"].values:
+            # ERCOT publishes two intervals with 2am timestamp
+            # during CDT to CST transition
+            # but skips the repeated 1am timestamp
+            # let's manually fix this before further timestamp parsing
+            df.loc[
+                (df["timestamp"] == "2024-11-03 02:00:00-0600")
+                & (df["dstFlag"] == "N"),
```
WillKoehrsen, line 314: This assumes the `dstFlag == "Y"` for the repeated hour only?

kmax12: not 100% I follow the question.  but there two 2am intervals with -6 offset but a missing 1am interval with -6 offset, so i am manually shifting one based on `dstFlag`

WillKoehrsen: How do you know to use "Y" or "N" to identify the interval to shift?

WillKoehrsen: If `dstFlag == "Y"` for the repeated hour and ERCOT labels rows by hour end then this logic makes sense.

WillKoehrsen (approved):

gridstatus/ercot.py
kmax12: yeah, it is dstFlag == "Y" when things repeat. 

this data is 5 minute though. every other interval gets labeled correctly. this is some weird error on ERCOT's side  i think

PR #470, opened 2024-11-04
SPP 5 Minute Dataset DST End Fixes

WillKoehrsen:
## Summary

- Fixes issues with SPP 5 Minute Datasets (LMP, Operating Reserves) that were not handling file names during the repeated hour
- NOTE: SPP is missing 1 hour of data for each of these datasets, verified in the source data

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -1,11 +1,18 @@
 from typing import BinaryIO
+from urllib.error import HTTPError
 
 import pandas as pd
 import requests
 import tqdm
 
 from gridstatus import utils
-from gridstatus.base import InterconnectionQueueStatus, ISOBase, Markets, NotSupported
+from gridstatus.base import (
+    InterconnectionQueueStatus,
+    ISOBase,
+    Markets,
+    NoDataFoundException,
+    NotSupported,
+)
 from gridstatus.decorators import FiveMinOffset, support_date_range
 from gridstatus.gs_logging import log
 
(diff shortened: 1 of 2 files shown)


gridstatus/spp.py
```diff
@@ -1206,15 +1219,36 @@ def _format_5_min_url(
...
         folder_day = start.strftime("%d")
 
         interval_str = "/By_Interval" if include_interval else ""
 
-        return f"{FILE_BROWSER_DOWNLOAD_URL}/{endpoint}?path=/{folder_year}/{folder_month}{interval_str}/{folder_day}/{file_prefix}-{end.strftime('%Y%m%d%H%M')}.csv"  # noqa
+        url = f"{FILE_BROWSER_DOWNLOAD_URL}/{endpoint}?path=/{folder_year}/{folder_month}{interval_str}/{folder_day}/{file_prefix}-{end.strftime('%Y%m%d%H%M')}.csv"  # noqa
+
+        status_code = requests.head(url).status_code
+
+        if status_code == 200:
```
kmax12, line 1238: does this actually work as intended? here's the case im wondering about

the user requests `Nov 3, 2024 01:30am CST` . I think this function will first try to look for the file at `1:30am CDT`. if that file isn't found it will look for it during standard time version by appending the `d`. 

if the CDT file is found, it seems like this function will return the wrong url

i know this past year it looked like the CST files were missing, but i dont think that is guaranteed to be the case.

gridstatus/spp.py
```diff
@@ -871,7 +878,13 @@ def _get_real_time_5_min_data(
 
         log(f"Getting data for {date} from {url}", verbose=verbose)
 
-        df = pd.read_csv(url)
+        try:
```
kmax12, line 881: why do we need this try except if we are calling `_format_5_min_url` above? isn't that supposed to handle this?

WillKoehrsen: Yeah, this isn't needed. It's leftover from a previous approach.

gridstatus/spp.py
WillKoehrsen: You were correct. I've committed a fix that will work for both the CDT and CST files and checked it on last year when both sets of files were published.

https://github.com/gridstatus/gridstatus/pull/470/commits/f6ea436cafe063c47190792da76b71ec8f0587ba

kmax12 (approved):

WillKoehrsen:
I've verified this works to backfill the data we missed 2024-11-03.

PR #471, opened 2024-11-04
fix[ENG 896] ISO NE fuel mix 2024 fallback transition time fix

Kladar:
## Summary
Handles the issue with ambiguous times during the "repeated" (really EDT and EST) 1am hour which is causing an issue in pulling the `latest` and `date=2024-11-03` data. 

### Details
Raw data, two 1am times that are ambiguous

CleanShot 2024-11-03 at 18 16 46@2x

CleanShot 2024-11-03 at 18 18 37@2x

Processed Data
CleanShot 2024-11-03 at 18 20 25@2x

CleanShot 2024-11-03 at 18 25 31@2x

Passes on the test given the ability to test it relies on being on the actual day, at the moment.


gridstatus/tests/test_isone.py
```diff
@@ -11,6 +11,8 @@
 # toggle for debugging
 VERBOSE = False
 
+# NOTE(kladar): Enumming the DST boundaries for all past and future DST transitions
```
kmax12, line 14: i agree. a constant with this info would be very helpful

gridstatus/isone.py
```diff
@@ -127,12 +149,26 @@ def get_fuel_mix(self, date, end=None, verbose=False):
...
+            df["Date"].dt.date == pd.Timestamp("2024-11-03").date()
+        ) & (df["Date"].dt.hour == 1)
+        transition_idx = None
+
+        if is_dst_transition.any():
+            logger.debug(
+                "Found DST transition. Looking for fallback transition time...",
+            )
+            time_diffs = df["Date"].diff()
+            negative_diffs = time_diffs[time_diffs < pd.Timedelta(minutes=-30)]
```
kmax12, line 163: i see the issue. in theory this approach of looking at the order is what pandas `ambiguous=infer` does. any idea why it stopped working? 

like if you try to run the original code on DST transition during november last year, it infers correctly

image

gridstatus/isone.py
```diff
@@ -106,8 +106,30 @@ def get_status(self, date, verbose=False):
...
+    def _assign_dst_aware_time(
+        self,
+        series: pd.Series,
+        transition_idx: int | None,
+    ) -> pd.Series:
+        if transition_idx is None:
+            return series.dt.tz_localize("US/Eastern", ambiguous="infer")
+
+        dates = pd.Series(index=series.index, dtype="datetime64[ns, US/Eastern]")
+        dates[series.index < transition_idx] = series[
```
kmax12, line 118: you can [pass a boolean mask](https://pandas.pydata.org/docs/reference/api/pandas.Series.tz_localize.html#:~:text=bool%2Dndarray%20where%20True%20signifies%20a%20DST%20time%2C%20False%20designates%20a%20non%2DDST%20time%20(note%20that%20this%20flag%20is%20only%20applicable%20for%20ambiguous%20times)) to `ambiguous`

would this be the same thing? 

`data.dt.tz_localize("US/Eastern", ambiguous=series.index < transition_idx)`

kmax12 (commented):
some comments, but wondering if you can ignore all of them. instead, maybes we should just migrate this to using the ISONE api instead? it looks like it has the timezone offset info in its response

https://webservices.iso-ne.com/docs/v1.1/rest.genfuelmix.html

i queried the endpoint above and compared one interval myself and it matches your screenshot
image

gridstatus/isone.py
Kladar: Hmm, interesting, yeah not sure why it stopped working then. Especially since it gives an `AmbiguousTimeError` which one would think a "here's what you do when it's ambiguous?" kwarg would cover. 🤷🏼

Kladar:
> some comments, but wondering if you can ignore all of them. instead, maybe we should just migrate this to using the ISONE api instead? it looks like it has the timezone offset info in its response
> 
> https://webservices.iso-ne.com/docs/v1.1/rest.genfuelmix.html
> 
> i queried the endpoint above and compared one interval myself and it matches your screenshot image

I think yeah we should definitely migrate to the API methods when possible, but I remember there were some gotchas with `load` and `load forecast` when we thought about doing that as part of the previous batches of ISO NE API integration datasets, and ultimately pointing new methods at existing tables had some other elements to think through.

I think we should have some sort of agreed upon ISO NE API swap strategy, like 
1. register "new" dataset (AKA point new method at new table)
2. backfill
3. add to cron
4. drop old table
5. rename new table to old table name, carry on
 
Or what have you. Like in theory yes we can, and you might be able to archmage your way through it, but with my lack of experience in this system IMO it's too easy to bork it and cause a bigger headache (and potentially lose historical data, depending on the dataset). But I tend to avoid the  
indy-swap

gridstatus/isone.py
```diff
@@ -106,8 +106,30 @@ def get_status(self, date, verbose=False):
...
+            series.index < transition_idx
+        ].dt.tz_localize("US/Eastern", ambiguous=True)
```
WillKoehrsen, line 120: We should use `self.default_timezone` instead of the string `US/Eastern`.

gridstatus/isone.py
WillKoehrsen: Load, LMP, solar and wind forecasts use the same ambiguous="infer" argument and they correctly parsed the transition yesterday. I wonder what makes the fuel mix different?

WillKoehrsen: Ohh, the problem is the "Other" fuel category doesn't have any repeats. Here's the data

Screenshot 2024-11-04 at 11 23 01

Compare to "Hydro" (or any other Fuel Category) that has repeats.

Screenshot 2024-11-04 at 11 24 06

Kladar: I think we can have the goal of moving all applicable datasets to the API by the next DST transition. [Here's a ticket to capture that](https://www.notion.so/gridstatus/Swap-ISO-New-England-methods-for-API-134e835f42aa802a9443e0ca2c5addbf?pvs=4).

WillKoehrsen: `ambiguous="infer"` works on all fuel categories except for "Other". I think we can simplify the fix:

```python
groups = []

# groupby Fuel Category to make it possible to infer DST changes
for _category, group in df.groupby("Fuel Category"):
    try:
        group["Date"] = group["Date"].dt.tz_localize(
            self.default_timezone,
            ambiguous="infer",
        )
    # Some groups ("Other") may not have duplicate times and thus Pandas cannot
    # infer the correct timezone. In these cases, assume the timestamps
    # are the first occurrence
    except pytz.AmbiguousTimeError:
        group["Date"] = group["Date"].dt.tz_localize(
            self.default_timezone, ambiguous=True
        )

    groups.append(group)

df = pd.concat(groups)
```

Kladar: swapping for the API

gridstatus/isone_api/isone_api.py
```diff
@@ -210,7 +210,8 @@ def get_fuel_mix(
         mix_df.columns.name = None
 
         mix_df = mix_df.rename(columns={"BeginDate": "Time"})
-        mix_df["Time"] = pd.to_datetime(mix_df["Time"]).dt.strftime(
+        mix_df["Time"] = mix_df["Time"].apply(self.parse_problematic_datetime)
+        mix_df["Time"] = mix_df["Time"].dt.strftime(
```
kmax12: not sure i understand the call to `strftime`. Don't we normally return the pandas timestamp object here?

kmax12: sorry, i see you are still working. i thought i got a notification you rerequested review.

Kladar: yeah was trying some things with gs-etl which was running into problems with the datetimeindex of the interpolation. Working now.

Kladar:
@kmax12 @WillKoehrsen this now leaves the existing method alone, adds the ISONE API method.

kmax12 (approved):
I didn't run, but code looks good me

gridstatus/isone_api/isone_api.py
```diff
@@ -175,6 +176,46 @@ def get_locations_all(self) -> pd.DataFrame:
...
+        mix_df = df.pivot_table(
+            index="BeginDate",
+            columns="FuelCategory",
+            values="GenMw",
+            aggfunc="first",
+        ).reset_index()
+        mix_df.columns.name = None
+
+        mix_df = mix_df.rename(columns={"BeginDate": "Time"})
+        mix_df["Time"] = mix_df["Time"].apply(self.parse_problematic_datetime)
```
WillKoehrsen, line 213: We should add `Interval Start` and `Interval End` to this dataset.

WillKoehrsen: Hmm, `ISONE().get_fuel_mix()` doesn't have an "Interval Start" and "Interval End"...

WillKoehrsen:
The API returns different columns on different days.

```python
ISONEAPI().get_fuel_mix('2024-11-01')

                         Time   Coal    Hydro  Landfill Gas  Natural Gas    Nuclear    Oil   Other   Refuse  Solar       Wind     Wood
0   2024-11-01 00:04:22-04:00 0.0000 217.0000        4.0000   6_489.0000 1_224.0000 0.0000  0.0000 287.0000 0.0000   982.0000 158.0000
1   2024-11-01 00:16:21-04:00 0.0000 217.0000        4.0000   6_274.0000 1_222.0000 0.0000 22.0000 303.0000 0.0000   998.0000 157.0000
2   2024-11-01 00:25:12-04:00 0.0000 213.0000        4.0000   6_210.0000 1_223.0000 0.0000  1.0000 298.0000 0.0000 1_012.0000 159.0000
3   2024-11-01 00:36:34-04:00 0.0000 214.0000        4.0000   6_252.0000 1_226.0000 0.0000  1.0000 303.0000 0.0000 1_019.0000 159.0000
4   2024-11-01 00:45:54-04:00 0.0000 213.0000        4.0000   6_285.0000 1_225.0000 0.0000 13.0000 307.0000 0.0000 1_026.0000 159.0000

ISONEAPI().get_fuel_mix('2024-11-04')

                         Time    Hydro  Landfill Gas  Natural Gas    Nuclear     Oil   Other   Refuse    Solar     Wind     Wood
0   2024-11-04 00:02:39-05:00 251.0000        4.0000   7_939.0000 1_231.0000  0.0000 11.0000 321.0000   9.0000 118.0000 171.0000
1   2024-11-04 00:04:18-05:00 266.0000        4.0000   7_941.0000 1_231.0000  0.0000 13.0000 321.0000   9.0000 116.0000 171.0000
2   2024-11-04 00:07:51-05:00 285.0000        4.0000   7_803.0000 1_231.0000  0.0000 57.0000 318.0000   9.0000 118.0000 176.0000
3   2024-11-04 00:10:50-05:00 300.0000        4.0000   7_832.0000 1_231.0000  0.0000  0.0000 324.0000   9.0000 121.0000 177.0000
4   2024-11-04 00:12:39-05:00 271.0000        4.0000   7_875.0000 1_231.0000  0.0000  0.0000 329.0000   9.0000 121.0000 174.0000
```

Should we add the missing columns as NAs when they are not present? I think returning a consistent set of columns is important.

gridstatus/isone_api/isone_api.py
WillKoehrsen: IMO we should still add these columns for consistency across datasets.

Kladar:
> The API returns different columns on different days.
> 
> ```python
> ISONEAPI().get_fuel_mix('2024-11-01')
> 
>                          Time   Coal    Hydro  Landfill Gas  Natural Gas    Nuclear    Oil   Other   Refuse  Solar       Wind     Wood
> 0   2024-11-01 00:04:22-04:00 0.0000 217.0000        4.0000   6_489.0000 1_224.0000 0.0000  0.0000 287.0000 0.0000   982.0000 158.0000
> 1   2024-11-01 00:16:21-04:00 0.0000 217.0000        4.0000   6_274.0000 1_222.0000 0.0000 22.0000 303.0000 0.0000   998.0000 157.0000
> 2   2024-11-01 00:25:12-04:00 0.0000 213.0000        4.0000   6_210.0000 1_223.0000 0.0000  1.0000 298.0000 0.0000 1_012.0000 159.0000
> 3   2024-11-01 00:36:34-04:00 0.0000 214.0000        4.0000   6_252.0000 1_226.0000 0.0000  1.0000 303.0000 0.0000 1_019.0000 159.0000
> 4   2024-11-01 00:45:54-04:00 0.0000 213.0000        4.0000   6_285.0000 1_225.0000 0.0000 13.0000 307.0000 0.0000 1_026.0000 159.0000
> 
> ISONEAPI().get_fuel_mix('2024-11-04')
> 
>                          Time    Hydro  Landfill Gas  Natural Gas    Nuclear     Oil   Other   Refuse    Solar     Wind     Wood
> 0   2024-11-04 00:02:39-05:00 251.0000        4.0000   7_939.0000 1_231.0000  0.0000 11.0000 321.0000   9.0000 118.0000 171.0000
> 1   2024-11-04 00:04:18-05:00 266.0000        4.0000   7_941.0000 1_231.0000  0.0000 13.0000 321.0000   9.0000 116.0000 171.0000
> 2   2024-11-04 00:07:51-05:00 285.0000        4.0000   7_803.0000 1_231.0000  0.0000 57.0000 318.0000   9.0000 118.0000 176.0000
> 3   2024-11-04 00:10:50-05:00 300.0000        4.0000   7_832.0000 1_231.0000  0.0000  0.0000 324.0000   9.0000 121.0000 177.0000
> 4   2024-11-04 00:12:39-05:00 271.0000        4.0000   7_875.0000 1_231.0000  0.0000  0.0000 329.0000   9.0000 121.0000 174.0000
> ```
> 
> Should we add the missing columns as NAs when they are not present? I think returning a consistent set of columns is important.

@will Yes I also noticed this, but I believe it is because this is currently handled downstream. Current behavior is consistent with the old method.

gridstatus/isone_api/isone_api.py
Kladar: Correct, and it also doesn't return consistent columns I don't believe.

WillKoehrsen (approved):
Thanks!

will:
> @will Yes I also noticed this, 

👋

PR #472, opened 2024-11-05
feat[ENG 883] PJM Constraints Datasets

Kladar:
## Summary
Adds methods to get the PJM `Real Time Marginal Value`, `Day Ahead Marginal Value`, and `Day Ahead Transmission Constraints` data from the PJM API. 

### Details
Tests passing locally for PJM (though it seemed to have hung for a while and had to kill it on the last test). 
CleanShot 2024-11-04 at 20 08 01@2x

Also adds some typehints to the `_get_pjm_json()` method.  

Extends the vcrpy functionality to `test_pjm.py` and makes the vcr slightly more flexible. Adds the VCR created fixtures.


gridstatus/tests/test_pjm.py
```diff
@@ -1815,3 +1821,90 @@ def test_get_forecasted_generation_outages_historical_range(self):
         assert (
             df["Interval End"] == df["Interval Start"] + pd.DateOffset(days=1)
         ).all()
+
+    @pytest.mark.parametrize(
+        "date,end",
+        [
+            ("2024-09-02", "2024-09-04"),
+        ],
+    )
```
WillKoehrsen, line 1830: What's the benefit of using `parameterize` if we only use one set of arguments?

Kladar: Good question. I could see a lot of benefit from continually add more as we find specific days with issues, like DST transitions and things. I can add a few more now if we want. I think I'm still figuring out the parameterization with the cassettes a bit, since 

1. I think it currently writes to a static (rather than parameterized) fixture path and 
2. the number and size of cassettes could grow quickly with each method having multiple fixtures and might make sense to move the fixtures elsewhere before adding a bunch of them.

Plus it might make sense to have a master list of these dates rather than add them to each test method, so still some open questions I think. Good topic for this afternoon testing meeting.

WillKoehrsen: Makes sense. I think it would be useful to have the dates in the fixture name. Can't we use the context manager within the test body to dynamically get the fixture name? (`with vcr.use_casette`)?

WillKoehrsen (approved):

gridstatus/tests/test_pjm.py
Kladar: oh nice, excellent, yes that works really well. 

I used the context manager, added last year's DST fall back, and pulled that parameter to the top of the test class. Lmk what you think.

WillKoehrsen: Nice, this should work well for testing multiple dates in the same test method.

PR #473, opened 2024-11-06
Fix ERCOT AS Reports for DST

WillKoehrsen:
## Summary

- Fixes `ErcotAPI().get_as_reports()` for DST end


gridstatus/ercot.py
```diff
@@ -3240,6 +3241,21 @@ def ambiguous_based_on_dstflag(df):
                     self.default_timezone,
                     ambiguous=ambiguous,
                 ) - pd.Timedelta(hours=1)
+            except pytz.AmbiguousTimeError:
+                # Sometimes ERCOT handles DST end by putting 25 hours in HourEnding
```
WillKoehrsen, line 3245: The raw data has 25 HourEnding values. Example:

```
"Delivery Date","Hour Ending","Total Self-Arranged AS - RRSPFR"
"11/03/2024","1","0"
"11/03/2024","2","0"
"11/03/2024","3","0"
"11/03/2024","4","0"
"11/03/2024","5","0"
"11/03/2024","6","0"
"11/03/2024","7","0"
"11/03/2024","8","0"
"11/03/2024","9","0"
"11/03/2024","10","0"
"11/03/2024","11","0"
"11/03/2024","12","0"
"11/03/2024","13","0"
"11/03/2024","14","0"
"11/03/2024","15","0"
"11/03/2024","16","0"
"11/03/2024","17","0"
"11/03/2024","18","0"
"11/03/2024","19","0"
"11/03/2024","20","0"
"11/03/2024","21","0"
"11/03/2024","22","0"
"11/03/2024","23","0"
"11/03/2024","24","0"
"11/03/2024","25","-10"
```

The solution assumes the `HourEnding` values are in order (in contrast to say HourEnding 25 being the repeated hour and being put on the end of the data).

gridstatus/ercot.py
```diff
@@ -3240,6 +3241,21 @@ def ambiguous_based_on_dstflag(df):
...
+                # which makes IntervalStart where HourEnding >= 3 an hour later than
+                # they should be. We correct this by subtracting an hour.
+                doc.loc[doc["HourEnding"] >= 3, "Interval Start"] = doc.loc[
```
kmax12, line 3248: to be safe, i think we probably want to also assert or wrap this logic with a conditional that confirms that there is a HourEnding that equals 25 in the data

kmax12 (approved):
just one comment. otherwise looks good

gridstatus/tests/fixtures/ercot_api/vcr_cassettes/test_get_as_reports_dst_end_2024_11_03.yaml
WillKoehrsen, line 1: This VCR uses an absolute date for DST end which is fine in this case because the `ErcotAPI` has no retention policy (all files are retained for all time).

gridstatus/tests/fixtures/ercot_api/vcr_cassettes/test_get_as_reports_full_columns_21_days_ago.yaml
WillKoehrsen, line 1: This VCR uses a relative date because the specific date doesn't matter. We should be able to re-record this cassette and the tests should still pass.

gridstatus/ercot.py
WillKoehrsen: Good idea, added the assertion.

PR #474, opened 2024-11-06
Fix for ERCOT DAM AS Offers with Repeated Offers

WillKoehrsen:
## Summary

- Addresses an issue where one resource in ERCOT DAM AS Offers has repeated offers for the same service at the same interval
- This issue previously caused the data retrieval to fail. The fix choses one of the rows to keep in a consistent manner.

### DIFF
--- a/gridstatus/ercot_60d_utils.py
+++ b/gridstatus/ercot_60d_utils.py
@@ -353,9 +353,18 @@ def process_as_offer_curves(df):
             )
 
             if len(subset) > 1:
-                raise ValueError(
-                    f"More than one row found for {column_list} for {resource_name}",
-                )
+                # We've identified an issue with this specific resource name where
+                # there are sometimes multiple offers for the same service at the same
+                # interval. In theory this should never happen. The QUANTITY MW are
+                # only different by 0.1, so we just take the row with the lowest
+                # quantity. This is a temporary fix until we can figure out why this
+                # is happening.
+                if resource_name == "CANYONRO_LD1":
+                    subset = subset.sort_values("QUANTITY MW1").iloc[0]
+                else:
+                    raise ValueError(
+                        f"More than one row found for {column_list} for {resource_name}",  # noqa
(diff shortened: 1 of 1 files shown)


gridstatus/ercot_60d_utils.py
```diff
@@ -353,9 +353,18 @@ def process_as_offer_curves(df):
...
-                    f"More than one row found for {column_list} for {resource_name}",
-                )
+                # We've identified an issue with this specific resource name where
+                # there are sometimes multiple offers for the same service at the same
+                # interval. In theory this should never happen. The QUANTITY MW are
+                # only different by 0.1, so we just take the row with the lowest
+                # quantity. This is a temporary fix until we can figure out why this
+                # is happening.
+                if resource_name == "CANYONRO_LD1":
+                    subset = subset.sort_values("QUANTITY MW1").iloc[0]
```
WillKoehrsen, line 363: Choosing the row with the lowest `QUANTITY MW1` is arbitrary. When cross-referencing with another dataset (SCED Load resource 60 day) there wasn't a reliable way to determine which row was "correct". The values in `QUANTITY MW1` are only different by 0.1.

kmax12 (approved):

gridstatus/tests/test_ercot_api.py
```diff
@@ -1072,6 +1072,37 @@ def test_get_dam_load_and_gen_60_day_resources_as_offers(self):
             assert df["Interval Start"].min() == start_date
             assert df["Interval End"].max() == end_date
 
+            assert df.groupby(["Interval Start", "Resource Name"]).size().max() == 1
+
+    def test_get_dam_load_and_gen_60_day_resources_as_offers_repeated_offers(self):
```
WillKoehrsen, line 1077: I did not use VCR for this test because the casette `yaml` file was 10.7 MB on disk and, once that file is added to the repo, it can never be removed from the repo history.

PR #475, opened 2024-11-09
Update EIA Fuel Mix for New Power Storage Columns

WillKoehrsen:
## Summary

- Updates `EIA()` fuel mix retrieval to handle new battery storage columns


gridstatus/eia.py
```diff
@@ -741,14 +755,49 @@ def _handle_fuel_type_data(df):
...
+        if col not in df.columns:
+            # This has to be np.nan not pd.NA because we are converting to float
+            df[col] = np.nan
+
     fuel_mix_cols = df.columns[4:]
 
-    # nans after pivot because not
-    # all respondents have all fuel types
+    # nans after pivot because not all respondents have all fuel types
+    # Why are we filling these with 0? A value of 0 means something different than NaN
```
WillKoehrsen, line 793: @kmax12 Why are the na values being filled with 0? I would think they should be kept as na to avoid creating data values.

kmax12: i think my logic was that if the fuel didn't get reported, it means it wasn't generated, hence it is okay to say zero. 

I'm okay if you want to change it to null.

if you do, we'll want to put in the description of the dataset that null means that it wasn't reported.

kmax12 (approved):

gridstatus/eia.py
```diff
@@ -734,20 +753,62 @@ def _handle_fuel_type_data(df):
...
+    # (e.g. "Pumped Storage","Pumped storage"). We can handle that through the pivot
+    # table by summing the duplicates.
+    df["type-name"] = df["type-name"].str.lower()
+
     # pivot on type
     df = df.pivot_table(
         index=["Interval Start", "Interval End", "Respondent", "Respondent Name"],
         columns="type-name",
         values="MW",
+        aggfunc="sum",
```
WillKoehrsen, line 766: This assumes that "Pumped storage" and "Pumped Storage" can be added together if they are both present.

gridstatus/eia.py
```diff
@@ -734,20 +753,62 @@ def _handle_fuel_type_data(df):
...
-    df.columns.name = None
+    for col in known_fuel_mix_columns:
+        if col not in df.columns:
+            # This has to be np.nan not pd.NA because we are converting to float
+            df[col] = np.nan
+
+    fixed_cols = ["Interval Start", "Interval End", "Respondent", "Respondent Name"]
+    fuel_mix_cols = [col for col in df.columns if col not in fixed_cols]
+
+    df[fuel_mix_cols] = df[fuel_mix_cols].astype(float)
```
WillKoehrsen, line 802: Updated to not fill in NAs with 0.

gridstatus/eia.py
```diff
@@ -734,20 +753,69 @@ def _handle_fuel_type_data(df):
...
+    # table by summing the duplicates.
+    df["type-name"] = df["type-name"].str.lower()
+
+    # These columns are grouped together by EIA as confirmed by inspection of the EIA
+    # fuel mix graphs
+    df["type-name"] = df["type-name"].replace(
+        {
+            "battery": "battery storage",
+            "solar battery": "solar with integrated battery storage",
+            "unknown energy": "unknown energy storage",
```
WillKoehrsen, line 767: These combinations are what EIA does on their charts. It's possible (likely?) EIA will change these again or start publishing additional fuel types.

https://www.eia.gov/electricity/gridmonitor/expanded-view/electric_overview/US48/US48/GenerationByEnergySource-4/edit

kmax12: i think you can leave this comment in the code too

kmax12 (approved):

PR #476, opened 2024-11-11
feat[ENG 892] MISO Constraints

Kladar:
## Summary
This adds the 8 datasets (10 total data getter methods, 2 of which are historical backfill methods) that constitute MISO Constraints. 

### Details
This is a fair number of new datasets in a single PR and might be a bit of a chore to review, so it may make sense to break up these tasks into either a single dataset or cap it at 5 or something. Can discuss further if we feel it's an issue.

#### Dataset Handling
Each of these files has gotchas, from persistent typos and slight variations in column headers to varying disclaimer text rows to file types. I think I catch them all, but it makes the code look a bit unkempt IMO. LMK if you have improvement suggestions.  

Some of these datasets are sparse, with the files having no useful rows on many days. There are a couple ways to treat this, and I seem to include several without being sure what exactly the Grid Status opinionated way of doing it is.

For instance, we want to separate the idea of other failures (file doesn't exist, date range invalid, etc) from the "file exists and was properly parsed, but there was no constraint data for that day". Right now I handle this with some explicit returning of an empty dataframe with the correct columns:

```
if data.empty:
            return pd.DataFrame(
                columns=[
                    "Interval Start",
                    "Interval End",
                    "Constraint Name",
                    "Shadow Price",
                    "Constraint Description",
                ],
            )
``` 
and subsequently skip the testing of values in the tests:

```
assert list(df.columns) == [
                "Interval Start",
                "Interval End",
                "Constraint Name",
                "Shadow Price",
                "Constraint Description",
            ]

            if df.empty:
                pytest.skip(
                    "No data available for this date range, so skipping data-comparison assertions",
                )

            assert min(df["Interval Start"]).date() == pd.to_datetime(date).date()
```

However, the decorator `@support_date_range` throws an error if there is no data in any of the dfs collected when it tries to concatenate them, making these approaches only work for single day returns (the decorator knowing to eject if there is not actually a date range it needs to handle). 

CleanShot 2024-11-11 at 14 21 09@2x

I'm thinking we should handle the case of "successful file pull but empty" with the `@support_date_range` decorator, but that decorator is used across the codebase so I don't want to touch it here (repo-wide refactoring confidence is not high enough currently, IMO). The decorator could also use some modernization in other areas (f-strings, logging, etc), so that might just be a separate effort, and I don't want it to be a blocker for these datasets. 

#### Tests
Passing Locally (except for the `test_get_lmp_weekly_historical_date_range`, which I don't touch in this PR)
CleanShot 2024-11-11 at 14 37 39@2x

Currently this punts a bit on figuring out flatfile/binary file VCR fixtures. VCR is really good for standard API response formats (JSON, XML), but the cassette yaml files for binary files are a bit strange, large, and don't seem to have the same replayability as the JSON responses. The binary fixtures are not useless, just not as much of a speedup as a JSON fixture. For now the tests function as the integration style tests on the initial local run, after which you will have fixtures.

Right now I put the fixture creation code in, but add the cassette directory to the `.gitignore` since 
1. The binary files (e.g. their yaml fixtures) are large, not suitable for source control
2. The cassette yaml structure gets caught up on the precommit hook `check-yaml`

Really the "fixture saving" and the "raw file saving" start to look similar in a flatfile implementation. So I will continue to investigate csv, xls, etc binary file fixtures, but sidestep it here, probably until there is a strategy for raw response files saving in general. There is a MISO API that will likely continue to be developed, and future datasets should pull from that if they can, in which case this is a temporary problem and will be slowly remedied by shifting everything we can to pulling from JSON response APIs where vcrpy is exceptionally useful. However, these datasets are not yet available from the MISO API, per @cwaldoch's investigation. 

## Future Work
1. Shift what we can to the MISO API, when applicable
2. Limit Dataset addition tickets to a manageable number of datasets TBD
3. Figure out strategy for flatfile fixtures with or w/o vcrpy. 
4. Handle successful-but-empty dataframe returns, modernize the `@support_date_range` decorator


gridstatus/miso.py
```diff
@@ -679,8 +679,600 @@ def _get_generation_outages_data(
...
+
+        data["Interval End"] = pd.to_datetime(
+            market_date.strftime("%Y-%m-%d")
+            + " "
+            + data[
+                "Hour of  Occurrence"
+            ],  # NOTE(kladar): sic, there are two spaces between "Hour of" and "Occurrence"
+        ).dt.tz_localize(
+            self.default_timezone,
+        )  # TODO Confirm that these are hour_ending times
```
Kladar, line 995: @cwaldoch , confirming, is the "Hour of Occurrence" in the realtime 5 minute binding constraint interval-ending?

gridstatus/miso.py
```diff
@@ -679,8 +679,600 @@ def _get_generation_outages_data(
...
+            },
+        )
+        data = data.iloc[
+            :-2
+        ]  # NOTE(kladar): The last two rows are report descriptions and disclaimers
+        data["Interval End"] = pd.to_datetime(
+            data["Market Date"] + " " + data["Hour of Occurrence"],
+        ).dt.tz_localize(
+            self.default_timezone,
+        )  # TODO Confirm that these are hour_ending times
```
Kladar, line 1056: @cwaldoch , confirming, is the "Hour of Occurrence" in the realtime 5 minute historical binding constraint interval-ending?

Kladar:
Followup on "what to do with empty dfs" is that downstream systems may not want to try to use empty data.

WillKoehrsen:
> Followup on "what to do with empty dfs" is that downstream systems may not want to try to use empty data.

For empty data, we often `raise NoDataFoundException`. Downstream systems can handle that however they need to.

gridstatus/miso.py
```diff
@@ -679,8 +679,586 @@ def _get_generation_outages_data(
...
+
+        Args:
+            date (str | pd.Timestamp): Start date
+            end (str | pd.Timestamp, optional): End date. Defaults to None.
+            verbose (bool, optional): Verbosity. Defaults to False.
+
+        Returns:
+            pandas.DataFrame: Supplemental binding constraints data
+        """
+        query_date = date - pd.Timedelta("1D")
```
WillKoehrsen, line 701: Is this a day ahead dataset? Should we add `day_ahead` to the method name?

gridstatus/miso.py
```diff
@@ -679,8 +679,586 @@ def _get_generation_outages_data(
...
+                "Reason": object,
+            },
+        )
+
+        data["Interval End"] = pd.to_datetime(
+            market_date.strftime("%Y-%m-%d")
+            + " "
+            + data[
+                "Hour of  Occurrence"
+            ],  # NOTE(kladar): sic, there are two spaces between "Hour of" and "Occurrence"
```
WillKoehrsen, line 1118: Does this occur for all files? This code feels a little brittle cause MISO could fix this at some point. Could we instead do `data = data.rename(columns={"Hour of  Occurrence": "Hour of Occurrence"})`?

gridstatus/miso.py
Kladar: I'm going by the Data Spec https://www.notion.so/gridstatus/Binding-Constraints-Supplemental-130e835f42aa801da5abf4e91dbf10e9?pvs=4

gridstatus/miso.py
Kladar: It does not occur for all files, but occurs in more than one. 

Part of me wants to defend if they fix it (if they fix it I think a `rename` would also break, with "KeyError: Column not found", so would need a try/except or a if/else), part of me wants to wait for schema breaks because conceivably it's hard to defend against arbitrary column renaming in general. 

And part of me wants to think eventually all of this will come from the MISO API in the future so doing too much flatfile parsing might be headed for (eventual) obsolescence.

gridstatus/miso.py
```diff
@@ -679,8 +679,586 @@ def _get_generation_outages_data(
...
+        ).dt.tz_localize(
+            self.default_timezone,
+        )
```
WillKoehrsen, line 991: When calling this method, the data starts 5 minutes before the passed date and ends 5 minutes before the end of the date. Is it possible this line should be for the `Interval Start`?

Screenshot 2024-11-12 at 15 36 13

Kladar: I had a similar thought, but Connor confirmed that these are indeed interval ending. https://github.com/gridstatus/gridstatus/pull/476#discussion_r1837435304

Since they are individual date files (e.g. `...2024-11-10_rt_bc.xls`), unless we want to get fancy with multiple file pulls+data stitching for a single day, it seems it will be off by an interval due to the MISO convention.

WillKoehrsen: Alright, thanks. It's very weird to me that MISO publishes the last interval of the previous day with the next day's data.

gridstatus/miso.py
```diff
@@ -679,8 +679,586 @@ def _get_generation_outages_data(
...
+                "BP3": float,
+                "PC3": float,
+                "BP4": float,
+                "PC4": float,
+                "REASON": object,
+            },
+        )
+
+        # NOTE(kladar): The last row is a text disclaimer, and there is a leading space
+        # in the column names, so we clean it all up.
```
WillKoehrsen, line 884: Can we skip the last row with `skipfooter` in the `pd.read_csv` call? 

https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

WillKoehrsen (approved):

gridstatus/miso.py
Kladar:
```
/Users/kladar/GitHub/gridstatus/gridstatus/miso.py:865: ParserWarning: Falling back to the 'python' engine because the 'c' engine does not support skipfooter; you can avoid this warning by specifying engine='python'.
```
We can, but that changes to a different engine apparently which we need to explicitly call to avoid the warning. Oh pandas...

Not going to sweat it for now.

WillKoehrsen: Yeah, we specify the engine in a few places.

https://github.com/gridstatus/gridstatus/blob/74215686eb4d4aac1f8338ae5f03e2c7dc831467/gridstatus/miso.py#L294-L301

PR #481, opened 2024-11-19
MISO Support Yesterday for LMP Real Time 5 Min

WillKoehrsen:
## Summary

- Support getting yesterday's LMP Real Time 5 Minute Data

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -348,7 +348,7 @@ class MISO(ISOBase):
 
     @lmp_config(
         supports={
-            Markets.REAL_TIME_5_MIN: ["latest", "today"],
+            Markets.REAL_TIME_5_MIN: ["latest", "today", "historical"],
             Markets.DAY_AHEAD_HOURLY: ["today", "historical"],
             Markets.REAL_TIME_HOURLY_FINAL: ["historical"],
             Markets.REAL_TIME_HOURLY_PRELIM: ["historical"],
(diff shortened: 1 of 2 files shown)


gridstatus/miso.py
```diff
@@ -365,20 +365,27 @@ def get_lmp(
...
         if market == Markets.REAL_TIME_5_MIN:
             latest_url = "https://api.misoenergy.org/MISORTWDBIReporter/Reporter.asmx?messageType=currentinterval&returnType=csv"  # noqa
             today_url = "https://api.misoenergy.org/MISORTWDBIReporter/Reporter.asmx?messageType=rollingmarketday&returnType=csv"  # noqa
+            yesterday_url = "https://api.misoenergy.org/MISORTWDBIReporter/Reporter.asmx?messageType=previousmarketday&returnType=csv"  # noqa
 
             if date == "latest":
                 url = latest_url
             elif utils.is_today(date, tz=self.default_timezone):
                 url = today_url
+            elif utils.is_yesterday(date, tz=self.default_timezone):
```
WillKoehrsen, line 383: I don't think we should support the actual string "yesterday" because we don't do that with any other methods. Instead, we should require the caller to use yesterday's date.

kmax12: okay, that's fine

kmax12 (approved):

gridstatus/tests/source_specific/test_miso.py
```diff
@@ -79,23 +79,15 @@ def test_get_lmp_weekly_today_or_latest_raises(self):
         with pytest.raises(NotSupported):
             self.iso.get_lmp_weekly("today")
 
-    def test_get_lmp_weekly_historical_date(self):
```
WillKoehrsen, line 82: This test is redundant since we have the `historical_date_range` test.

PR #482, opened 2024-11-21
feat(ENG 944) IESO Resource Adequacy Report

Kladar:
## Summary
This adds the resource adequacy report from IESO, and is able to handle both past and future dates where they are available. Also includes some IESO quality-of-life, like type annotations and marking tests as integration. 

### Details
I pity the reviewer on this one. This is a pretty gnarly source and report, and there are definitely further optimizations and simplifications of code that can be made here, but wanted to get some eyes on it now that I've been looking at it for a few days. This report involves varied levels of nesting, XML cruft, evolving file names in the public listing, data updates over time, and a TBD data availability publishing and rolloff. That said, the current solution handles all the edgecases I've encountered, is mildly legible, avoids a lot of pandas mashing, and is performant. If we care about the output, then this surely gets us solid output. 90 glorious columns of it for every day currently available from the market operator (~119 of them).

If we care about process, it could definitely be improved, but also would likely build on the explicit structure provided already. 

The approach here is 
```
correct XML report -> JSON -> stdlib data structures via explicit data map -> output dataframe
```
I think that JSON -> dataframe step could be more direct, but I attempted several much shorter `read_xml() / read_json()` methods to try to use something off the shelf and all of them seemed to fall down in one way or another when getting to a wide-column format. A lot of these methods are fairly "all or nothing", inflexible in some way or another and either fails outright or leads to excessive subprocessing and postprocessing that ultimately nullified a lot of their upfront efficiency and sticking you in pandas world.

For example, even if a `pd.read_json()` worked as desired, there would be significant normalization and manual file path traversal such that explicitly mapping out and parsing various columns as I've done here seemed to be a bit more of an incremental approach that I could evolve deliberately to meet the report where it was, rather than stumble forward and hack through the thicket KeyError by KeyError. So that while _now_ we might be able to implement a slick two-liner somehow, it took the learnings and mappings herein to make that visible. The current approach looks pretty verbose, but I feel offers a fair amount of readability and flexibility for now. 

The structure map is a little hideous, but the method to the madness here is to try to keep straight all the nesting and overloaded terms in the report.

There are basically two styles of report: a completed, past report, and a future, forecast report:

**Past Report, values filled in**
CleanShot 2024-11-20 at 16 02 46@2x
[Source](https://reports-public.ieso.ca/public/Adequacy2/PUB_Adequacy2_20240724_v125.xml)

**Future Report, many null values**
CleanShot 2024-11-20 at 16 04 05@2x
[Source](https://reports-public.ieso.ca/public/Adequacy2/PUB_Adequacy2_20241224_v2.xml)

This method handles both, and fills in the columns with None where applicable, such that the resulting dataframe is always the same size (24 x 90). 

Tests passing locally
CleanShot 2024-11-20 at 16 19 22@2x

Unit-test only is very fast, once the fixtures are recorded. 
CleanShot 2024-11-20 at 16 20 51@2x
Punting on including fixtures (added to gitignore) until we can s3-ify them

There's some pretty verbose logging in there still, which we could pare down, but it's all already marked `DEBUG` so I'm tempted to leave it should this report ever change or need to be modified. We might be circling back as it gets used. Here's a short script one could run to see the outputs for oneself:

```
from gridstatus import IESO
import json
from datetime import datetime, timedelta
from typing import List, Tuple

ieso = IESO()

# Get an example report into JSON
json_data = ieso._get_resource_adequacy_json("2024-09-09")
with open("ieso_resource_adequacy_report.json", "w") as f:
    json.dump(json_data, f)

# Collect all reports in a given date range and save them as CSVs
def collect_ieso_reports(start_date: str, end_date: str) -> Tuple[List[str], List[str]]:
    
    failed_dates: List[str] = []
    successful_dates: List[str] = []
    
    start = datetime.strptime(start_date, "%Y-%m-%d")
    end = datetime.strptime(end_date, "%Y-%m-%d")
    
    current = start
    while current <= end:
        date_str = current.strftime("%Y-%m-%d")
        try:
            df = ieso.get_resource_adequacy_report(date_str)
            df.to_csv(f"ieso_adequacy_report_{date_str}.csv", index=False)
            successful_dates.append(date_str)
        except Exception as e:
            failed_dates.append(date_str)
        
        current += timedelta(days=1)
    
    return successful_dates, failed_dates

successful, failed = collect_ieso_reports("2024-11-21", "2024-12-24") # Modify these
print(f"Successfully processed dates: {len(successful)}")
print(f"Failed dates: {failed}")
```

I've QA'd a significant number of values at I've been going, and fairly sure everything's good, but spot checking some stuff would be a good idea I think. Here's the source URL: https://reports-public.ieso.ca/public/Adequacy2/

Adding @cwaldoch for his visibility.


gridstatus/ieso.py
```diff
@@ -844,3 +909,758 @@ def _request(self, url, verbose):
...
+        json_data = self._get_resource_adequacy_json(date)
+        return self._parse_resource_adequacy_report(json_data)
+
+    # Note(Kladar): This might be fairly generalizable to other XML reports from IESO
+    def _get_resource_adequacy_json(
+        self,
+        date: str | datetime.date | datetime.datetime,
+    ) -> dict:
+        """Retrieve the Resource Adequacy Report for a given date and convert to JSON. There are often many
+        files for a given date, so this function will return the file with the highest version number. It does
```
WillKoehrsen, line 934: It seems likely we want to retrieve multiple versions of a report for a given date. I think we should support retrieving all versions in addition to the latest version.

Kladar: Okay, I've implemented that and will continue to test it out. 

It seems like the unversioned number (if it exists) also constitutes the latest of a date, but still confirming that and will update to handle that file as the latest. (It should be handled in the _all method now, but not yet the _latest method)

Kladar: Okay, I've seen enough to confirm that the unversioned file is also equivalent to the highest version number, and that it constitutes the `latest`
[ieso_versions_2024-09-08.csv](https://github.com/user-attachments/files/17849878/ieso_versions_2024-09-08.csv)
[ieso_versions_2024-09-09.csv](https://github.com/user-attachments/files/17849880/ieso_versions_2024-09-09.csv)
[ieso_versions_2024-12-01.csv](https://github.com/user-attachments/files/17849882/ieso_versions_2024-12-01.csv)

WillKoehrsen: Nice. That's what I saw as well.

gridstatus/ieso.py
```diff
@@ -844,3 +911,883 @@ def _request(self, url, verbose):
             )
 
         return r
+
+    def get_resource_adequacy_report(
```
WillKoehrsen, line 915: Do we want to decorate this with `@support_date_range`? We can use `@support_date_range(frequency="DAY_START")` then the function should work for date ranges.

Kladar: definitely. added along with some new tests to cover the ranges and vintage options.

gridstatus/ieso.py
```diff
@@ -844,3 +911,887 @@ def _request(self, url, verbose):
...
+                            ],
+                            column_name=section_name,
+                            value_key=section_config["value_key"],
+                            report_data=report_data,
+                        )
+
+        # NOTE(kladar): This is the first place where pandas is truly invoked, leaving it open for more modern
+        # dataframe libraries to be swapped in in the future
+        df = pd.DataFrame(report_data)
+        df["Interval Start"] = delivery_date + pd.to_timedelta(
```
WillKoehrsen, line 1269: I think we want to add the timezone to "Interval Start", "Interval End", and "Publish Time".

Kladar: sounds good. 

I've added that when they are parsed from the document and moved that parsing to the rest of the date handling here as well.

WillKoehrsen (approved):
The data looks good to me. We might want to wait to merge until we've confirmed this works with our ETL code.

Kladar:
I guess I probably don't need the publish_time functionality with `vintage=latest`, since it wall always return the latest publish time.

Kladar:
changed it to `last_modified` since that's technically what it is (publish time is different, only available in the report itself).

gridstatus/ieso.py
```diff
@@ -844,3 +910,788 @@ def _request(self, url, verbose):
             )
 
         return r
+
+    @support_date_range(frequency="DAY_START")
+    def get_resource_adequacy_report(
+        self,
+        date: str | datetime.date | datetime.datetime,
+        end: datetime.date | datetime.datetime | None = None,
+        vintages: Literal["all", "latest"] = "latest",
```
WillKoehrsen, line 919: In `ISONEAPI` methods, we use singular "vintage" for the parameter. Should we be consistent?

gridstatus/ieso.py
```diff
@@ -844,3 +910,788 @@ def _request(self, url, verbose):
...
 
         return r
+
+    @support_date_range(frequency="DAY_START")
+    def get_resource_adequacy_report(
+        self,
+        date: str | datetime.date | datetime.datetime,
+        end: datetime.date | datetime.datetime | None = None,
+        vintage: Literal["all", "latest"] = "latest",
+        last_modified: str | datetime.date | datetime.datetime | None = None,
```
WillKoehrsen, line 920: Do you think we should add `Last Modified` as a column to the returned dataframe for auditing? Might be helpful to have if there are issues.

Kladar: Yeah I think it could help. There is also an issue with the `vintage="all"` in that the unversioned file and the highest versioned file contain the same data, which results in duplicates. 

I'm thinking for all vintages leaving behind the unversioned file, with the knowledge that the versioned files contain all the unique data.

Kladar: Conceivably a `drop_duplicates()` at the end of the processing would also work

gridstatus/ieso.py
Kladar: done.

gridstatus/ieso.py
Kladar: added `last_modified` as a column to the returned df.

Kladar: Okay, added last modified as a column and will do a drop duplicates in places that need it.

WillKoehrsen: Sorry, what I meant for "Last Modified" was the value associated with the file from this page: https://reports-public.ieso.ca/public/Adequacy2/

Kladar: Ah. 

Okay I've implemented this, I believe it's working correctly.

WillKoehrsen (approved):
Re-approving after the recent changes. This is a very complicated dataset, and the approach taken here works well for filtering data.

PR #483, opened 2024-11-21
EIA Fix Grid Monitor (CO2 Emissions)

WillKoehrsen:
## Summary

- Fixes `EIA().get_grid_monitor()` because EIA changed column names

### DIFF
--- a/gridstatus/tests/source_specific/test_eia.py
+++ b/gridstatus/tests/source_specific/test_eia.py
@@ -1,6 +1,5 @@
 import numpy as np
 import pandas as pd
-import pytest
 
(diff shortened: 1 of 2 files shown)


gridstatus/tests/source_specific/test_eia.py
```diff
@@ -272,7 +271,6 @@ def test_get_coal_spots():
     assert d["coke_exports"].shape[0] > 0
 
 
-@pytest.mark.slow
```
WillKoehrsen, line 275: The reason I didn't catch this earlier was this test wasn't running!

PR #485, opened 2024-11-22
EIA Regional Data Keep NAs

WillKoehrsen:
## Summary

- Keep NA values in Ercot regional data

### DIFF
--- a/gridstatus/eia.py
+++ b/gridstatus/eia.py
@@ -663,6 +663,7 @@ def _handle_region_data(df):
         index=["Interval Start", "Interval End", "Respondent", "Respondent Name"],
         columns="Type",
         values="MW",
+        dropna=False,
     ).reset_index()
 
     df.columns.name = None
--- a/gridstatus/eia.py
+++ b/gridstatus/eia.py
@@ -658,12 +658,13 @@ def _handle_region_data(df):
 
     df["MW"] = df["MW"].astype(float)
 
-    # pivot on type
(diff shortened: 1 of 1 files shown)


gridstatus/eia.py
```diff
@@ -663,6 +663,7 @@ def _handle_region_data(df):
         index=["Interval Start", "Interval End", "Respondent", "Respondent Name"],
         columns="Type",
         values="MW",
+        dropna=False,
```
WillKoehrsen, line 666: Should we also do this for the fuel data?

gridstatus/eia.py
```diff
@@ -658,8 +658,10 @@ def _handle_region_data(df):
 
     df["MW"] = df["MW"].astype(float)
 
-    # pivot on type
-    df = df.pivot_table(
+    df = df[df["Type"].notna()]
+
+    # pivot on Type
+    df = df.pivot(
```
WillKoehrsen, line 664: Pivot is the right choice here because we do not expect duplicates on the columns specified by `index`. `pivot` also does not drop nulls. `pivot_table` is used for aggregations of duplicates (like we do [here](https://github.com/gridstatus/gridstatus/blob/2407391dc446b8c28a327730c035266452560bd6/gridstatus/eia.py#L757-L778)).

https://pandas.pydata.org/docs/user_guide/reshaping.html#reshaping

kmax12 (approved):

Issue #487, opened 2024-11-25
IESO verify=False change to True


pjrulez:
Unverified HTTPS request is being made to host 'reports-public.ieso.ca'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings

ieso.py:

```
    def _request(self, url, verbose):
        msg = f"Fetching URL: {url}"
        log(msg, verbose)

        max_retries = 3
        retry_num = 0
        sleep = 5

        while retry_num < max_retries:
            r = requests.get(url, **verify=False**) # TODO change to True

            if r.ok:
                break
...
```

WillKoehrsen:
Thanks for the suggestion. I tried that file and got the same SSL error that caused us to use `verify=False`.

```python
import requests
requests.get('https://ieso.ca/-/media/Files/IESO/uploaded/Chart/ontario_demand_multiday.xml')

SSLError: HTTPSConnectionPool(host='ieso.ca', port=443): Max retries exceeded with url: /-/media/Files/IESO/uploaded/Chart/ontario_demand_multiday.xml (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)')))
```

It's interesting this URL fails as well. Using `curl` to access the URL does work even though curl does SSL verification by default (`curl https://ieso.ca/-/media/Files/IESO/uploaded/Chart/ontario_demand_multiday.xml`). The fact that curl works led me to open https://github.com/gridstatus/gridstatus/pull/488

WillKoehrsen:
This was finally fixed by #496.

PR #488, opened 2024-11-25
[ENG-965] Fix SSL Verification with IESO URL

WillKoehrsen:
## Summary

- Fixes the issue with no verification for IESO urls by using PycURL for .ashx URLs
- Addresses https://github.com/gridstatus/gridstatus/issues/487

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -1,7 +1,9 @@
 import time
 import xml.etree.ElementTree as ET
+from io import BytesIO
 
 import pandas as pd
+import pycurl
 import requests
 
 from gridstatus import utils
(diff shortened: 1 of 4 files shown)


gridstatus/ieso.py
```diff
@@ -1,7 +1,9 @@
 import time
 import xml.etree.ElementTree as ET
+from io import BytesIO
 
 import pandas as pd
+import pycurl
```
WillKoehrsen, line 6: I don't know why PycURL is able to access the `.ashx` URL with SSL verification while requests cannot. I tried:

* `requests` with/without [`certifi`](https://pypi.org/project/certifi/) 
* `urllib3` with/without `certifi`
* `http.client` with/without `certifi`

Something about this specific URL causes issues with `requests`. I noticed the URL works fine with cURL (which does SSL verification by default) so found a library that uses cURL.

WillKoehrsen:
Since the other URL also requires the same PycURL approach, I think we can merge this.

https://github.com/gridstatus/gridstatus/issues/487#issuecomment-2502410001

WillKoehrsen:
Closing in favor of #496.

PR #489, opened 2024-11-25
Fix Tests for is_today and is_yesterday

WillKoehrsen:
## Summary

- The tests for `is_today` and `is_yesterday` are broken in that they fail at certain times of day, when the EST/CST date is different from the UTC date
- This fixes the tests by freezing time using time_machine and makes the tests more robust by checking both times when the EST/CST date is the same and different from the UTC date


Kladar:
+1 for this

Kladar:
Getting some early ieso test failures. Reran to make sure they are persistent. https://github.com/gridstatus/gridstatus/actions/runs/12014081185/job/33552618134?pr=489

Kladar:
Still getting some random iso-related failures, now it's NYISO but seems fairly unrelated/minor.

WillKoehrsen:
> Still getting some random iso-related failures, now it's NYISO but seems fairly unrelated/minor.

This PR doesn't change any source code so I don't think it would cause ISO specific tests to fail.

Kladar:
> > Still getting some random iso-related failures, now it's NYISO but seems fairly unrelated/minor.
> 
> This PR doesn't change any source code so I don't think it would cause ISO specific tests to fail.

Agreed.

Kladar (approved):

PR #491, opened 2024-11-26
fix: Ignore fixtures

Kladar:
## Summary
Ignores the VCR fixtures for now

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,8 +61,11 @@ coverage.xml
 .pytest_cache/
 cover/
 mypy_report/
-# VCR cassettes for binary files, for now until we can figure out how to handle them
+# Ignore local vcr cassette fixtures
 gridstatus/tests/fixtures/miso/vcr_cassettes
+gridstatus/tests/fixtures/isone/vcr_cassettes
+gridstatus/tests/fixtures/pjm/vcr_cassettes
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -61,8 +61,11 @@ coverage.xml
 .pytest_cache/
 cover/
 mypy_report/
-# VCR cassettes for binary files, for now until we can figure out how to handle them
+# Ignore local vcr cassette fixtures
 gridstatus/tests/fixtures/miso/vcr_cassettes
+gridstatus/tests/fixtures/isone/vcr_cassettes
```
WillKoehrsen, line 66: IIUC, any cassettes already in `git` will have their changes tracked even with these updates to `.gitignore`. So there could be a situation where you re-recorded the cassettes already in `git` locally then committed new creds. We'd have to delete the cassettes and then this `gitignore` should prevent them from being re-added.

Kladar: Ah okay. I will make sure to pull in `main` to any active branches and delete fixtures from my local, and not add or commit any fixtures going forward until we are sure.

WillKoehrsen: I think we should be okay after we merge #492 but if we want to just remove the fixtures for now, I'm fine with that too. We can still use them locally.

WillKoehrsen (approved):

PR #493, opened 2024-12-02
[ENG-939] MISO API Pricing Data

WillKoehrsen:
## Summary

- Adds initial support for the MISO API, specifically the pricing product
- Adds Methods for retrieving real time and day ahead LMP prices

### Details

- Methods
1. `get_lmp_day_ahead_hourly_ex_ante`
2. `get_lmp_day_ahead_hourly_ex_post`
3. `get_lmp_real_time_hourly_ex_post_prelim`
4. `get_lmp_real_time_hourly_ex_post_final`
5. `get_lmp_real_time_5_min_ex_ante`
6. `get_lmp_real_time_5_min_ex_post_prelim`
7. `get_lmp_real_time_5_min_ex_post_final`


gridstatus/miso_api.py
```diff
@@ -0,0 +1,321 @@
...
+        return df
+
+    def _get(self, url, verbose=False, max_retries=3):
+        headers = self._headers()
+        data_list = []
+
+        if verbose:
+            logger.info(f"Getting data from {url}")
+
+        response = requests.get(url, headers=headers, verify=False)
```
WillKoehrsen, line 263: Unfortunately, this doesn't work unless `verify=False`. The error is:

```
(Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)')))
```

I've tried using `verify=certifi.where()` which fails with the same error. We don't want to skip validation so I'm looking into other solutions.

.env.template
```diff
@@ -11,3 +11,7 @@ PJM_API_KEY=
 ERCOT_API_USERNAME=
 ERCOT_API_PASSWORD=
 ERCOT_API_SUBSCRIPTION_KEY=
+
+# Sign up here and subscribe to the pricing product
+# https://data-exchange.misoenergy.org/apis
```
WillKoehrsen, line 16: The Sign Up link here just redirects to the Sign In page so I don't actually know how to create a new account.

gridstatus/miso_api.py
```diff
@@ -0,0 +1,321 @@
...
+            end,
+            retrieval_func=self._get_lmp_real_time_5_min_ex_post,
+            market=Markets.REAL_TIME_5_MIN_EX_POST_FINAL,
+            prelim_or_final=FINAL_STRING,
+            verbose=verbose,
+        )
+
+    # NOTE: this method does not use the support_date_range decorator. Instead
+    # it takes the output of a decorated function and processes that output all at once
+    # which is more efficient than processing each iteration of the decorator
```
WillKoehrsen, line 109: Hopefully, this is clear. We don't want to call `process_pricing_data` on each loop of the decorator function, so, we return the raw data lists instead then combine them into one list and do all the processing at once.

gridstatus/miso_api.py
WillKoehrsen: Running a test against the server shows an incomplete chain of certificates. It seems that a browser is able to find those missing certificates, but requests does not.

https://www.ssllabs.com/ssltest/analyze.html?d=apim.misoenergy.org
https://stackoverflow.com/a/66111417/5755357

Screenshot 2024-12-01 at 20 25 37

gridstatus/miso_api.py
```diff
@@ -0,0 +1,322 @@
...
+        df["Interval Start"] = pd.to_datetime(df["start"]).dt.tz_localize(
+            self.default_timezone,
+        )
+        df["Interval End"] = pd.to_datetime(df["end"]).dt.tz_localize(
+            self.default_timezone,
+        )
+
+        node_to_type_mapping = (
+            MISO()
+            ._get_node_to_type_mapping()
```
kmax12, line 222: our approach here is to query the `{today}_da_expost_lmp.csv` file for to obtain node maps. this feels unreliable for historical days when nodes may have been different. that file seems to have data going back to `20220101`. perhaps we can can pass the date of the end of the requested range (but not earlier than `20220101`) to this function to get a better file.

gridstatus/base.py
```diff
@@ -26,11 +26,25 @@ class Markets(Enum):
...
+    REAL_TIME_5_MIN_EX_ANTE = "REAL_TIME_5_MIN_EX_ANTE"
+
+    REAL_TIME_5_MIN_EX_POST_PRELIM = "REAL_TIME_5_MIN_EX_POST_PRELIM"
+    REAL_TIME_5_MIN_EX_POST_FINAL = "REAL_TIME_5_MIN_EX_POST_FINAL"
     REAL_TIME_5_MIN_FINAL = "REAL_TIME_5_MIN_FINAL"
+
     REAL_TIME_15_MIN = "REAL_TIME_15_MIN"
     REAL_TIME_HOURLY = "REAL_TIME_HOURLY"
+
+    REAL_TIME_HOURLY_EX_POST_PRELIM = "REAL_TIME_HOURLY_EX_POST_PRELIM"
```
kmax12, line 39: should we update `MISO` to correctly use these new market names rather than `Markets.REAL_TIME_HOURLY_FINAL`, `Markets.REAL_TIME_HOURLY_PRELIM`, `Markets.DAY_AHEAD_HOURLY`? 

perhaps we also consider making a breaking change downstream in our api to just get the "market" column and the dataset names correct

WillKoehrsen: > should we update MISO to correctly use these new market names

Based on our discussion, it sounds like no.

gridstatus/miso_api.py
WillKoehrsen: I pulled the data from 2022-01-01 and 2024-12-11 and compared the mapping of `Node` to `Type`. There were no differences. (The 2024 data has more nodes but none of the mappings have changed).

gridstatus/decorators.py
```diff
@@ -40,10 +40,11 @@ def date_range_maker(start, end, freq, inclusive="neither"):
 # current or latest endpoints that are automatically handled. Currently cannot refactor this confidently
 # without improved testing since it touches many methods
 class support_date_range:
-    def __init__(self, frequency, update_dates=None):
+    def __init__(self, frequency, update_dates=None, convert_to_dataframe=True):
```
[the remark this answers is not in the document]
WillKoehrsen: Updated to `return_raw=False`

PR #497, opened 2024-12-05
tune(ENG 914) VCR Setup (Part 1)

Kladar:
## Summary
This PR is the first in a series to move to a unit-test-with-fixtures setup, which allows us to run more and more targeted tests more often. This is another brick in the road for eventually being able to do more radical refactors without breaking downstream systems.

#### Commit-level testing
First, it splits up the commit-level testing from the PR-level testing, setting a course for unit-level testing to run automatically on commit in GitHub Actions. 

You can install the GitHub Actions VSCode extension and get the test run right in your IDE. Once the unit style tests are truly independent of their external dependencies for these tests, it will provide a valuable fast feedback loop for developers and eliminate the need to run tests manually. 
CleanShot 2024-12-05 at 09 00 29@2x

It also allows you to jump right to the GitHub Action in the browser from your IDE.
CleanShot 2024-12-05 at 09 02 45@2x

The PR/main branch tests are unchanged, except for making the ISO-specific tests not "fail fast", meaning if one ISO fails the others will still run. We were running into a bit of an issue that if a unrelated ISO test failed, the rest of the ISO testing would cancel, which is not super helpful and usually those then had to be re-run. 

#### Marking integration tests
It also goes through and marks all tests currently not using a fixture for an external dependency. It adds a VCR setup to each source-specific file for now, such that we can slowly over time update existing tests and develop new tests that use the fixture regime. 

#### Unit-only GitHub action
Lastly, it adds the "unit-only" command to the Makefile and to the commit-level testing. As mentioned above, this unit-level testing is still running in integration mode until the fixtures can be pulled from s3, which is tracked here [ENG-993](https://www.notion.so/gridstatus/Enable-github-actions-to-access-test-fixtures-on-s3-153e835f42aa80aa8a68e16dc5fd7b03?pvs=4)

### Current Testing Stats
Total counts attempts to remove parameterizations, so if a single test has 3 different parameter values, it tries to count it as 1 test rather than 3. The default for `pytest --collect-only` is to count parameterizations as separate tests, which is less useful for this exercise. These are estimates:

| Category | Count |
|-----------------------------------------------|-------|
| Total Tests Collected by Pytest | 611 |
| Source-Specific Tests (ISO-related) | 596 |
| Source-Specific Integration Tests (w/ params)| 422 |
| Source-Specific Integration Tests (unique) | 398 |
| Tests Using a VCR Fixture (via grep) | 31 |
| Total Fixturable Tests | 429 |
| Percentage of Fixturable Tests Using VCR | ~7% |


.github/workflows/tests-commit.yaml
```diff
@@ -0,0 +1,37 @@
+name: Commit Tests
+on: # Run on all commit pushes to all feature branches.
```
WillKoehrsen, line 2: This runs on all commits regardless of whether the branch is in a PR?

Kladar: yes

.gitignore
```diff
@@ -67,6 +67,7 @@ gridstatus/tests/fixtures/isone/vcr_cassettes
 gridstatus/tests/fixtures/pjm/vcr_cassettes
 gridstatus/tests/fixtures/ercot_api/vcr_cassettes
 gridstatus/tests/fixtures/ieso/vcr_cassettes
+gridstatus/tests/fixtures/isone_api/vcr_cassettes
```
WillKoehrsen, line 70: Would it make more sense to use `**/vcr_cassettes` for now? Seems like individually listing the paths is more error prone.

WillKoehrsen (approved):

.gitignore
Kladar: agreed, updated.

PR #498, opened 2024-12-05
Rename Columns for PJM Load Forecast Hourly

WillKoehrsen:
## Summary

- Replace '&' with '' and '\' with '_' in PJM Load Forecast Hourly Data


Kladar (approved):

Kladar:
The ISONE test (and the ISO-based test "fail fast" that stops the other ISOs from running when one fails) is fixed in https://github.com/gridstatus/gridstatus/pull/497

PR #501, opened 2024-12-09
chore(ENG 1005): rename to `load forecast`

Kladar:
## Summary
Renames the `Load` column in `isone_reliability_region_load_forecast` to `Load Forecast` to follow the convention set across other ISOs and datasets.

### Details
This also updates the methods for `ISONEAPI.get_hourly_load_forecast()`, which is the API method for hourly load forecast. There is also a "flat file" scraper for this data which is more heavily used since it came first. More in this ticket on ISONE API implementation https://www.notion.so/gridstatus/Future-ISONE-Datasets-122e835f42aa806f8f0adec85ac3e6d2?pvs=4

### DIFF
--- a/gridstatus/isone_api/isone_api.py
+++ b/gridstatus/isone_api/isone_api.py
@@ -613,15 +613,15 @@ class ISONEAPI:
             "Interval End": "Interval End",
             "CreationDate": "Publish Time",
             "ReliabilityRegion": "Location",
-            "LoadMw": "Load",
+            "LoadMw": "Load Forecast",
             "ReliabilityRegionLoadPercentage": "Regional Percentage",
         }
         system_cols = {
             "BeginDate": "Interval Start",
             "Interval End": "Interval End",
             "CreationDate": "Publish Time",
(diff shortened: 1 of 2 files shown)


gridstatus/isone_api/isone_api.py
```diff
@@ -613,15 +613,15 @@ def _handle_load_forecast(
...
             "ReliabilityRegionLoadPercentage": "Regional Percentage",
         }
         system_cols = {
             "BeginDate": "Interval Start",
             "Interval End": "Interval End",
             "CreationDate": "Publish Time",
-            "LoadMw": "Load",
-            "NetLoadMw": "Net Load",
+            "LoadMw": "Load Forecast",
+            "NetLoadMw": "Net Load Forecast",
```
WillKoehrsen, line 624: Do we need to update the corresponding tests?

Kladar: yar

WillKoehrsen (approved):

PR #502, opened 2024-12-11
fix: Disambiguate `last_modified` timezone better

Kladar:
## Summary
Before, we were localizing the given `last_modified` time parameter into the ISO local time (always EST). However, this is not the desired behavior and really `last_modified` can have any (or no) timezone.

### Details
Particularly this would cause an issue if the last modified was meant to be in UTC, as it was localized to EST, resulting in a radically different time than was given to the function and used to compare.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -1081,7 +1081,10 @@ class IESO(ISOBase):
 
         if last_modified:
             if last_modified.tz is None:
-                last_modified = last_modified.tz_localize("UTC")
+                last_modified = utils._handle_date(
+                    last_modified,
+                    tz=self.default_timezone,
+                )
             filtered_files = [
                 (file, time)
                 for file, time in file_rows
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -998,7 +998,10 @@ class IESO(ISOBase):
 
         if last_modified:
             if last_modified.tz is None:
-                last_modified = last_modified.tz_localize(self.default_timezone)
+                last_modified = utils._handle_date(
+                    last_modified,
+                    tz=self.default_timezone,
+                )
             filtered_files = [
                 (file, time)
                 for file, time in files_and_times
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -998,7 +998,7 @@ class IESO(ISOBase):
 
         if last_modified:
             if last_modified.tz is None:
-                last_modified = last_modified.tz_localize("UTC")
+                last_modified = last_modified.tz_localize(self.default_timezone)
             filtered_files = [
                 (file, time)
                 for file, time in files_and_times
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -964,7 +964,7 @@ class IESO(ISOBase):
     def _get_latest_resource_adequacy_json(
         self,
         date: str | datetime.date | datetime.datetime,
-        last_modified: str | datetime.date | datetime.datetime | None = None,
+        last_modified: pd.Timestamp | None = None,
     ) -> tuple[dict, datetime.datetime]:
         """Retrieve the Resource Adequacy Report for a given date and convert to JSON. There are often many
         files for a given date, so this function will return the file with the highest version number. It does
@@ -997,7 +997,8 @@ class IESO(ISOBase):
             )
 
         if last_modified:
-            last_modified = pd.Timestamp(last_modified, tz=self.default_timezone)
+            if last_modified.tz is None:
+                last_modified = last_modified.tz_localize("UTC")
             filtered_files = [
                 (file, time)
                 for file, time in files_and_times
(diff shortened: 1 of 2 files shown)


Kladar:
Looks like the files have rolled off, causing tests to fail 
CleanShot 2024-12-10 at 16 29 56@2x

I will update the tests

gridstatus/tests/source_specific/test_ieso.py
```diff
@@ -549,12 +549,13 @@ def _check_get_generator_report_hourly(self, df):
     """get_resource_adequacy_report"""
 
     # NOTE(kladar): we will see how future data rolls in and historical rolls off
+    # NOTE(kladar, 2024-12-11): Tests rolled off, earliest is currently 2024-09-10, 92 days ago
     RESOURCE_ADEQUACY_TEST_DATES = [
         (
-            (pd.Timestamp.now(tz=default_timezone) - pd.Timedelta(days=119)).strftime(
+            (pd.Timestamp.now(tz=default_timezone) - pd.Timedelta(days=92)).strftime(
```
Kladar, line 555: i'll continue to dial this in

gridstatus/tests/source_specific/test_ieso.py
```diff
@@ -581,10 +582,10 @@ def _check_get_generator_report_hourly(self, df):
             ),
         ),
         (
-            (pd.Timestamp.now(tz=default_timezone) + pd.Timedelta(days=28)).strftime(
+            (pd.Timestamp.now(tz=default_timezone) + pd.Timedelta(days=31)).strftime(
                 "%Y-%m-%d",
             ),
-            (pd.Timestamp.now(tz=default_timezone) + pd.Timedelta(days=30)).strftime(
+            (pd.Timestamp.now(tz=default_timezone) + pd.Timedelta(days=34)).strftime(
```
Kladar, line 588: this is the farthest out it forecasts, according to the data spec from the ISO

gridstatus/ieso.py
```diff
@@ -997,7 +997,8 @@ def _get_latest_resource_adequacy_json(
             )
 
         if last_modified:
-            last_modified = pd.Timestamp(last_modified, tz=self.default_timezone)
+            if last_modified.tz is None:
+                last_modified = last_modified.tz_localize("UTC")
```
WillKoehrsen, line 1001: I don't understand the conversion to UTC. Aren't the last modified times from [this page](https://reports-public.ieso.ca/public/Adequacy2/?C=M;O=D) in EST? 

`date` and `end` are converted to EST so it would be inconsistent to convert `last_modified` to UTC.

gridstatus/ieso.py
```diff
@@ -997,7 +997,8 @@ def _get_latest_resource_adequacy_json(
             )
 
         if last_modified:
-            last_modified = pd.Timestamp(last_modified, tz=self.default_timezone)
+            if last_modified.tz is None:
```
kmax12, line 1000: if needed, maybe you should raise an error if a tz-naive timestamp is passed. 

if you don't like the raising an error approach and need to pick a timezone, we generally assume tz-naive timestamps are in the the default timezone of the market throughout this library

gridstatus/ieso.py
WillKoehrsen: To clarify, when a function is decorated with `@support_date_range`, the `date` and `end` parameters get converted to the local timezone using the `utils._handle_date` method. So, maybe we should pass `last_modified` through `utils._handle_date`.

https://github.com/gridstatus/gridstatus/blob/88acf30dcefa0e7c6941ac5fc04193fa83c41718/gridstatus/decorators.py#L95-L108

Kladar: Yeah mostly I was converting to UTC here to make it unambiguous and explicit. The issue that was occurring was that a (timezone-naive) UTC time was being given to the kwarg `last_modified` which was then being (incorrectly) localized to local time here, resulting in no data. 

So while yes the last modified times from the page are in EST, the date comparison used to filter out the files can compare two timestamps in different timezones correctly, as long as there is a known timezone to each. (If one is timezone-naive and the other is timezone-aware, the date comparison will error out with the classic `Error: Cannot convert tz-naive Timestamp, use tz_localize to localize`, which is why the localization was added here in the first place). But if the naive time is actually in UTC, then localizing to EST is not correct. 

So the options are:
1. Expect a timezone-aware time, error if not
2. Handle a timezone-naive time to something (but only if it is naive). 

I opt for 2, but can change that handling to EST rather than UTC (perhaps using `utils._handle_date`).

gridstatus/ieso.py
Kladar: Yeah I think either is fine, I'm doing the latter now albeit to UTC, which I can change to EST (via `self.default_timezone`).

gridstatus/ieso.py
kmax12: > I opt for 2, but can change that handling to EST rather than UTC (perhaps using utils._handle_date).

this would be consistent with the rest of the library.

gridstatus/tests/source_specific/test_ieso.py
Kladar: dialing in continues...

gridstatus/ieso.py
```diff
@@ -1076,7 +1080,8 @@ def _get_all_resource_adequacy_jsons(
...
+                last_modified = last_modified.tz_localize("UTC")
```
WillKoehrsen, line 1084: Do we want to do the same handling here as above?

Kladar: oops, yes.

Kladar: updated.

WillKoehrsen (approved):
Looks good.

PR #503, opened 2024-12-12
Update url base for caiso outlook

kmax12:
## Summary
The URLs changed

### DIFF
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -24,8 +24,8 @@ from gridstatus.decorators import support_date_range
 from gridstatus.gs_logging import log
 from gridstatus.lmp_config import lmp_config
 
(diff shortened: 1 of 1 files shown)


kmax12:
@Kladar I see the caiso test are passing. is there any thing I need to do to update fixtures?

WillKoehrsen (approved):

Kladar (approved):
I haven't looked too closely, but looks like the other ISO tests are failing for various unrelated reasons.  

For fixtured tests you can change the env var `RECORD_MODE` to be `="all"` to re-record everything if you ever wanted, but this would just be for your local machine at this time and not effect the GH actions tests. Deleting the fixtures from your machine would have the same effect, as it would notice they aren't there and record them.

PR #504, opened 2024-12-13
feat(ENG 978): ERCOT Indicative LMPs

Kladar:
## Summary
Adds the Indicative LMPs by Settlement Point. Adds methods for both current and historical data getting. Also continuing quality of life.

_**Question 1: Is there more functionality I should add to the historical API method to limit the amount of data it returns?**_

_**Question 2: Should I trim the data such that the `Interval Start` is the argument date only? Right now, because it operates on publish time, we also get the following ~hour**_

Example for 2021-12-08
CleanShot 2024-12-12 at 16 10 38@2x

#### Data Breakdown
It looks like ERCOT publishes data for ~944 locations every 5 minutes, for each 5 minute interval of an hour leading up to an interval. These are distinguished by `Interval ID`, and can also be seen in the `RTDTimestamp`, which stands in for a publish time. Here's a quick "horizon analysis", with the difference between our publish time and the `Interval End` time, grouping by `Interval ID`:

| Interval Id | min | max | mean |
|------------|-----|-----|------|
| 1 | 4.933333 | 4.95 | 4.94919 |
| 2 | 9.933333 | 9.95 | 9.94919 |
| 3 | 14.933333 | 14.95 | 14.94919 |
| 4 | 19.933333 | 19.95 | 19.94919 |
| 5 | 24.933333 | 24.95 | 24.94919 |
| 6 | 29.933333 | 29.95 | 29.94919 |
| 7 | 34.933333 | 34.95 | 34.94919 |
| 8 | 39.933333 | 39.95 | 39.94919 |
| 9 | 44.933333 | 44.95 | 44.94919 |
| 10 | 49.933333 | 49.95 | 49.94919 |
| 11 | 54.933333 | 54.95 | 54.94919 |

This means that we have 
```
288 intervals/day x 944 locations x 11 horizons/interval = 2,990,952 intervals per day of data
```

The historical data goes back to June 27, 2012, so we have ~4,550 days of data. A full historical backfill of this data is ~13.6 billion rows.

#### Tests
Scrubbing of credentials is working nicely. Still having some trouble with DST transitions though. 
```
Error: Cannot infer dst time from 2024-11-03 01:55:03, try using the 'ambiguous' argument
Args: {'self': <gridstatus.ercot_api.ercot_api.ErcotAPI object at 0x1024b7f20>, 'date': Timestamp('2024-11-03 00:00:00-0500', tz='US/Central'), 'end': Timestamp('2024-11-03 03:00:00-0600', tz='US/Central')}
```

I move the `test_save_to.py` to `source_specific/` and fixturize it with VCR, since it makes external dependency calls.
I also ignore the commit test on PRs, since everything is running on PRs and we don't need to run the commit level tests there.

#### Quality of Life
I add function type hints to each function that gets created or used in the PR. I also swap out some `log` logging for the more structured `logger`.


gridstatus/ercot.py
```diff
@@ -3351,3 +3353,56 @@ def _weather_zone_column_name_order(self):
...
+                "Interval Start",
+                "Interval End",
+                "RTDTimestamp",
+                "Interval Id",
+                "Interval Repeated Hour Flag",
+                "Location",
+                "Location Type",
+                "LMP",
+            ]
+        ]
```
WillKoehrsen, line 3408: I think we can move this to a utility function (something like `_handle_indicative_lmp_by_settlement_point`) in this class then call that method here and in `ErcotAPI`.

Kladar: sounds good, done.

gridstatus/ercot.py
```diff
@@ -3351,3 +3353,56 @@ def _weather_zone_column_name_order(self):
...
+            "RepeatedHourFlag": "DSTFlag",
+            "IntervalId": "Interval Id",
+            "IntervalEnding": "Interval End",
+            "IntervalRepeatedHourFlag": "Interval Repeated Hour Flag",
+            "SettlementPoint": "Location",
+            "SettlementPointType": "Location Type",
+            "LMP": "LMP",
+        }
+        df.rename(columns=columns_to_rename, inplace=True)
+        df["RTDTimestamp"] = pd.to_datetime(df["RTDTimestamp"]).dt.tz_localize(
```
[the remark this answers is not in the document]
Kladar: Got it. I think I've implemented this logic correctly, though I kind of copy it into the `_handle_indicative_lmp_by_settlement_point` rather than try to use/refactor `parse_doc` for now. 

Still running some tests, but the fall back hours seems to be working in the TestErcotAPI test, and the spring forward date (full day of 2024-03-10) seems to work when I pull data with the getter method. 

But trying to have the test just test the hours from 1am to 4am seems to fail a bit still.

gridstatus/ercot.py
```diff
@@ -3351,3 +3353,81 @@ def _weather_zone_column_name_order(self):
...
+
+        df["Interval Start"] = df["Interval End"] - pd.Timedelta(minutes=5)
+        df = df.sort_values("Interval Start").reset_index(drop=True)
+        return df[
+            [
+                "Interval Start",
+                "Interval End",
+                "RTDTimestamp",
+                "Interval Id",
+                "Interval Repeated Hour Flag",
```
kmax12, line 3428: im fine if we want to return this value here, but i dont think we should store `"Interval Repeated Hour Flag"` downstream in our database. we should be already encoding this data the the timezone of the the timestamp

gridstatus/ercot.py
```diff
@@ -3351,3 +3353,81 @@ def _weather_zone_column_name_order(self):
...
+        ambiguous = ambiguous_based_on_dstflag(df)
+
+        df["RTDTimestamp"] = pd.to_datetime(df["RTDTimestamp"]).dt.tz_localize(
+            self.default_timezone,
+            ambiguous=ambiguous,
+            nonexistent="NaT",
+        )
+        df["Interval End"] = pd.to_datetime(df["Interval End"]).dt.tz_localize(
+            self.default_timezone,
+            ambiguous=ambiguous,
```
kmax12, line 3416: is using `ambiguous=ambiguous` correct here? I'd think you need to base it off `"Interval Repeated Hour Flag"`

gridstatus/ercot.py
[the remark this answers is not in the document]
kmax12: is this exactly copy and pasted? if so, i'd this we can figure out how to refactor safely. im okay if we want to leave it as a todo, but then let's add the todo comment

gridstatus/ercot.py
Kladar: good call - I need to investigate how that column actually operates a bit (whether it uses boolean or "Y/N") but directionally separating out those flags and their uses first

Kladar:
Converting to draft until ready for review again.

Kladar:
Took me a while to figure it out, but I was actually running into a `pd.concat` error in the `@support_date_range` decorator when `frequency=None` on the spring forward DST day (and this seems unrelated to the DSTFlag handling that I thought was causing the error). 

With any other dates the data getting and transformation would finish fine, but on 2024-03-10, the data getting would proceed as normal but then error out at the final (decorator-based) concatenation. With `frequency=None`
CleanShot 2024-12-16 at 12 02 27@2x

@WillKoehrsen @kmax12 have either of you seen this issue before?

Kladar:
Okay, I believe this is working (handling DST and non existent times correctly, passing tests). 

I will open a new, separate PR for sending multiple `docId`s to the ERCOT API download endpoint

.github/workflows/tests-commit.yaml
```diff
@@ -1,8 +1,12 @@
 name: Commit Tests
-on: # Run on all commit pushes to all feature branches.
+
```
kmax12, line 2: didn't review what these changes do. will leave to @WillKoehrsen to review

gridstatus/ercot.py
```diff
@@ -3351,3 +3353,63 @@ def _weather_zone_column_name_order(self):
...
+            nonexistent="shift_forward",
+        )
+
+        df["Interval Start"] = df["Interval End"] - pd.Timedelta(minutes=5)
+        df = df.sort_values("Interval Start").reset_index(drop=True)
+        return df[
+            [
+                "Interval Start",
+                "Interval End",
+                "RTDTimestamp",
```
kmax12, line 3409: we should name this `RTD Timestamp` since it is two words

Kladar: done

gridstatus/tests/source_specific/test_ercot_api.py
```diff
@@ -1436,3 +1436,40 @@ def _endpoints_map_check(self, endpoint_dict: dict) -> list[str]:
...
+    def test_get_indicative_lmp_by_settlement_point(self, date, end):
+        with api_vcr.use_cassette(
+            f"test_get_indicative_lmp_historical_{date}_{end}.yaml",
+        ):
+            df = self.iso.get_indicative_lmp_by_settlement_point(date, end)
+
+            assert df.columns.tolist() == [
+                "Interval Start",
+                "Interval End",
+                "RTDTimestamp",
```
kmax12, line 1457: need to add space here and in `test_ercot.py`

image

gridstatus/ercot.py
```diff
@@ -3351,3 +3353,64 @@ def _weather_zone_column_name_order(self):
...
+        df["Interval End"] = pd.to_datetime(df["Interval End"]).dt.tz_localize(
+            self.default_timezone,
+            ambiguous=df["IntervalRepeatedHourFlag"] == "N",
+            nonexistent="shift_forward",
+        )
+
+        df["Interval Start"] = df["Interval End"] - pd.Timedelta(minutes=5)
+        df = df.sort_values("Interval Start").reset_index(drop=True)
+        return df[
+            [
```
kmax12, line 3407: i'm thinking about the column order. i think this makes more sense and also matches the ercot source 

```
 [
                "RTD Timestamp",
                "Interval Start",
                "Interval End",
                "Interval Id",
                "Location",
                "Location Type",
                "LMP",
            ]
```

gridstatus/ercot.py
```diff
@@ -3351,3 +3353,64 @@ def _weather_zone_column_name_order(self):
...
+            published_before=end,
+            published_after=date,
+            verbose=verbose,
+        )
+        df = self.read_docs(docs, parse=False, verbose=verbose)
+        return self._handle_indicative_lmp_by_settlement_point(df)
+
+    def _handle_indicative_lmp_by_settlement_point(self, df: pd.DataFrame):
+        columns_to_rename = {
+            "IntervalId": "Interval Id",
```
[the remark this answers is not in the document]
Kladar: Yeah I think it is redundant with the combination of `RTD Timestamp` and `Interval End` as it says what interval out are they forecasting. 

CleanShot 2024-12-17 at 09 56 43@2x

WillKoehrsen (approved):

PR #505, opened 2024-12-17
feat(ENG 1028): Bulk ERCOT API Downloads

Kladar:
## Summary
[DRAFT, waiting for https://github.com/gridstatus/gridstatus/pull/504 which I will merge in here first] Turns out you can have up to 1000 `docId`s in the download endpoint request (but it does seem to want it to be a `POST` request)

### Details
Adds bulk download functionality to the `get_historical_data()` method stack. I don't assume this works for every dataset and affects the lower level `make_api_request()` which a lot of methods use, but we can evolve it to other methods if we want.


kmax12 (commented):
some comments. however, im going to put up a PR that refactors the approach.

heads up, i also changed the base branch of the PR to make the diff easier to read

Kladar:
Closing for #506

PR #506, opened 2024-12-17
Implement another approach to ERCOT API bulk document downloading

kmax12:
Took another stab at implementing the bulk data download.

Per our discussion this morning, I wanted to separate the downloading of the data from the processing of it. In the future, it should be a little bit easier to tease these piece apart. for example, if we wanted to write these data to S3 before processing it. 

In my opinion it also made this code more maintainable since the parsing logic is now the same regardless of it we are bulk downloading or individually downloading. 

My biggest concern is this approach takes more memory since we are storing both file bytes and parsed dataframes in memory. Before, we only stored the parsed dataframe. I'm not sure this matters, but it might on render where we try to run jobs on as small boxes as possible.

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -3380,8 +3380,8 @@ class Ercot(ISOBase):
     def _handle_indicative_lmp_by_settlement_point(self, df: pd.DataFrame):
         columns_to_rename = {
             "IntervalId": "Interval Id",
-            "RTDTimestamp": "RTD Timestamp",
             "IntervalEnding": "Interval End",
+            "RTDTimestamp": "RTD Timestamp",
             "SettlementPoint": "Location",
             "SettlementPointType": "Location Type",
             "LMP": "LMP",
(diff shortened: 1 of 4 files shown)


Kladar:
I'm getting a test failure on the fallback day stemming from a docId order issue

To reproduce:
```
df = ercotapi.get_indicative_lmp_by_settlement_point(
    date="2024-11-03 00:00:00",
    end="2024-11-03 04:00:00",
)
```

CleanShot 2024-12-17 at 07 53 54@2x

WillKoehrsen:
> @WillKoehrsen do you want to test this branch with the wind/solar backfill you are doing?

Yep, it works running locally. As you mentioned, we might have memory issues on Render but it's good to have this option.

Kladar (approved):
Nice, working for me now

kmax12:
I feel good about merging this to your branch, but I still want to work through the failing ercot tests. at least one is preexisting, but not sure about all of them

PR #507, opened 2024-12-17
Fix ERCOT Tests

WillKoehrsen:
## Summary

- Fix failing ERCOT tests

### DIFF
--- a/gridstatus/tests/source_specific/test_ercot.py
+++ b/gridstatus/tests/source_specific/test_ercot.py
@@ -1584,10 +1584,8 @@ class TestErcot(BaseTestISO):
             df["Interval End"] - df["Interval Start"] == pd.Timedelta(hours=1)
         ).all()
 
-    # For some reason, the datasets (confirmed in the raw data) start with
-    # HourEnding 02:00 and end with HourEnding 01:00
-    temperature_forecast_start_offset = -pd.DateOffset(days=3, hours=-1)
-    temperature_forecast_end_offset = pd.DateOffset(days=9, hours=1)
+    temperature_forecast_start_offset = -pd.DateOffset(days=3)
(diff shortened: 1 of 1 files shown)


kmax12:
i'm only seeing one test failing (`test_get_system_wide_actual_load_latest`), but when i run it locally, it passes, so maybe it's a flaky test?

gridstatus/ercot_api/ercot_api.py
```diff
@@ -909,29 +909,14 @@ def get_lmp_by_bus_dam(self, date, end=None, verbose=False):
...
-
-            data = self.hit_ercot_api(
-                endpoint=DAM_LMP_ENDPOINT,
-                page_size=DEFAULT_PAGE_SIZE,
-                verbose=verbose,
-                **api_params,
-            )
+        # For historical data, we need to subtract a day because we filter by
+        # posted date and this is day-ahead data
+        data = self.get_historical_data(
```
WillKoehrsen, line 914: I think it may make sense to always use the historical data method but I need to check if the data there is delayed.

kmax12: if we are using the api, we can always use `get_historical_data`. i think if we need low latency, we can use emil

WillKoehrsen:
> i'm only seeing one test failing (`test_get_system_wide_actual_load_latest`), but when i run it locally, it passes, so maybe it's a flaky test?

Yeah, some of the tests are flaky even when using 5 reruns.

kmax12 (approved):
Seems like this is a step in the right direction. we can work on further improving in another PR

WillKoehrsen:
There are two Ercot tests that failed on the latest run. However, these tests pass when run locally so I think they are flaky or it's a rate limit problem. 

https://github.com/gridstatus/gridstatus/actions/runs/12437966401/job/34728972803?pr=507

PR #508, opened 2024-12-18
Change Bulk Download Default to True

WillKoehrsen:
## Summary

- Set `bulk_download` default to `True` for `ErcotAPI`


kmax12 (approved):
let's keep an eye next few days on making sure this doesn't break anything

PR #510, opened 2024-12-18
Remove date parameter

kmax12:
I noticed `get_indicative_lmp_by_settlement_point` wasn't updating with new data. The problem was we were passing an extra data parameter to `_get_documents` that was filtering data more than we intended. just `published_before` and `published_after` are enoguh


WillKoehrsen (approved):

Kladar:
Catching up a bit, thanks for handling this.

PR #522, opened 2025-01-16
feat(ENG 1120): Expand CAISO Load Forecast

Kladar:
## Summary
Adds the other time horizons to the CAISO Load Forecast method. Also takes a one-time pass at updating `caiso.py` and `test_caiso.py`. See comment call-outs to aid in review.

### "New Source Review" Approach
I'm taking a bit of a "New Source Review" (NSR) approach to the codebase. NSR is the part of the EPA Clean Air Act that states that if an existing power plant that was previously grandfathered in to a certain level of EPA emissions standards has any additions or upgrades, the plant loses its grandfathered status and must fully upgrade to comply with the current standard (even if the additions or upgrades are otherwise unrelated to the standard). Basically, "if you're touching it at all, you gotta upgrade everything that's out of date".

Applying that logic to this codebase, the time when a new dataset is added to an existing source (e.g. `caiso.py`), I upgrade that existing source's typehints and tests (and anything else that could be fairly straightforwardly improved). The nice thing is that usually this only takes an extra ~day, and only needs to happen once per source. I hadn't really touched CAISO before this, so it gets the full upgrade this time around (and future additions are a bit easier as they do not require this upgrade). 

### CAISO Load Forecast
Expands the `OASIS_DATASET_CONFIG` to allow for other `market_run_id` inputs, such that different forecasts can be retrieved. Where there was previously only the "DAM" forecast, now there are "DAM", "2DA", "7DA", "ACTUAL", and "RTM" (which is all of them). 

Since the `publish_time` is added by gridstatus (it doesn't come as data in the dataset from the ISO), this function has been updated as well. I have made a few assumptions on publish times, but I believe this is consistent with how the ISO publishes the data. 

The RTM data contains both 5-minute and 15-minute data, so these are further separated into separate datasets, and the user-provided `RTM5` or `RTM15` is needed to get one or the other. 

There is a note about the 5minute data publish schedule, from CAISO:

> “RTM 5-Minute Load Forecast is posted every five minutes, for the next 11 intervals. The postings occur every 5-minutes for a rolling 11 interval period.”

This would seem to indicate that there are multiple horizons in a single 5-minute forecast publishing, and as new forecasts are published the intervals are updated in-place by CAISO. For historical data this is simply treated as a collated 5-minute horizon, since the most recent publish of an interval will be its 5-minute horizon, and thus all historical data is of a 5-minute horizon. How this looks for a real-time ETL process for 5-minute actions (e.g. bidding) or watching the 5-minute forecast evolve over the course of the current hour is a bit unclear, however. Here's an example of some 5-minute raw data (filtered by `LABEL` and zone) and it looks how you would expect it would. 

```
INTERVALSTARTTIME_GMT,INTERVALENDTIME_GMT,LOAD_TYPE,OPR_DT,OPR_HR,OPR_INTERVAL,MARKET_RUN_ID,TAC_AREA_NAME,LABEL,XML_DATA_ITEM,POS,MW,EXECUTION_TYPE,GROUP
2025-01-15T08:00:00-00:00,2025-01-15T08:05:00-00:00,0,2025-01-15,1,1,RTM,CA ISO-TAC,RTM 5Min Load Forecast,SYS_FCST_5MIN_MW,5.0,21914.02,RTD,24
2025-01-15T08:05:00-00:00,2025-01-15T08:10:00-00:00,0,2025-01-15,1,2,RTM,CA ISO-TAC,RTM 5Min Load Forecast,SYS_FCST_5MIN_MW,5.0,21880.5,RTD,24
2025-01-15T08:10:00-00:00,2025-01-15T08:15:00-00:00,0,2025-01-15,1,3,RTM,CA ISO-TAC,RTM 5Min Load Forecast,SYS_FCST_5MIN_MW,5.0,21848.31,RTD,24
2025-01-15T08:15:00-00:00,2025-01-15T08:20:00-00:00,0,2025-01-15,1,4,RTM,CA ISO-TAC,RTM 5Min Load Forecast,SYS_FCST_5MIN_MW,5.0,21830.46,RTD,24
2025-01-15T08:20:00-00:00,2025-01-15T08:25:00-00:00,0,2025-01-15,1,5,RTM,CA ISO-TAC,RTM 5Min Load Forecast,SYS_FCST_5MIN_MW,5.0,21794.98,RTD,24
2025-01-15T08:25:00-00:00,2025-01-15T08:30:00-00:00,0,2025-01-15,1,6,RTM,CA ISO-TAC,RTM 5Min Load Forecast,SYS_FCST_5MIN_MW,5.0,21731.41,RTD,24
```

You can run a simple script to see the data yourself.
```
from gridstatus import CAISO

caiso = CAISO()

df = caiso.get_load_forecast(date="20250115", forecast_vintage="2DA")
df2 = caiso.get_load_forecast(date="20250115", forecast_vintage="7DA")
df3 = caiso.get_load_forecast(date="20250115", forecast_vintage="DAM")
df4 = caiso.get_load_forecast(date="20250115", forecast_vintage="ACTUAL")
df5 = caiso.get_load_forecast(date="20250115", forecast_vintage="RTM5")
df6 = caiso.get_load_forecast(date="20250115", forecast_vintage="RTM15")

print(df)
print(df2)
print(df3)
print(df4)
print(df5)
print(df6)
```

### caiso.py quality of life
Adds function typehints to all methods, and a bit of code reorganization in the file to be easier to understand and navigate.

### test_caiso.py update
VCR-ifies all the tests in the CAISO tests (except some of the `super()` LMP tests), as well as breaks some of them into separate tests where there are multiple data calls in a single test. It also fixes a few failing tests. Speed up over pure integration style testing is observed:

**Current on `main` (8 threads, >2 min):**
CleanShot 2025-01-15 at 17 05 25@2x

**This PR (8 threads, ~1min):**
CleanShot 2025-01-15 at 17 10 57@2x

(NOTE: The one integration-style LMP test still failing is failing on `main`)

Reminder that fast, high-quality testing and things like static type checking are probably a pre-requisite to any large refactor of major parts or all of this repository. And most of that only has to be implemented once up front.


gridstatus/caiso.py
```diff
@@ -30,8 +31,245 @@
 DAY_AHEAD_MARKET_MARKET_RUN_ID = "DAM"
 REAL_TIME_DISPATCH_MARKET_RUN_ID = "RTD"
 
+OASIS_DATASET_CONFIG = {
```
Kladar, line 34: Moved this to the top of the page since it's a constant.

gridstatus/caiso.py
```diff
@@ -30,8 +31,245 @@
...
+        "params": {
+            "groupid": ["DAM_ENE_SCH_BY_TIE_GRP"],
+            "market_run_id": DAY_AHEAD_MARKET_MARKET_RUN_ID,
+        },
+    },
+}
 
-def determine_lmp_frequency(args):
+
+def determine_lmp_frequency(args: dict) -> str:
```
Kladar, line 272: Moved the bare functions next, such that they aren't hidden below the `CAISO` class and ambiguous on whether they belong or not.

gridstatus/caiso.py
```diff
@@ -111,19 +433,268 @@ def get_status(self, date="latest", verbose=False) -> str:
         else:
             raise NotSupported()
 
+    def list_oasis_datasets(self, dataset: str | None = None):
```
Kladar, line 436: Helper and meta functions to the top of the class, above the interval data-getting methods.

gridstatus/caiso.py
```diff
@@ -111,19 +433,268 @@ def get_status(self, date="latest", verbose=False) -> str:
...
+            df.insert(0, "Time", df["Interval Start"])
+
+        # avoid rate limiting
+        time.sleep(sleep)
+
+        return df
+
     @support_date_range(frequency="DAY_START")
-    def get_fuel_mix(self, date, start=None, end=None, verbose=False):
+    def get_fuel_mix(
```
Kladar, line 680: Just updating type-hints for existing class methods, no logic changes.

gridstatus/caiso.py
```diff
@@ -178,18 +762,40 @@ def _get_historical_load(self, date, verbose=False):
         return df
 
     @support_date_range(frequency="31D")
-    def get_load_forecast(self, date, end=None, sleep=4, verbose=False):
-        """Returns load forecast for a previous date in 1 hour intervals
+    def get_load_forecast(
```
Kladar, line 765: Actually new functionality for load forecasts.

gridstatus/caiso.py
```diff
@@ -342,29 +971,73 @@ def _add_forecast_publish_time(
...
-        data["Publish Time"] = np.where(
-            data["Interval Start"].dt.date > future_forecasts_publish_time.date(),
-            future_forecasts_publish_time,
-            data["Interval Start"].apply(
-                lambda x: x.floor("D").replace(hour=hour_offset, minute=minute_offset),
-            )
-            # DAM is for the next day
-            - pd.Timedelta(days=1),
-        )
+        match forecast_horizon:
```
Kladar, line 974: new functionality here too, to support the expanded load forecasts

gridstatus/tests/vcr_utils.py
```diff
@@ -7,7 +7,7 @@
 
 # NOTE(Kladar): Set VCR_RECORD_MODE to "all" to update the fixtures as an integration test,
 # say on a weekly or monthly job.
-RECORD_MODE = os.getenv("VCR_RECORD_MODE", "once")
+RECORD_MODE = os.getenv("VCR_RECORD_MODE", "new_episodes")
```
Kladar, line 10: makes the "latest" and "today" tests easier to do, since they have a conflict on the "once" record mode. 

essentially, "new_episodes" will look at the existing fixture for that "today" or "latest" run, and if it sees it's different than the call it would make now, it makes a new call to the external source and updates the fixture (rather than throw an error). It's very close to the "once" mode but has that advantage.

gridstatus/caiso.py
```diff
@@ -212,10 +819,19 @@ def get_load_forecast(self, date, end=None, sleep=4, verbose=False):
         df = self._add_forecast_publish_time(
```
kmax12, line 819: i haven't looked at your code yet, but wanted to surface any issue I think we have in our current logic for parsing publish time for the DAM forecast: https://github.com/gridstatus/gs-etl/pull/244#discussion_r1792573092

it's very possible your changes fix it

Kladar: I left the existing logic and slightly expanded it to handle the other forecast cases, but I will take a look

gridstatus/caiso.py
```diff
@@ -198,38 +804,51 @@ def get_load_forecast(self, date, end=None, sleep=4, verbose=False):
...
         )
 
         df = df.rename(
-            columns={"MW": "Load Forecast"},
+            columns={"MW": "Load Forecast", "TAC_AREA_NAME": "TAC Area Name"},
         )
 
-        # returns many areas, we only want one overall iso
-        df = df[df["TAC_AREA_NAME"] == "CA ISO-TAC"]
-
```
Kladar, line 209: This is a schema change (removes the single-region filter and adds a new column for `TAC Area Name`), which could hypothetically break certain downstream systems. 

Will also expand the data be ~35x. There are 34-35 TAC Areas per forecast horizon.

gridstatus/caiso.py
WillKoehrsen: Since these util functions are only used in this file, I think they should be prefixed with `_`.

gridstatus/caiso.py
```diff
@@ -178,18 +762,40 @@ def _get_historical_load(self, date, verbose=False):
         return df
 
     @support_date_range(frequency="31D")
-    def get_load_forecast(self, date, end=None, sleep=4, verbose=False):
-        """Returns load forecast for a previous date in 1 hour intervals
+    def get_load_forecast(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        forecast_vintage: Literal[
```
WillKoehrsen, line 769: CAISO calls these options "Market/Process". Is Market a better descriptor than vintage?

Screenshot 2025-01-19 at 20 17 11

gridstatus/caiso.py
```diff
@@ -342,29 +969,73 @@ def _add_forecast_publish_time(
...
+                        minute=minute_offset,
+                    )
+                    - pd.Timedelta(days=7),
+                )
+            case "RTM5":
+                data["Publish Time"] = data["Interval Start"] - pd.Timedelta(minutes=5)
+            case "RTM15":
+                data["Publish Time"] = data["Interval Start"] - pd.Timedelta(minutes=15)
+            case "ACTUAL":
+                data["Publish Time"] = data["Interval Start"]
```
WillKoehrsen, line 1020: Is "ACTUAL" the real recorded values? In that case, I don't think we need a publish time.

gridstatus/caiso.py
```diff
@@ -198,38 +804,51 @@ def get_load_forecast(self, date, end=None, sleep=4, verbose=False):
...
-        # returns many areas, we only want one overall iso
-        df = df[df["TAC_AREA_NAME"] == "CA ISO-TAC"]
-
         df = df.sort_values("Time")
 
         df = self._add_forecast_publish_time(
             df,
             current_time,
+            forecast_vintage=forecast_vintage,
+            # TODO(kladar): verify exactly  when other forecast types are published
```
WillKoehrsen, line 820: Here is the publication schedule for demand load forecasts. The daily forecasts are supposed to be published at 9:10 am. 

Screenshot 2025-01-19 at 20 29 38

gridstatus/caiso.py
WillKoehrsen: If we are going to have a publish time, it should be +3 minutes after the interval.

Screenshot 2025-01-19 at 20 32 35

gridstatus/caiso.py
```diff
@@ -342,29 +969,73 @@ def _add_forecast_publish_time(
...
+            case "7DA":
+                data["Publish Time"] = data["Interval Start"].apply(
+                    lambda x: x.floor("D").replace(
+                        hour=hour_offset,
+                        minute=minute_offset,
+                    )
+                    - pd.Timedelta(days=7),
+                )
+            case "RTM5":
+                data["Publish Time"] = data["Interval Start"] - pd.Timedelta(minutes=5)
```
WillKoehrsen, line 1016: Should be -2.5 minutes according to the schedule.

gridstatus/caiso.py
```diff
@@ -342,29 +969,73 @@ def _add_forecast_publish_time(
...
+            case "RTM15":
+                data["Publish Time"] = data["Interval Start"] - pd.Timedelta(minutes=15)
```
WillKoehrsen, line 1018: Should be -22.5 minutes.

gridstatus/caiso.py
```diff
@@ -342,29 +969,73 @@ def _add_forecast_publish_time(
...
-            data["Interval Start"].apply(
-                lambda x: x.floor("D").replace(hour=hour_offset, minute=minute_offset),
-            )
-            # DAM is for the next day
-            - pd.Timedelta(days=1),
-        )
+        match forecast_vintage:
+            case "DAM":
+                # If the current time is after the publish time, then future forecasts were
+                # published today. Otherwise, they were published in a previous day.
```
WillKoehrsen, line 975: I think this may be the incorrect logic @kmax12 refers to [here](https://github.com/gridstatus/gridstatus/pull/522/files#r1917620296).

Kladar: yep it is

gridstatus/caiso.py
Kladar: Yeah I guess it's the question of passing through what the market operator calls it vs what it represents semantically. Here they are smuggling in a couple ideas: market ("Real Time" or "Day Ahead" are typically the only options), and the "process" that creates a given forecast. I tend to fall on the side of starting to be opinionated about what "Grid Status" thinks this field is vs what the market operator calls it (with proper documentation to understand they are the same). Every market operator might call it something different, but we know what it is: "when a forecast is published" is it's vintage.

Kladar: I'm actually going to break these up into different methods rather than try to bundle them into the same get_load_forecast, to stick with how we are doing it in other datasets.

gridstatus/caiso.py
Kladar: done

Kladar:
@WillKoehrsen I'm going to make final changes to:
1. Publish time for each dataset method (RTM and Actual are good; DAM, 7DA, 2DA are close).
2. Tests for each of those methods

And then we should be good. I'll comment back here when those are in.

Kladar:
@WillKoehrsen Okay, logic and tests are working for all new CAISO Load Forecast methods. I left in a `get_load_forecast()` method because the `base_test_iso` wants to use one and the test harness has some slightly different column requirements than our current methods. [See the ticket](https://www.notion.so/gridstatus/Fix-CAISO-DAM-Load-Forecast-Publish-Times-121e835f42aa8074906ed05fa20dd92d?pvs=4) for more details on the CAISO Load Forecast Publish Time issue (which I believe is now solved).

gridstatus/caiso.py
```diff
@@ -161,92 +734,394 @@ def _get_historical_fuel_mix(self, date, verbose=False):
...
+        unique_dates = sorted(df["date"].unique())
+
+        for forecast_date in unique_dates:
+            publish_time = (
+                pd.Timestamp(forecast_date, tz=self.default_timezone)
+                - pd.Timedelta(days=day_offset)
+            ).replace(
+                hour=9,
+                minute=10,
+            )
```
WillKoehrsen, line 1050: Can we add a comment about why this is 09:10 am? It's from http://oasis.caiso.com/mrioasis/logon.do > Atlas Reference > Publications > OASIS Publications Schedule.

gridstatus/caiso.py
```diff
@@ -161,92 +734,394 @@ def _get_historical_fuel_mix(self, date, verbose=False):
...
+            df.loc[df["date"] == forecast_date, "Publish Time"] = publish_time
+
+        df = df.drop(columns=["date"])
+        return df
+
+    @support_date_range(frequency="31D")
+    def get_load_hourly(
```
WillKoehrsen, line 1057: Are we going to add a new load dataset because our existing one doesn't have the `TAC Area Name`?

Kladar: I believe the [existing load dataset](https://www.gridstatus.io/datasets/caiso_load) is 5 minute (this would be hourly).

gridstatus/caiso.py
Kladar: done

gridstatus/caiso.py
```diff
@@ -161,92 +734,396 @@ def _get_historical_fuel_mix(self, date, verbose=False):
...
-    def get_solar_and_wind_forecast_dam(self, date, end=None, verbose=False):
-        """Return wind and solar forecast in hourly intervals
-
-        Data at: http://oasis.caiso.com/mrioasis/logon.do  at System Demand >
-        Wind and Solar Forecast
-        """
-        if date == "latest":
-            return self.get_solar_and_wind_forecast_dam("today", verbose=verbose)
+    @support_date_range(frequency="31D")
+    def get_load_forecast_15_min(
```
WillKoehrsen, line 832: I'm running this function locally and getting some errors:

* Using `CAISO().get_load_forecast_15_min('today')` the error is `ValueError: Parameter market_run_id not supported for dataset demand_forecast`
* Using any combination of start and end dates, for example, `CAISO().get_load_forecast_15_min('2025-01-01', '2025-01-21')` the error is `ValueError: No objects to concatenate`

gridstatus/caiso.py
WillKoehrsen: Yeah, you're right!

gridstatus/caiso.py
Kladar: yep just ran into that as well, I think 1) I need to support "today" and "latest" and 2) is the market_run_id got flipped somewhere. one second

WillKoehrsen: Nice, seems to be working now. I think we are missing a test for `get_load_forecast_15_min` though.

Kladar: good catch - added.

WillKoehrsen (approved):
👍🏽

kmax12:
nice job breaking this up and making the new datasets functions. very clear and maintainable!

PR #529, opened 2025-01-23
feat(ENG 1140): CAISO GHG component of LMP

Kladar:
## Summary
Adds the `GHG` component of LMP as a column to CAISO `get_lmp()` output dataframes for all markets.


gridstatus/tests/base_test_iso.py
```diff
@@ -404,6 +404,7 @@ def _check_is_datetime_type(self, series):
         "Energy",
         "Congestion",
         "Loss",
+        "GHG",
```
WillKoehrsen, line 407: Since this is in the base test class, doesn't it affect tests for all ISOs, not just CAISO?

Kladar: Ah yeah, but I was thinking the logic where it is used was checking that the df columns were a subset of these columns (meaning its fine if the df columns don't have every LMP column). Now realizing it's the other way around: it checks for all the lmp columns being present in the df. 

```
assert set(
            self.lmp_cols,
        ).issubset(df.columns)
```

eventually there is probably a logical refactor for this base test harness

WillKoehrsen: Can't you override `lmp_cols` in `TestCAISO` and run the same test? That's what we do for PJM LMPs.

https://github.com/gridstatus/gridstatus/blob/36998865719012de1b7b805fed69407f06de4cff/gridstatus/tests/source_specific/test_pjm.py#L77-L91

Kladar: Ah yeah that will work...and makes me want to refactor the base_test stuff even more lol.

Kladar:
I believe this is ready. 

- RT5Min grabs the GHG which is nonzero. 
- RT15Min grabs the GHG which is nonzero. 
- Even though the DAM seems to be always 0.0, I'm still grabbing it as the API does output that value when querying nodes other than `ALL`, and it could change to be nonzero in the future IMO. Plus the database is set up to receive it.

WillKoehrsen (approved):

PR #532, opened 2025-01-26
Query current day data from real-time hourly market in ISO-NE

### LINKED ISSUE #515: ISO-NE Today RT LMP hourly
ISO-NE publishes hourly RTLMP CSV’s for current day. I don’t think the comment on the code where it says they provide only rolling 3 hour is true anymore. I checked it by modifying the line where it checks if date < today to <= today and it works perfectly. I also checked the URL and the CSV is being published. Can someone check and change the code to work for today? Thanks.

brianstrauch:
## Summary
Closes #515. Allow data from the current day to be queried from the real-time hourly market in ISO-NE, since this is now supported.

Also fixed the runtime error if a future date is passed.

### Details
Verified with the following script:
```
from gridstatus import ISONE, Markets
import pandas as pd

print(ISONE().get_lmp(
    date=pd.Timestamp.now(tz="US/EASTERN").date(),
    market=Markets.REAL_TIME_HOURLY,
))
```

### DIFF
--- a/gridstatus/isone.py
+++ b/gridstatus/isone.py
@@ -479,7 +479,9 @@ class ISONE(ISOBase):
 
         elif market == Markets.REAL_TIME_HOURLY:
             if date.date() > now.date():
-                raise RuntimeError(f"date {date.date()} is in the future and cannot be used to query real-time data")
+                raise RuntimeError(
+                    f"date {date.date()} is in the future and cannot be used to query real-time data"
+                )
 
             url = f"https://www.iso-ne.com/static-transform/csv/histRpts/rt-lmp/lmp_rt_prelim_{date_str}.csv"  # noqa
             data = _make_request(
--- a/gridstatus/isone.py
+++ b/gridstatus/isone.py
@@ -478,6 +478,9 @@ class ISONE(ISOBase):
             data = data.rename(columns={"Local Time": "Interval Start"})
 
         elif market == Markets.REAL_TIME_HOURLY:
+            if date.date() > now.date():
+                raise RuntimeError(f"date {date.date()} is in the future and cannot be used to query real-time data")
+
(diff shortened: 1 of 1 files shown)


gridstatus/isone.py
```diff
@@ -478,19 +478,12 @@ def get_lmp(
...
-                    skiprows=[0, 1, 2, 3, 5],
-                    verbose=verbose,
-                )
-            else:
-                # iso only publishes rolling 3 hours of data for current
-                # day real time hourly. idk why
-                raise RuntimeError(
-                    "Today not supported for hourly lmp. Try latest",
-                )
+            url = f"https://www.iso-ne.com/static-transform/csv/histRpts/rt-lmp/lmp_rt_prelim_{date_str}.csv"  # noqa
```
brianstrauch, line 481:
```suggestion
            if date.date() > now.date():
                raise RuntimeError(f"date {date.date()} is in the future and cannot be used to query real-time data")

            url = f"https://www.iso-ne.com/static-transform/csv/histRpts/rt-lmp/lmp_rt_prelim_{date_str}.csv"  # noqa
```

WillKoehrsen:
@brianstrauch These changes look good and the isone tests pass when running locally. Can you run the linter (`make lint-fix`) and then update the PR? The lint check is failing.

gridstatus/isone.py
```diff
@@ -478,19 +478,15 @@ def get_lmp(
...
-                    verbose=verbose,
-                )
-            else:
-                # iso only publishes rolling 3 hours of data for current
-                # day real time hourly. idk why
-                raise RuntimeError(
-                    "Today not supported for hourly lmp. Try latest",
-                )
+            if date.date() > now.date():
+                raise RuntimeError(f"date {date.date()} is in the future and cannot be used to query real-time data")
```
brianstrauch, line 482:
```suggestion
                raise RuntimeError(
                    f"date {date.date()} is in the future and cannot be used to query real-time data"
                )
```

brianstrauch:
Updated! @WillKoehrsen

PR #535, opened 2025-01-27
Fix CAISO Solar and Wind Forecast DAM Publish Time

WillKoehrsen:
## Summary

- Fixes the publish time for CAISO `get_solar_and_wind_forecast_dam`

### DIFF
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -1115,8 +1115,6 @@ class CAISO(ISOBase):
         if date == "latest":
             return self.get_solar_and_wind_forecast_dam("today", verbose=verbose)
 
-        current_time = pd.Timestamp.now(tz=self.default_timezone)
-
         data = self.get_oasis_dataset(
             dataset="wind_and_solar_forecast",
             date=date,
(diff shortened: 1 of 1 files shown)


gridstatus/caiso.py
```diff
@@ -1168,13 +1158,24 @@ def _process_solar_and_wind_forecast_dam(
...
+        for forecast_date in unique_dates:
+            publish_time = (
+                pd.Timestamp(forecast_date, tz=self.default_timezone)
+                # One day because this is a day-ahead forecast
+                - pd.Timedelta(days=1)
+            ).replace(
+                hour=7,
+                minute=0,
+            )
+            df.loc[df["date"] == forecast_date, "Publish Time"] = publish_time
```
WillKoehrsen, line 1176: This is the same logic we now use for the load forecasts as implemented in #522.

Kladar:
Looks like a relevant CAISO test is maybe failing https://github.com/gridstatus/gridstatus/actions/runs/12980044579/job/36196572869?pr=535#step:7:913

```
assert df["Publish Time"].max() < self.local_now()
```
in the test it seems, I was able to reproduce locally for the same failure.

WillKoehrsen:
> Looks like a relevant CAISO test is maybe failing 

The test failure is related to how we calculate the publish time. Depending on when you fetch the data, the publish time may be in the future relative to the current time. This is why I originally implemented a convoluted method for calculating the publish time.

Kladar:
> > Looks like a relevant CAISO test is maybe failing
> 
> The test failure is related to how we calculate the publish time. Depending on when you fetch the data, the publish time may be in the future relative to the current time. This is why I originally implemented a convoluted method for calculating the publish time.

I think we need to rewrite it then, such that it passes.

WillKoehrsen:
Okay, I made a mistake. I thought this dataset was a day-ahead forecast, so, to get the publish time, we only needed to subtract 1 day from the interval start. This dataset actually goes out seven days.

What confused me is this dataset has a market of "DAM" in OASIS.

Screenshot 2025-02-19 at 14 11 26

WillKoehrsen:
This PR should be ready for review once we decide how to handle the seven day solar and wind forecast.

PR #539, opened 2025-02-01
[ENG-1217] NYISO Interface Flows and Lake Erie Circulation

WillKoehrsen:
## Summary

- Adds NYISO interface limits and flows
- Adds NYISO Lake Erie Circulation Real Time and Day Ahead

### DIFF
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -317,7 +317,7 @@ class NYISO(ISOBase):
                 "Positive Limit MWH",
                 "Negative Limit MWH",
             ]
-        ].sort_values(["Interval Start", "Interface Name"])
+        ]
 
         return data
 
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -1,4 +1,4 @@
-from typing import BinaryIO
+from typing import BinaryIO, NamedTuple, Union
 
 import pandas as pd
 import requests
@@ -24,6 +24,35 @@ LOAD_FORECAST_DATASET = "isolf"
 INTERFACE_LIMITS_AND_FLOWS_DATASET = "ExternalLimitsFlows"
 LAKE_ERIE_CIRCULATION_REAL_TIME_DATASET = "eriecirculationrt"
 LAKE_ERIE_CIRCULATION_DAY_AHEAD_DATASET = "eriecirculationda"
+
+
+class DatasetInterval(NamedTuple):
+    time_type: str
+    interval_duration_minutes: Union[int, None]
+
+
+DATASET_INTERVAL_MAP = {
+    # (time_type, interval_duration_minutes)
+    # load
+    "pal": DatasetInterval("instantaneous", None),
+    # fuel mix
+    "rtfuelmix": DatasetInterval("instantaneous", None),
+    # load forecast
+    LOAD_FORECAST_DATASET: DatasetInterval("start", 60),
+    # dam lmp
+    "damlbmp": DatasetInterval("start", 60),
+    # rt lmp
+    "realtime": DatasetInterval("end", 5),
+    # real time events
+    "RealTimeEvents": DatasetInterval("instantaneous", None),
+    # btm solar
+    "btmactualforecast": DatasetInterval("start", 60),
+    # btm solar forecast
+    "btmdaforecast": DatasetInterval("start", 60),
+    INTERFACE_LIMITS_AND_FLOWS_DATASET: DatasetInterval("start", 5),
+    LAKE_ERIE_CIRCULATION_REAL_TIME_DATASET: DatasetInterval("start", 5),
+    LAKE_ERIE_CIRCULATION_DAY_AHEAD_DATASET: DatasetInterval("start", 60),
+}
 
 
 class NYISO(ISOBase):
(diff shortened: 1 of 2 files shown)


gridstatus/nyiso.py
```diff
@@ -1040,6 +1131,9 @@ def get_capacity_prices(self, date=None, verbose=False):
     "btmactualforecast": ("start", 60),
     # btm solar forecast
     "btmdaforecast": ("start", 60),
+    INTERFACE_LIMITS_AND_FLOWS_DATASET: ("start", 5),
+    LAKE_ERIE_CIRCULATION_REAL_TIME_DATASET: ("start", 5),
+    LAKE_ERIE_CIRCULATION_DAY_AHEAD_DATASET: ("start", 60),
 }
```
Kladar, line 1137: Neither here nor there, but in the spirit of ongoing typesafety and legibility discussion, this stands out to me as being particularly opaque as to what it is and does. "A top-level dictionary of tuples of mixed types created when nyiso.py is imported, at the bottom of the page". Potentially this 👇🏼, and moved to the top of the file alongside the Dataset name strings?

```suggestion
 class DatasetInterval(NamedTuple):
    time_type: str
    interval_duration_minutes: Union[int, None]

DATASET_INTERVAL_MAP: Dict[str, DatasetInterval] = {
    # load
    "pal": DatasetInterval("instantaneous", None),
    # fuel mix
    "rtfuelmix": DatasetInterval("instantaneous", None),
    # load forecast
    LOAD_FORECAST_DATASET: DatasetInterval("start", 60),
    # dam lmp
    "damlbmp": DatasetInterval("start", 60),
    # rt lmp
    "realtime": DatasetInterval("end", 5),
    # real time events
    "RealTimeEvents": DatasetInterval("instantaneous", None),
    # btm solar
    "btmactualforecast": DatasetInterval("start", 60),
    # btm solar forecast
    "btmdaforecast": DatasetInterval("start", 60),
    INTERFACE_LIMITS_AND_FLOWS_DATASET: ("start", 5),
    LAKE_ERIE_CIRCULATION_REAL_TIME_DATASET: ("start", 5),
    LAKE_ERIE_CIRCULATION_DAY_AHEAD_DATASET: ("start", 60),
}
```

And I could see combining this dataset info at the top and bottom of the page and creating a custom type, `NYISODataset` (rather than `DatesetInterval`) that has the `DATASET`s, `time_type`s, and `interval_duration_minutes`, etc in it. And this `map` is then a dictionary of those objects. Probably worth collecting these various nyiso.py memory objects into a single, visible place.

gridstatus/tests/source_specific/test_nyiso.py
```diff
@@ -545,6 +545,81 @@ def test_zonal_load_forecast_historical_date_range(self):
...
 
+    """get_interface_limits_and_flows_5_min"""
+
+    @pytest.mark.integration
+    @api_vcr.use_cassette(
+        "test_get_interface_limits_and_flows_5_min_historical_date_range.yaml",
+    )
+    def test_get_interface_limits_and_flows_5_min_historical_date_range(self):
+        start = self.local_start_of_today() - pd.DateOffset(days=10)
+        end = start + pd.Timedelta(days=1)
```
Kladar, line 556: I think we're still figuring out our conventions a bit, so open to doing it a different way. But I think when we implement VCR on a test I think we 

1. can remove the integration `mark`. 
2. `pytest.parameterize` the date args if possible (`start_of_local_today()` has proved mildly tricky/not possible and I do what you do here in that case, but maybe a more sustainable fix is available down the road)
3. use the context manager (vs the decorator) so we can dynamically name the fixture and do potentially some other things I'm forgetting at the moment. In this case in particular, it doesn't break like it used to with RECORD_MODE="once", but with RECORD_MODE="new_episodes" this is silently rerecording vs explicitly creating a new file (and I know I definitely prefer the latter).

```suggestion
    def test_get_interface_limits_and_flows_5_min_historical_date_range(self):
        start = self.local_start_of_today() - pd.DateOffset(days=10)
        end = start + pd.Timedelta(days=1)
        with nyiso_vcr.use_cassette(f"test_get_interface_limits_and_flows_5_min_historical_date_range_{date.strftime('%Y-%m-%d')}_{end.strftime('%Y-%m-%d')}"
        # test logic
```

gridstatus/nyiso.py
```diff
@@ -21,6 +21,9 @@
 GENERATOR = "generator"
 
 LOAD_FORECAST_DATASET = "isolf"
+INTERFACE_LIMITS_AND_FLOWS_DATASET = "ExternalLimitsFlows"
+LAKE_ERIE_CIRCULATION_REAL_TIME_DATASET = "eriecirculationrt"
+LAKE_ERIE_CIRCULATION_DAY_AHEAD_DATASET = "eriecirculationda"
```
Kladar, line 26: Collect with dataset interval properties (see below).

gridstatus/nyiso.py
Kladar, line 1: Add function typehints + docstrings to new methods.

Kladar (approved):
Logic works. With 
```
from gridstatus import nyiso
nyiso = nyiso.NYISO()
df = nyiso.get_interface_limits_and_flows_5_min(date="2025-02-02")
print(df)
```
I get data
CleanShot 2025-02-03 at 13 52 25

The rest is mostly just quality-of-life nits.

gridstatus/nyiso.py
WillKoehrsen: Good idea, updated.

gridstatus/tests/source_specific/test_nyiso.py
WillKoehrsen: Okay, I've implemented 1 and 3 but I don't think 2 is relevant in this case.

gridstatus/nyiso.py
WillKoehrsen: Added function typehints.

WillKoehrsen:
Waiting to merge pending discussions about treating these as intervals vs instantaneous data.

WillKoehrsen:
> Waiting to merge pending discussions about treating these as intervals vs instantaneous data.

We decided to treat Lake Erie circulation as instantaneous and interface flow as interval data.

PR #542, opened 2025-02-07
feat(ENG 911): PJM Area Control Error

Kladar:
## Summary
Adds the 15-second dataset `pjm_area_control_error`. Also triggers some "New Source Review" upgrades to PJM.

### Details
To get area control error data yourself:
```
pjm = PJM()
df = pjm.get_area_control_error(date="2025-02-01")
print(df)
```

#### New Source Review
Updates to all typehinted functions and tackles a good chunk of VCR testing, covering about 40% of current integration tests. Reduces local testing by ~85%, all things equal.

**Old**
CleanShot 2025-02-06 at 18 19 10

**New**
CleanShot 2025-02-06 at 18 56 27

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2650,3 +2650,30 @@ class PJM(ISOBase):
         )
 
         return df
+
+    @support_date_range(frequency=None)
+    def get_area_control_error(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieves the area control error data from:
+        https://dataminer2.pjm.com/feed/area_control_error/definition
+        """
+        if date == "latest":
+            date = "today"
+
+        df = self._get_pjm_json(
+            "area_control_error",
+            start=date,
+            end=end,
+            params={
+                "fields": "datetime_beginning_utc,ace_mw",
+            },
+            interval_duration_min=0.25,
+            verbose=verbose,
+        )
+
+        return df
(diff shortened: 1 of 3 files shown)


gridstatus/pjm.py
```diff
@@ -2577,3 +2650,35 @@ def get_day_ahead_demand_bids(self, date, end=None, verbose=False):
...
+            interval_duration_min=0.25,
+            verbose=verbose,
+        )
+        df = df.rename(
+            columns={
+                "ace_mw": "Area Control Error",
+            },
+        )
+
+        return df[["Interval Start", "Interval End", "Area Control Error"]]
```
kmax12, line 2684: we don't want this to be interval data. it should just return `Time` and `Area Control Error`

Kladar: ah got it. I believe this should work now by just returning the Time column instead, not touching any `_get_pjm_json()` logic, based on the way I'm reading what's happening in that function.

Kladar:
Weird so many PJM tests failing when all passing locally. Hmm

WillKoehrsen (approved):
Code and tests run well locally.

gridstatus/tests/source_specific/test_pjm.py
```diff
@@ -1915,83 +1980,120 @@ def test_get_transmission_constraints_day_ahead_hourly(self, date, end):
...
+            "today",
+            "latest",
+        ],
+    )
+    def test_get_area_control_error_today_or_latest(self, date: str):
+        """Test getting area control error data for today and latest"""
+        with pjm_vcr.use_cassette(f"test_get_area_control_error_{date}.yaml"):
+            df = self.iso.get_area_control_error(date)
+
+            assert self.iso.get_area_control_error("latest").equals(df)
```
WillKoehrsen, line 2099: When the `date` parameter is "latest", this isn't really testing anything. I think we should be checking the columns and also the time (min/max).

Kladar: improved

PR #543, opened 2025-02-14
IESO MCP and HOEP

WillKoehrsen:
## Summary

- Adds IESO price datasets: HOEP and MCP

### DIFF
--- a/gridstatus/decorators.py
+++ b/gridstatus/decorators.py
@@ -157,7 +157,7 @@ class support_date_range:
                     frequency = MonthBeginOffset()
 
                 elif frequency == "HOUR_START":
-                    frequency = pd.DateOffset(hours=1)
+                    frequency = HourBeginOffset()
 
                 elif frequency == "5_MIN":
                     frequency = FiveMinOffset()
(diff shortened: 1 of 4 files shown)


gridstatus/decorators.py
```diff
@@ -157,7 +157,7 @@ def wrapped_f(*args, **kwargs):
                     frequency = MonthBeginOffset()
 
                 elif frequency == "HOUR_START":
-                    frequency = pd.DateOffset(hours=1)
```
WillKoehrsen, line 160: The `"HOUR_START"` offset was not working as expected. When calling a decorated `"HOUR_START"` function with `'2025-02-15 15:05:00-0500'` and `'2025-02-15 16:02:00-0500'`, the function ran once with 

```python
`date='2025-02-15 15:05:00-0500'`
`end='2025-02-15 16:02:00-0500'`
``` 

This is what I'd expect from an offset called `"HOUR"`. However, since the offset is `"HOUR_START"`, the function should run twice with

```
`date='2025-02-15 15:05:00-0500'`
`end='2025-02-15 16:00:00-0500'`

`date='2025-02-15 16:00:00-0500'`
`end='2025-02-15 16:02:00-0500'`
```

Using the custom offset achieves this.

kmax12 (approved):
ran all the datasets locally and examined the data. looks good!

PR #544, opened 2025-02-20
[ENG-1303] EIA Generators

WillKoehrsen:
## Summary

- Adds EIA generators. `EIA().get_generators({date})` returns a dict with
  - `"operating"` generators
  - `"planned"` generators
  - `"retired"` generators


Kladar (commented):
Got this working after some certifi certificate run arounds. Still running into some machine certificate issues but that's on my end. 

```
# Create unverified context
ssl._create_default_https_context = ssl._create_unverified_context

eia = EIA()
dict_of_dfs = eia.get_power_plants(date="2024-01-01")

for key, df in dict_of_dfs.items():
    print(key)
    print(df.head())
    print("\n")
```
yields data. 
CleanShot 2025-02-24 at 08 29 34

Curious why we are giving a dict of dfs rather than concatenating dfs and adding the power plant status as a column?

WillKoehrsen:
> Curious why we are giving a dict of dfs rather than concatenating dfs and adding the power plant status as a column?

We're using a dict of dfs because the dfs have different columns.

kmax12 (approved):

PR #545, opened 2025-02-24
ERCOT API SCED 60 Day Disclosure

WillKoehrsen:
## Summary

- Adds ERCOT SCED 60 Day Disclosure reports to the ErcotAPI scraper to fetch historical data

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -3,6 +3,7 @@ import json
 import os
 import random
 import time
+from typing import Dict
 from zipfile import ZipFile
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


gridstatus/ercot_api/ercot_api.py
```diff
@@ -1191,6 +1196,67 @@ def get_60_day_dam_disclosure(
...
+        if end:
+            end = end + pd.DateOffset(days=60)
+        else:
+            end = date + pd.DateOffset(days=1)
+
+        df_list = []
+
+        # Get data once since both endpoints return the same zipfile
+        data_bytes = self.get_historical_data(
+            endpoint=SCED_60_DAY_SMNE_ENDPOINT,
```
kmax12, line 1238: am i understanding correctly that `SCED_60_DAY_SMNE_ENDPOINT` returns a zip file with all of the datasets we care about, not just the `sced_smne`?

gridstatus/ercot_api/ercot_api.py
```diff
@@ -1191,6 +1196,67 @@ def get_60_day_dam_disclosure(
         # Take the list of dictionaries and concat the dataframes for each key
         return {key: pd.concat([d[key] for d in df_list]) for key in df_list[0].keys()}
 
+    @support_date_range(frequency=None)
+    def get_60_day_sced_disclosure(
```
kmax12, line 1200: would it be straightforward to make `process` a parameter here like we do for `Ercot.get_60_day_sced_disclosure`?

gridstatus/ercot_api/ercot_api.py
WillKoehrsen: Yes. If you hit the endpoint for any of the files within the zipfile, it will return the same zipfile.

PR #546, opened 2025-02-25
feat(ENG 1315): MISO Look Ahead outages

Kladar:
## Summary
Adds the MISO `look_ahead_hourly` dataset. Also triggers MISO New Source Review, so typehints, logging, and fixturized tests have been implemented. 

### Data
Each file has a 7 day forecast of hourly data for Outage and MLTF information for each of the 4 regions. Each final df then has 24 hours * 7 days * 4 regions * 2 types = 1344 interval values. Once widen the Outage and MLTF are a column, so 672 rows per df. 
CleanShot 2025-02-25 at 11 52 48

### DST Behavior
It seems this dataset does not observe DST, as both the spring forward and fall back dates contain all 24 hours. It would significantly affect their data model if they had a different number of hours, so makes sense they skipped it. Also helps the fairly naive, brittle trimming of the dataframe I am doing to leave behind the stats and disclaimer. 

Spring Forward 2024
CleanShot 2025-02-25 at 10 27 55@2x

Fall Back 2024
CleanShot 2025-02-25 at 10 31 32@2x

### New Source Review
VCRification of the existing MISO tests was fairly straightforward, as was adding typehints and our system logger. LMK if questions here. 

VCR on second run reduces test time by ~20%

Initial run (integration-mode)
CleanShot 2025-02-25 at 12 30 18

Subsequent local runs (fixturized unit-mode)
CleanShot 2025-02-25 at 12 34 37

It's the `test_get_lmp_real_time_5_min_final_historical_date_range` that take a long time, whether fixturized or not (it's the data serialization into memory that takes a while methinks. Just a lot of data). Skipping that results in a super fast test suite, 80% faster than integration mode:

Skipping slow test:
CleanShot 2025-02-25 at 12 41 36

Neat

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -1394,3 +1394,20 @@ class MISO(ISOBase):
             self.default_timezone,
         )
         return market_date, publish_date
+
+    @support_date_range(frequency="DAY_START")
+    def get_look_ahead_hourly(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        if date == "latest":
+            return self.get_look_ahead_hourly(date="today", verbose=verbose)
+
+        url = f"https://docs.misoenergy.org/marketreports/{date.strftime('%Y%m%d')}_sr_la_rg.csv"
+        logger.info(f"Downloading look-ahead hourly data from {url}")
+
+        df = pd.read_csv(url, skiprows=4)
+
+        return df
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -1,4 +1,3 @@
-import io
 import json
 import re
 import urllib
@@ -1408,11 +1407,8 @@ class MISO(ISOBase):
 
         url = f"https://docs.misoenergy.org/marketreports/{date.strftime('%Y%m%d')}_sr_la_rg.csv"
         logger.info(f"Downloading look-ahead hourly data from {url}")
-        response = requests.get(url)
-        content = response.content.decode("utf-8")
-
         publish_date = pd.read_csv(
-            io.StringIO(content),
+            url,
             nrows=1,
             skiprows=1,
             header=None,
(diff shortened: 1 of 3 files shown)


Kladar:
Noting that failing MISO tests are from `MISOAPI`

gridstatus/miso.py
```diff
@@ -1323,35 +1395,86 @@ def _get_constraint_header_dates_from_excel(
...
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        if date == "latest":
+            return self.get_look_ahead_hourly(date="today", verbose=verbose)
 
-def add_interval_end(df: pd.DataFrame, duration_min: int) -> pd.DataFrame:
-    """Add an interval end column to a dataframe
+        url = f"https://docs.misoenergy.org/marketreports/{date.strftime('%Y%m%d')}_sr_la_rg.csv"
+        logger.info(f"Downloading look-ahead hourly data from {url}")
+        publish_date = pd.read_csv(
```
WillKoehrsen, line 1410: Is the `publish_date` always the date in the file name? In that case, I think we could do `publish_time = date.normalize()`.

gridstatus/miso.py
```diff
@@ -1323,35 +1395,86 @@ def _get_constraint_header_dates_from_excel(
...
+        ).iloc[0, 0]
+        publish_time = pd.to_datetime(publish_date.split(": ")[1]).tz_localize(
+            self.default_timezone,
+        )
 
-    Args:
-        df (pandas.DataFrame): Dataframe with a time column
-        duration_min (int): Interval duration in minutes
+        df_raw = pd.read_csv(url, skiprows=3)
+        df = df_raw.iloc[:96]  # 24 hours * 4 regions per hour
```
WillKoehrsen, line 1421: Nit: if the number of rows to ignore at the bottom of the csv are the same every time, we could use `skipfooter` in the `read_csv` call.

WillKoehrsen (approved):
A few minor comments, otherwise this looks good. I tried the latest date, the earliest date (2022-01-01) and DST start and end. All look correct!

PR #549, opened 2025-02-28
EIA Generators Fix Data Types

WillKoehrsen:
## Summary

- Fixes some EIA data types that were floats when they should have been integers
- Improves testing by checking data types and refacting tests

### DIFF
--- a/gridstatus/eia.py
+++ b/gridstatus/eia.py
@@ -775,11 +775,14 @@ class EIA:
                 )
 
         for col in GENERATOR_INT_COLUMNS:
-            if col in df.columns and df[col].dtype == "object":
+            if col in df.columns and (
+                df[col].dtype == "object" or df[col].dtype == "float"
+            ):
                 df[col] = (
                     df[col]
                     .astype(str)
                     .str.strip()
+                    .str.replace(".0", "")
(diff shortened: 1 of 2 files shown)


gridstatus/eia.py
```diff
@@ -775,11 +775,14 @@ def _handle_generator_data(
                 )
 
         for col in GENERATOR_INT_COLUMNS:
-            if col in df.columns and df[col].dtype == "object":
```
WillKoehrsen, line 778: `Entity ID` for some of the datasets was returned as a float. The issue was this conditional did not account for that column being a float in the dataset before conversion. Entity ID should be an integer looking at the raw data.

gridstatus/tests/source_specific/test_eia.py
```diff
@@ -426,6 +431,65 @@ def test_get_henry_hub_natural_gas_spot_prices_historical_date_range():
     assert df["Interval End"].max() == pd.Timestamp("2024-01-03", tz=HENRY_HUB_TIMEZONE)
 
 
+def _check_generators_data(
```
WillKoehrsen, line 434: Refactored the common checks into one method instead of repeating them in each test.

PR #551, opened 2025-02-28
feat(ENG 1319): ISONE 5 Min LMPs (via API)

Kladar:
## Summary
Adds the `isone_lmp_real_time_5_min_final` and `isone_lmp_real_time_5_min_prelim` datasets via the ISO NE API. There is an existing method that pulls website CSVs for `isone_lmp_real_time_5_min_prelim` though the existing dataset is referred to just as `isone_lmp_real_time_5_min`. This explicit prelim dataset via the API can replace that existing CSV dataset in downstream consumption if needed.

### ISO NE API Endpoints
Not all the endpoints I tried worked or worked as expected. Here is a summary:
CleanShot 2025-02-28 at 13 19 44@2x

The `ServiceInfo` for both `prelim/info` and `/info` is the same, and the non-specified `{type}` endpoints return the same data.
CleanShot 2025-02-28 at 13 27 46

`/prelim/current/all`
CleanShot 2025-02-28 at 13 47 36

`/current/all`
CleanShot 2025-02-28 at 13 48 20

This also matches with a comment that this dataset currently called just ISONE RT 5 Min LMP is actually Prelim.

However using the only endpoint that seems to work for `final/` data for all nodes, one that requests a `starthour/`, it took about 3 minutes to pull a single day, ~350,000 rows.  
CleanShot 2025-02-28 at 13 55 29

Same with `prelim/`, 3 minutes to pull one day. 
CleanShot 2025-02-28 at 14 03 40

There isn't another endpoint that appears like it would work, even with multiple calls. The `location_type/` endpoints are only available in `current/` format, otherwise that might be the most efficient. But I think the current implementation is the most efficient endpoint-wise.

### Tests
This is a lot of data, pulled over multiple slow calls, and parameterized to run several date ranges. This is where fixturization shines. Thought this was interesteding: local tests went from 11 minutes to _11 seconds_ from first (integration) to second (fixture) runs.

First
CleanShot 2025-02-28 at 14 42 38
 
Second
CleanShot 2025-02-28 at 14 44 22

This will slow down this test suite in CI though, until we can get the pull request CI tests fixturized.

### DIFF
--- a/gridstatus/isone_api/isone_api.py
+++ b/gridstatus/isone_api/isone_api.py
@@ -881,6 +881,8 @@ class ISONEAPI:
         """
 
         if date == "latest":
+            # NB: We don't quite know when this is published each day,
+            # and don't have a /current/all option for final data, so grab the full day on "latest"
             return self.get_lmp_real_time_5_min_prelim("today")
 
         url = f"{self.base_url}/fiveminutelmp/final/day/{date.strftime('%Y%m%d')}/starthour/{date.hour:02d}"
--- a/gridstatus/isone_api/isone_api.py
+++ b/gridstatus/isone_api/isone_api.py
@@ -836,6 +836,44 @@ class ISONEAPI:
         ].sort_values(["Interval Start", "Location"])
 
     @support_date_range("DAY_START")
+    def get_lmp_real_time_5_min_final(
+        self,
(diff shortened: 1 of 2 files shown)


WillKoehrsen (approved):
Code looks good and runs as expected locally. Nice!

PR #553, opened 2025-03-04
feat(ENG 1346): PJM dispatch reserves

Kladar:
## Summary
Adds the `pjm_dispatched_reserves_prelim` and `pjm_dispatched_reserves_verifed` datasets.

```
pjm = PJM()
df = pjm.get_dispatched_reserves_prelim(date="2025-03-03")
print(df)

df2 = pjm.get_dispatched_reserves_verified(date="2025-03-03")
print(df2)
```

### Details
Ran into this the other day with another set of `prelim` and `verified` datasets, where the `verified` is published the next day (and thus not available in a `today` format for `support_date_range`. I put a TODO in to add a `yesterday` option to `support_date_range` which I can do in this PR if we think it'll not have much in the way of side effects, or hold off until later. 

Also adds some final typehints to `PJM()` methods. I might also smuggle in another round of fixturization of some tests, just to keep that rolling since there are a lot for PJM.

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2763,7 +2763,12 @@ class PJM(ISOBase):
         https://dataminer2.pjm.com/feed/rt_dispatch_reserves/definition
         """
         if date == "latest":
-            date = "today"
+            # TODO: This is a hack to get the data for the previous day
+            # because the data is not available for the current day. Thinking about
+            # adding a "yesterday" to @support_date_range
+            date = (
+                pd.Timestamp.now(self.default_timezone) - pd.Timedelta(days=1)
+            ).strftime("%Y-%m-%d")
 
         df = self._get_pjm_json(
             "rt_dispatch_reserves",
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -1729,7 +1729,7 @@ class PJM(ISOBase):
 
         return df.sort_values("Publish Time").reset_index(drop=True)
 
-    def to_local_datetime(self, df, column_name):
+    def to_local_datetime(self, df: pd.DataFrame, column_name: str) -> pd.Series:
         return pd.to_datetime(df[column_name]).dt.tz_localize(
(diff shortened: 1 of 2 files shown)


gridstatus/tests/source_specific/test_pjm.py
```diff
@@ -2109,3 +2109,147 @@ def test_get_area_control_error_today_or_latest(self, date: str):
...
+
+            assert df["Ancillary Service"].dtype == object
+            assert df["Area"].dtype == object
+            assert df["Reserve Quantity"].dtype in [np.float64, np.int64]
+            assert df["Reserve Requirement"].dtype in [np.float64, np.int64]
+            assert df["Reliability Requirement"].dtype in [np.float64, np.int64]
+            assert df["Extended Requirement"].dtype in [np.float64, np.int64]
+            assert df["MW Adjustment"].dtype in [np.float64, np.int64]
+            assert df["Market Clearing Price"].dtype in [np.float64, np.int64]
+            assert df["Shortage Indicator"].dtype in [bool]
```
WillKoehrsen, line 2158: Could we also add a check for the unique values of "Area" we expect?

WillKoehrsen: And maybe the unique values of `Ancillary Service` too.

WillKoehrsen (approved):

gridstatus/pjm.py
```diff
@@ -2680,3 +2690,127 @@ def get_area_control_error(
...
+        df["Ancillary Service"] = df["Area"] + "-" + df["Reserve Type"]
+
+        df = df.replace({"Area": self.locale_abbreviated_to_full})
+
+        return df[
+            [
+                "Interval Start",
+                "Interval End",
+                "Ancillary Service",
+                "Area",
```
kmax12, line 2743: let's not drop "Reserve Type". i think it's useful to have that data still

PR #555, opened 2025-03-11
feat(ENG 1317): CAISO Scheduling Point / Tie Combo LMPs

Kladar:
## Summary
Adds the `caiso_lmp_scheduling_point_tie_combination_5_min`, `caiso_lmp_scheduling_point_tie_combination_15_min`, and `caiso_lmp_scheduling_point_tie_combination_hourly` datasets.

```
caiso = CAISO()

df = caiso.get_lmp_scheduling_point_tie_real_time_5_min(date="2025-03-10")
print(df)

df = caiso.get_lmp_scheduling_point_tie_real_time_15_min(date="2025-03-10")
print(df)

df = caiso.get_lmp_scheduling_point_tie_day_ahead_hourly(date="2025-03-10")
print(df)
```

### DIFF
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -2372,7 +2372,6 @@ class CAISO(ISOBase):
                 "Node",
                 "Tie",
                 "Market",
-                "GRP_TYPE",
             ],
             columns="LMP_TYPE",
             values="PRC",
@@ -2389,7 +2388,7 @@ class CAISO(ISOBase):
             },
         )
 
-        df["Node Tie"] = df["Node"] + " " + df["Tie"]
+        df["Location"] = df["Node"] + " " + df["Tie"]
         return df[
             [
                 "Interval Start",
--- a/gridstatus/caiso.py
+++ b/gridstatus/caiso.py
@@ -196,6 +196,9 @@ OASIS_DATASET_CONFIG = {
             "node": None,
             "grp_type": [None, "ALL", "ALL_APNODES"],
         },
+        "meta": {
+            "max_query_frequency": "1h",
+        },
     },
     "lmp_scheduling_point_tie_combination_15_min": {
         "query": {
@@ -209,6 +212,9 @@ OASIS_DATASET_CONFIG = {
             "node": None,
             "grp_type": [None, "ALL", "ALL_APNODES"],
         },
+        "meta": {
+            "max_query_frequency": "1h",
+        },
     },
     "lmp_scheduling_point_tie_combination_hourly": {
         "query": {
@@ -221,6 +227,9 @@ OASIS_DATASET_CONFIG = {
             "market_run_id": "DAM",
             "node": None,
             "grp_type": [None, "ALL", "ALL_APNODES"],
+        },
+        "meta": {
+            "max_query_frequency": "1d",
         },
     },
     "demand_forecast": {
@@ -2352,7 +2361,6 @@ class CAISO(ISOBase):
             columns={
                 "NODE": "Location",
                 "TIE": "Tie",
-                "GROUP": "Group",
                 "MARKET_RUN_ID": "Market",
             },
         )
@@ -2363,9 +2371,7 @@ class CAISO(ISOBase):
                 "Interval End",
                 "Location",
                 "Tie",
-                "Group",
                 "Market",
-                "POS",
                 "GRP_TYPE",
             ],
             columns="LMP_TYPE",
@@ -2391,9 +2397,7 @@ class CAISO(ISOBase):
                 "Market",
                 "Location",
                 "Node Tie",
-                "POS",
                 "Tie",
-                "Group",
                 "GRP_TYPE",
                 "Energy",
                 "Congestion",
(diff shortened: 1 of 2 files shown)


gridstatus/caiso.py
```diff
@@ -2247,3 +2286,118 @@ def get_as_procurement(
...
+        return df[
+            [
+                "Interval Start",
+                "Interval End",
+                "Market",
+                "Location",
+                "Node Tie",
+                "POS",
+                "Tie",
+                "Group",
```
kmax12, line 2396: we don't need to include `group`. I think that is just a number that corresponds to "Node Tie"

gridstatus/caiso.py
```diff
@@ -2247,3 +2286,118 @@ def get_as_procurement(
...
+                "GRP_TYPE",
```
kmax12, line 2397: there might be some valuable info in this column, but not we drop it for other LMP datasets.

unique values are `['ALL', 'ALL_APNODES']`

gridstatus/caiso.py
```diff
@@ -2247,3 +2286,118 @@ def get_as_procurement(
         df.columns.name = None
 
         return df
+
+    def get_lmp_scheduling_point_tie_combination_5_min(
```
kmax12, line 2290: I believe we need a `support_date_range` decorator on all these methods

gridstatus/caiso.py
kmax12: same applies to `POS` and `GRP_TYPE`

gridstatus/caiso.py
Kladar: should I drop `GRP_TYPE`?

gridstatus/caiso.py
Kladar: You're right though in that I needed to add the `meta` field for frequency to the `OASIS_DATASET_CONFIG` to use that `get_oasis_dataset` decorator. 5 and 15 min are hour_start, hourly is day_start

gridstatus/caiso.py
Kladar: Removed `POS` and `group`. lmk in comment below what we should do with `GRP_TYPE`.

gridstatus/caiso.py
```diff
@@ -2247,3 +2295,113 @@ def get_as_procurement(
...
+                "Market",
+                "GRP_TYPE",
+            ],
+            columns="LMP_TYPE",
+            values="PRC",
+            aggfunc="first",
+        ).reset_index()
+
+        df.columns.name = None
+        df = df.rename(
```
WillKoehrsen, line 2383: Should there be an LMP column in these datasets?

Kladar: Ah yes, nice catch, it's there until the final column filtering (and CAISO calls it LMP so doesn't need renaming)

WillKoehrsen:
Now I'm getting a KeyError using the latest code.

```python

CAISO().get_lmp_scheduling_point_tie_day_ahead_hourly('2025-01-01')

> KeyError: "['Location'] not in index"

```

Kladar:
> Now I'm getting a KeyError using the latest code.
> 
> ```python
> CAISO().get_lmp_scheduling_point_tie_day_ahead_hourly('2025-01-01')
> 
> > KeyError: "['Location'] not in index"
> ```

Oop, yep, columns in the helper function (and the tests) needed to be updated. Updated and running them locally now...

WillKoehrsen:
> Now I'm getting a KeyError using the latest code.

Fixed with the latest code!

gridstatus/caiso.py
```diff
@@ -2247,3 +2295,112 @@ def get_as_procurement(
         df.columns.name = None
 
         return df
+
+    def get_lmp_scheduling_point_tie_real_time_5_min(
```
WillKoehrsen, line 2299: We don't necessarily need to address this, but calling these functions with "latest" gives an error: `AttributeError: 'str' object has no attribute 'tz_convert'`.

Kladar: mm yeah, I think it's worth adding since it's pretty standard. Added it as `now` for real time. seems to be working pretty well and accurate to what is published.

For day ahead, it tries "tomorrow", and since it is likely published sometime in the morning for the following day (what I ran into), if that fails then do "today".

gridstatus/caiso.py
```diff
@@ -2247,3 +2295,134 @@ def get_as_procurement(
...
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        if date == "latest":
+            try:
+                df = self.get_lmp_scheduling_point_tie_day_ahead_hourly(
+                    pd.Timestamp.now(tz=self.default_timezone).normalize()
+                    + pd.Timedelta(days=1),
+                )
+            except ValueError:
```
WillKoehrsen, line 2362: I think the error in this case is a `KeyError`. 

```python
CAISO().get_lmp_scheduling_point_tie_day_ahead_hourly('2025-03-14')

> KeyError: 'PRC'
```

gridstatus/caiso.py
WillKoehrsen: Works well, thanks.

gridstatus/caiso.py
WillKoehrsen: No problem, thanks for working through it.

WillKoehrsen (approved):

PR #556, opened 2025-03-11
Remove missing column in SPP interconnection queue

bendnorman:
## Summary
Remove an extra column in the SPP interconnection queue data that was removed by SPP in the raw data.


WillKoehrsen (approved):
This seems to fix the problem. Thanks for the fix!

PR #557, opened 2025-03-19
feat(ENG 1371): CAISO Hasp LMP

Kladar:
## Summary
Adds the `caiso_lmp_hasp_15_min` dataset. Also does some slight reorg of caiso files. 

### Data
Columns are similar to the lmp scheduling point datasets so I followed decisions made in those. I think there is an error in the columns coming from source, as there is a `MW` column that should be a `PRC` column given that these are LMP prices, with price components. 

###Reorganization
Puts `caiso.py` and `caiso_utils.py` into a `caiso/` dir, and moves the `OASIS_DATASET_CONFIG` and a few other constants into the new `caiso_constants.py` file. Tests passing locally.


WillKoehrsen:
Instead of changing all the imports, can we add `from gridstatus.caiso.caiso import *  # noqa` to `gridstatus/caiso/__init__.py`?

docs/Examples/caiso/Ancillary Services.ipynb
Kladar, line 1: As always, notebooks in git are a nightmare

docs/Examples/caiso/DAM Heat Map.ipynb
Kladar, line 1: ugh

WillKoehrsen:
Or, maybe it should be more specific: `from gridstatus.caiso.caiso import CAISO` to avoid "polluting the namespace".

Kladar:
Okay, should be good now. Agreed on avoiding * imports where possible.

WillKoehrsen (approved):
Code runs without issue and test pass locally.

PR #558, opened 2025-03-20
feat(ENG-1403): NYISO AS Prices

Kladar:
## Summary
Adds the `nyiso_as_prices_day_ahead_hourly` and `nyiso_as_prices_real_time_5_min` datasets. Also does New Source Review of the `nyiso.py` and `test_nyiso.py` files.

```
from gridstatus.nyiso import NYISO
nyiso = NYISO()

df = nyiso.get_as_prices_day_ahead_hourly("latest")
df2 = nyiso.get_as_prices_day_ahead_hourly("2016-06-23")
df3 = nyiso.get_as_prices_real_time_5_min("latest")
df4 = nyiso.get_as_prices_real_time_5_min("2016-06-23")

print(df)
print(df2)
print(df3)
print(df4)
```

### Data
This data goes back to 1999 conceivably, but changes format a couple times over the years. It has a consistent format back to ~2016-07-01, so that's all that this supports right now. Seems pretty straightforward.

### Quality of Life
Updates the nyiso files to
- have function typehints
- use the new logging
- include the hidden method `_handle_time()` since it's only used in the NYISO class (`self._handle_time`)
- break up some tests that are doing many things (`various_edt_to_est`, etc)
- update the tests to use VCR

**Before Local Fixtures**
CleanShot 2025-03-20 at 14 37 27@2x

**After Local Fixtures** _Time reduction: ~70%_
CleanShot 2025-03-20 at 15 29 32@2x

### DIFF
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -31,6 +31,8 @@ BTM_SOLAR_FORECAST_DATASET = "btmdaforecast"
 INTERFACE_LIMITS_AND_FLOWS_DATASET = "ExternalLimitsFlows"
 LAKE_ERIE_CIRCULATION_REAL_TIME_DATASET = "eriecirculationrt"
 LAKE_ERIE_CIRCULATION_DAY_AHEAD_DATASET = "eriecirculationda"
+AS_PRICES_DAY_AHEAD_HOURLY_DATASET = "damasp"
+AS_PRICES_REAL_TIME_5_MIN_DATASET = "rtasp"
 
 """
 Pricing data:
@@ -55,6 +57,8 @@ DATASET_INTERVAL_MAP: Dict[str, DatasetInterval] = {
     INTERFACE_LIMITS_AND_FLOWS_DATASET: DatasetInterval("start", 5),
     LAKE_ERIE_CIRCULATION_REAL_TIME_DATASET: DatasetInterval("instantaneous", None),
     LAKE_ERIE_CIRCULATION_DAY_AHEAD_DATASET: DatasetInterval("instantaneous", None),
+    AS_PRICES_DAY_AHEAD_HOURLY_DATASET: DatasetInterval("start", 60),
+    AS_PRICES_REAL_TIME_5_MIN_DATASET: DatasetInterval("start", 5),
 }
 
 
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -1298,15 +1298,7 @@ class NYISO(ISOBase):
                 report (will refer to month and year)
         """
         if date == "latest":
-            try:
-                return self.get_as_prices_real_time_5_min(
-                    (
-                        pd.Timestamp.now(tz=self.default_timezone).normalize()
-                        + pd.DateOffset(days=1)
-                    ).strftime("%Y-%m-%d"),
-                )
-            except urllib.error.HTTPError:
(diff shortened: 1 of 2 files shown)


gridstatus/nyiso.py
```diff
@@ -64,8 +73,73 @@ class NYISO(ISOBase):
...
+            )
+
+        return df
+
     @support_date_range(frequency="MONTH_START")
-    def get_status(self, date, end=None, verbose=False):
+    def get_status(
+        self,
+        date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
+        end: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp] | None = None,
```
WillKoehrsen, line 140: Since `date` and `end` are arguments to nearly all functions and they all have the same allowed types, how do you feel about defining constants for these types instead of having to write them out for every function?

```python
DATE_TYPE_ANNOTATION = str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp]
END_TYPE_ANNOTATION = str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp] | None
```

could go in `gridstatus/utils.py`.

gridstatus/nyiso.py
```diff
@@ -1151,74 +1244,104 @@ def get_capacity_prices(self, date=None, verbose=False):
...
+            date=date,
+            verbose=verbose,
+            dataset_name="damasp",
         )
+        df = self._handle_as_prices(df, 60)
+        return df
 
-    df = df.rename(columns={time_stamp_col: "Time"})
+    @support_date_range(frequency="DAY_START")
+    def get_as_prices_real_time_5_min(
```
WillKoehrsen, line 1288: The first interval returned by this method is from 0005 to 0010. Is this the desired behavior? With this behavior, to get the first interval of the day, you need to query for the previous day.

```python
NYISO().get_as_prices_real_time_5_min('2025-01-01')['Interval Start'].min()

>  Timestamp('2025-01-01 00:05:00-0500', tz='US/Eastern')

NYISO().get_as_prices_real_time_5_min('2025-01-01')['Interval Start'].max()

> Timestamp('2025-01-02 00:00:00-0500', tz='US/Eastern')
```

Some other NYISO methods have this same behavior (`get_fuel_mix`) and others do not (`get_load`, `get_lmp` with `Markets.REAL_TIME_5_MIN`).

Kladar: Oh yeah good catch, it looks like according to the data spec that real time is interval ending. 
CleanShot 2025-03-21 at 09 38 30

Kladar: I also think the real_time("latest") check for tomorrow data is not needed, since there is never day ahead data for real time.

Kladar: okay these have been fixed.

gridstatus/nyiso.py
WillKoehrsen: Yeah, I agree there's a more thorough refactoring of date handling we can do here. That can wait for future work.

gridstatus/nyiso.py
WillKoehrsen: Thanks, data looks as expected now.

WillKoehrsen (approved):
👍🏽

PR #561, opened 2025-03-24
feat(ENG 1371): CAISO 15 Min Tie Flows and Renewable Forecast

Kladar:
## Summary
Adds the `hasp_renewable_forecast` and `tie_flows_real_time_15_min` datasets for CAISO

```
from gridstatus import CAISO
caiso = CAISO()

df = caiso.get_hasp_renewable_forecast_hourly("2025-03-20")
print(df)

df = caiso.get_tie_flows_real_time_15_min("2025-03-20")
print(df)
```

### Details
There is already a real time tie flows dataset called `Real Time`, which is actually the `5 Minute Real Time` version of this 15 Min tie flow dataset. It has the same data model and was straightforward to add.

The renewables forecast data is technically a day ahead, hourly forecast. The market operator does not provide a publish time timestamp. We can try to add one if we want. The description from CAISO is that it is published each day for the next day before the market clearing results are published. 
CleanShot 2025-03-24 at 11 45 04@2x

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -2236,7 +2236,7 @@ class CAISO(ISOBase):
         end: str | pd.Timestamp | None = None,
         verbose: bool = False,
     ) -> pd.DataFrame:
-        """Get solar generation HASP hourly data from CAISO.
+        """Get solar and wind generation HASP hourly data from CAISO.
 
         Args:
             date (str | pd.Timestamp): date to return data
@@ -2245,7 +2245,7 @@ class CAISO(ISOBase):
             verbose (bool, optional): print out url being fetched. Defaults to False.
 
         Returns:
-            pandas.DataFrame: A DataFrame of solar generation HASP hourly data
+            pandas.DataFrame: A DataFrame of solar and wind generation HASP hourly data
         """
         if date == "latest":
             try:
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -2274,7 +2274,7 @@ class CAISO(ISOBase):
                 "Interval Start",
                 "Interval End",
                 "Location",
-                "Renewable Type",
-                "MW",
+                "Solar",
+                "Wind",
             ]
         ]
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -2248,17 +2248,9 @@ class CAISO(ISOBase):
             pandas.DataFrame: A DataFrame of solar and wind generation HASP hourly data
         """
         if date == "latest":
-            try:
-                df = self.get_hasp_renewable_forecast_hourly(
-                    pd.Timestamp.now(tz=self.default_timezone).normalize()
-                    + pd.Timedelta(days=1),
-                )
-            except KeyError:
-                df = self.get_hasp_renewable_forecast_hourly(
-                    "today",
-                )
-
-            return df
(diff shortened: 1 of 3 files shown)


gridstatus/caiso/caiso.py
```diff
@@ -2209,3 +2229,72 @@ def _handle_lmp_hasp_15_min(self, df: pd.DataFrame) -> pd.DataFrame:
...
             ]
         ]
+
+    def get_hasp_renewable_forecast_hourly(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """Get solar generation HASP hourly data from CAISO.
```
WillKoehrsen, line 2239: Nit: this is a solar and wind forecast dataset right?

WillKoehrsen:
> The market operator does not provide a publish time timestamp. We can try to add one if we want. 

Since this is a forecast, we should try to add a publish time if we can. If I'm reading the OASIS publications schedule correctly, it looks like this dataset is published 90 minutes ahead of time.

Screenshot 2025-03-24 at 15 32 35

WillKoehrsen:
I think the text in the screenshot from the description ("The forecast MW value for each...") is incorrect. The forecast looks to only go out two hours and is published hourly.

Screenshot 2025-03-24 at 15 42 04

Kladar:
Added publish time. I think I don't need to modify the `max_query_frequency` from `1d`, but lmk if you see it differently. 1h was giving some strange query behavior.

gridstatus/caiso/caiso.py
```diff
@@ -2209,3 +2229,74 @@ def _handle_lmp_hasp_15_min(self, df: pd.DataFrame) -> pd.DataFrame:
...
+            verbose (bool, optional): print out url being fetched. Defaults to False.
+
+        Returns:
+            pandas.DataFrame: A DataFrame of solar and wind generation HASP hourly data
+        """
+        if date == "latest":
+            try:
+                df = self.get_hasp_renewable_forecast_hourly(
+                    pd.Timestamp.now(tz=self.default_timezone).normalize()
+                    + pd.Timedelta(days=1),
```
WillKoehrsen, line 2254: Should this be 1 or 2 hours instead of 1 day?

Kladar: Yeah. I think I have this ~correct: added some logic to figure out what the latest hour is (basically, if the publish times are correct and it is 90 minutes ahead, then during the back half of an hour the latest time will be two hours ahead, otherwise one hour ahead).

- At 1:29 pm the latest forecast available is 2pm
- At 1:30 pm the latest forecast available is 3pm (T+90 minutes)
- 3pm will continue to be the latest available until 2:30pm, when the 4pm interval becomes available.

Kladar: And current when I use this (it's 1:51pm PT), I get the 3pm interval, which is expected.

WillKoehrsen: Hmm, I think always looking ahead two hours might be more reliable since we don't know the exact publish time.

Kladar: the issue is that the 2 hour ahead interval might not exist, and then you have to fall back to retry logic anyway.

Kladar: but yeah we can try 2 hours ahead and if not exist then 1 hour ahead

WillKoehrsen: Nice, thanks!

PR #563, opened 2025-03-27
feat: Replace tabula (java) dependency

Kladar:
## Summary
The `tabula-py` library is a wrap around the java library, which then makes the repo and systems that use it require Java. This replaces it with the well-regarded pure python repository `pdfplumber`, maintaining the same functionality but removing the Java dependency.

```

caiso = CAISO()
df = caiso.get_curtailment(date="2024-01-01")
df2 = caiso.get_curtailment(date="2025-01-02")
df3 = caiso.get_curtailment(date="2021-12-02")

print(df)
print(df2)
print(df3)
```

### Details
I also expand the tests to check some of the dates mentioned in the code, such as 2021-12-02 which has a weird label and something after 2024-05-31 which is when the PDF URL changes.


kmax12 (approved):
I ran this locally and before and after I got the same 1921 rows

```python
import gridstatus; 
iso = gridstatus.CAISO();
df = iso.get_curtailment(start="2025-01-01", end="today")
```

It also ran in half the time with the new library

gridstatus/tests/source_specific/test_caiso.py
```diff
@@ -324,6 +324,12 @@ def test_get_curtailment_3_pages(self, date):
             assert df.shape == (76, 8)
             self._check_curtailment(df)
 
+    @pytest.mark.parametrize("date", ["2021-12-02", "2025-01-02"])
```
WillKoehrsen, line 327: What's special about these dates?

Kladar: 2021-12-02 is called out in the code as having a wrong year in the filename 
```
# Handle specific case where dec 02, 2021 has wrong year in file name
        if date_str == "dec02_2021":
            date_str = "02dec_2020"
```

2025-01-02 is after the 2024-05-31 date where the URL changes (didn't have any existing tests for that code path).

WillKoehrsen: Thanks for the explanation.

WillKoehrsen:
There's one small issue: when pulling data for 2025-03-26 (the most [recent report](https://www.caiso.com/documents/wind-solar-real-time-dispatch-curtailment-report-mar-26-2025.pdf)), the old code correctly identifies the 3 NA values for "Curtailment (MW)" while the new code stores these as empty strings.

__Old__ 

```python
df = CAISO().get_curtailment('2025-03-26')

df.isna().sum()

Time                  0
Interval Start        0
Interval End          0
Curtailment Type      0
Curtailment Reason    0
Fuel Type             0
Curtailment (MWh)     0
Curtailment (MW)      3
```

__New__ 

```python
df = CAISO().get_curtailment('2025-03-26')

df.isna().sum()

Time                  0
Interval Start        0
Interval End          0
Curtailment Type      0
Curtailment Reason    0
Fuel Type             0
Curtailment (MWh)     0
Curtailment (MW)      0
```

The difference is that in the new method, the empty cells from the source data are stored as '' in the dataframe. Hopefully, this is easy to fix using a `df.replace`.

Kladar:
> ```python
> df = CAISO().get_curtailment('2025-03-26')
> 
> df.isna().sum()
> ```

Nice find, added a `df.replace`

PR #564, opened 2025-03-27
feat(ENG 1413): PJM Regulation Market

Kladar:
## Summary
Adds the `pjm_regulation_market_monthly` dataset

```
pjm = PJM()
df = pjm.get_regulation_market_monthly(date="2024-01-01", end="2025-01-01")
print(df)
```

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2887,6 +2887,24 @@ class PJM(ISOBase):
             },
         )
 
+        df = df.astype(
+            {
+                "RegD SSMW": float,
+                "RegA SSMW": float,
+                "RegD Procure": float,
+                "RegA Procure": float,
+                "Total MW": float,
+                "Deficiency": float,
+                "RTO Perfscore": float,
+                "RegA Mileage": float,
+                "RegD Mileage": float,
+                "RegA Hourly": float,
+                "RegD Hourly": float,
+                "Is Approved": bool,
+            },
+        )
+        df["Modified Datetime UTC"] = pd.to_datetime(df["Modified Datetime UTC"])
+
         return df[
             [
                 "Interval Start",
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2817,7 +2817,7 @@ class PJM(ISOBase):
             ]
         ]
 
-    @support_date_range(frequency=None)
+    @support_date_range(frequency="MONTH_START")
     def get_regulation_market_monthly(
         self,
         date: str | pd.Timestamp,
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2816,3 +2816,69 @@ class PJM(ISOBase):
                 "Deficit",
             ]
         ]
+
+    @support_date_range(frequency=None)
+    def get_regulation_market_monthly(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieves the PJM Regulation Market Monthly data from:
+        https://dataminer2.pjm.com/feed/reg_market_results/definition
+        """
+        if date == "latest":
+            date = "today"
+
+        df = self._get_pjm_json(
+            "reg_market_results",
+            start=date,
+            end=end,
+            params={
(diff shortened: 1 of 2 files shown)


gridstatus/tests/source_specific/test_pjm.py
```diff
@@ -2303,3 +2303,80 @@ def test_get_dispatched_reserves_verified_historical_range(self):
...
+        "RTO Perfscore",
+        "RegA Mileage",
+        "RegD Mileage",
+        "RegA Hourly",
+        "RegD Hourly",
+        "Is Approved",
+        "Modified Datetime UTC",
+    ]
+
+    def test_get_regulation_market_monthly_latest(self):
```
WillKoehrsen, line 2326: This test is failing for me locally, probably because it's before the 7th.

gridstatus/pjm.py
```diff
@@ -2816,3 +2816,94 @@ def get_dispatched_reserves_verified(
...
+                "requirement": "Requirement",
+                "total_mw": "Total MW",
+                "deficiency": "Deficiency",
+                "rto_perfscore": "RTO Perfscore",
+                "rega_mileage": "RegA Mileage",
+                "regd_mileage": "RegD Mileage",
+                "rega_hourly": "RegA Hourly",
+                "regd_hourly": "RegD Hourly",
+                "is_approved": "Is Approved",
+                "modified_datetime_utc": "Modified Datetime UTC",
```
WillKoehrsen, line 2886: Do we want to convert `Modified Datetime UTC` to a datetime?

gridstatus/tests/source_specific/test_pjm.py
Kladar: hmm, I was hoping this would work no matter what - I'll workshop it a bit

Kladar: okay, I added some better step back logic to look back more months if data isn't found in `latest` (up to a year to avoid an endless situation). Should be working now.

gridstatus/pjm.py
```diff
@@ -2816,3 +2816,108 @@ def get_dispatched_reserves_verified(
...
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieves the PJM Regulation Market Monthly data from:
+        https://dataminer2.pjm.com/feed/reg_market_results/definition
+        """
+        if date == "latest":
```
WillKoehrsen, line 2831: This approach doesn't seem to be getting the actual latest data. 


```python
PJM().get_regulation_market_monthly('latest')['Interval Start'].max()

> Timestamp('2025-02-01 23:00:00-0500', tz='US/Eastern')

PJM().get_regulation_market_monthly('2025-02-28')['Interval Start'].max()

>  Timestamp('2025-02-28 23:00:00-0500', tz='US/Eastern')
```

I think this is because it's querying for the first day of the month. The latest data is from the last day of the month.

gridstatus/tests/source_specific/test_pjm.py
WillKoehrsen: Yep, it's passing now.

gridstatus/pjm.py
WillKoehrsen: We can fix the issue with the `MonthEnd` offset:

```python
current_date = pd.Timestamp.now(self.default_timezone).replace(
    day=1,
    hour=0,
    minute=0,
    second=0,
    microsecond=0,
) + pd.tseries.offsets.MonthEnd(0)
```

https://pandas.pydata.org/docs/reference/api/pandas.tseries.offsets.MonthEnd.html

Kladar: Ah yeah my mistake. Was thinking this was returning the full month of data, but it's returning a days worth.

gridstatus/pjm.py
```diff
@@ -2816,3 +2816,108 @@ def get_dispatched_reserves_verified(
                 "Deficit",
             ]
         ]
+
+    @support_date_range(frequency="MONTH_START")
```
WillKoehrsen, line 2820: This can be `frequency=None` right because we can fetch more than one month of data at a time?

Kladar: done

WillKoehrsen (approved):
Thanks for addressing all my comments.

PR #565, opened 2025-03-31
Specify Format for PJM Datetimes

WillKoehrsen:
## Summary

- Specify `format="ISO8601"` when parsing PJM datetimes
- Required because PJM recently started publishing datetimes with milliseconds


WillKoehrsen:
`pjm` tests are passing.

PR #566, opened 2025-04-01
Security Updates April 2025

WillKoehrsen:
## Summary

- Updates libraries for Dependabot security warnings


pyproject.toml
```diff
@@ -50,7 +50,9 @@ certifi = "^2024.7.4"
 setuptools = "^70.0.0"
 cryptography = "^43.0.1"
 urllib3 = "^2.2.2"
-tornado = "^6.4.1"
+tornado = "^6.4.2"
```
kmax12, line 53: strikes me as a little weird `tornado` is a dep in this library. are you able to figure out what is adding it? perhaps we can remove

kmax12: im wondering if it is actually one of the dev dependencies

WillKoehrsen: Yeah, I think you're right:

```
$ poetry show tornado


 name         : tornado                                                                                                    
 version      : 6.4.2                                                                                                      
 description  : Tornado is a Python web framework and asynchronous networking library, originally developed at FriendFeed. 

required by
 - ipykernel requires >=6.1
 - jupyter-client requires >=6.2
 - jupyter-server requires >=6.2.0
 - jupyterlab requires >=6.2.0
 - notebook requires >=6.2.0
 - terminado requires >=6.1.0
 ```

The same applies to `jinja2`. I've moved both of those pins to dev dependencies.

kmax12 (approved):

PR #567, opened 2025-04-02
feat(ENG 1433): MISO Load Zonal Hourly

Kladar:
## Summary
Adds the `miso_load_zonal_hourly` dataset. Consolidates some code between the `load_forecast` method as well, since it is the same file.

### Details
Column names are from the dataset spec.

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -164,11 +164,67 @@ class MISO(ISOBase):
         if date == "latest":
             return self.get_load_forecast(date="today", verbose=verbose)
 
+        df = self._get_load_forecast_file(date)
+        df = df.loc[
+            df["Interval Start"] >= date,
+            [col for col in df if "ActualLoad" not in col],
+        ]
+
+        return df.sort_values("Interval Start").reset_index(drop=True)
+
+    @support_date_range(frequency="DAY_START")
+    def get_load_zonal_hourly(
+        self,
+        date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        https://docs.misoenergy.org/marketreports/YYYYMMDD_df_al.xls
+        """
+        if date == "latest":
+            return self.get_load_zonal_hourly(date="today", verbose=verbose)
+
+        df = self._get_load_forecast_file(date)
+
+        df = df.rename(
+            columns={
+                "LRZ1 ActualLoad": "LRZ1 Load",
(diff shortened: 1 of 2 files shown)


gridstatus/miso.py
```diff
@@ -164,11 +164,73 @@ def get_load_forecast(
...
+        ).drop(columns=["Market Day", "HourEnding"])
+
+        df = df.sort_values("Interval Start").reset_index(drop=True)
+        df = df.dropna()
+        return df[
+            [
+                "Interval Start",
+                "Interval End",
+                "Publish Time",
+                "LRZ1 Load",
```
kmax12, line 220: im leaning toward not including "Load" in the processed column name since that is already in the dataset id.

gridstatus/miso.py
```diff
@@ -164,11 +164,73 @@ def get_load_forecast(
...
+            ["Interval Start", "Interval End", "Publish Time"],
+        ).drop(columns=["Market Day", "HourEnding"])
+
+        df = df.sort_values("Interval Start").reset_index(drop=True)
+        df = df.dropna()
+        return df[
+            [
+                "Interval Start",
+                "Interval End",
+                "Publish Time",
```
kmax12, line 219: i don't think we need publish time here. im fine if you want to return it, but we should drop in downstream before uploading to API

kmax12 (approved):
looking good to me. just comments on the schema

gridstatus/miso.py
Kladar: I'm in favor of this

gridstatus/miso.py
```diff
@@ -164,11 +164,73 @@ def get_load_forecast(
...
+        df = utils.move_cols_to_front(
+            df,
+            ["Interval Start", "Interval End", "Publish Time"],
+        ).drop(columns=["Market Day", "HourEnding"])
+        return df.sort_values("Interval Start").reset_index(drop=True)
 
-        logger.info(f"Downloading load forecast data from {url}")
-        df = pd.read_excel(url, sheet_name="Sheet1", skiprows=4, skipfooter=1)
+    @support_date_range(frequency="DAY_START")
+    def get_load_zonal_hourly(
```
WillKoehrsen, line 179: Nit: I would call this `get_zonal_load_hourly` because the end dataset will be `miso_zonal_load_hourly`.

gridstatus/miso.py
Kladar: I'll drop it here

gridstatus/miso.py
Kladar: done

Kladar:
tests passing locally for MISO

PR #571, opened 2025-04-04
Move rounding before pivot in PJM load

kmax12:


kmax12:
beat me to it in #570

PR #572, opened 2025-04-08
feat(ENG 1460): ISONE Capacity 7 Day

Kladar:
## Summary
Adds the `isone_capacity_forecast_7_day` dataset via the ISONE API

### Details
There are two backfill endpoints: the `day/{day}` and `day/{day}/all` endpoints - they seem to give back the same data most of the time, but turns out they sometimes publish multiple forecasts for the same day, one after another. Sometimes up to 5. For instance, on `2023-05-07`, there are several forecasts, one published at `09:22:56-04:00` and another at `09:23:05-04:00`. I opt for the `all/` endpoint because we can grab all the data for a backfill.

It's not the most elegant JSON response parsing, especially for the weather columns, but it's consistent and fast. I spot checked a few weather parsings manually from the response object itself.


gridstatus/tests/source_specific/test_isone_api.py
```diff
@@ -808,3 +808,108 @@ def test_get_lmp_real_time_5_min_final_date_range(self, date: str, end: str):
...
+            result = self.iso.get_capacity_forecast_7_day(date="latest")
+
+            assert isinstance(result, pd.DataFrame)
+            assert len(result) > 0
+            assert list(result.columns) == self.capacity_forecast_7_day_columns
+            assert result["High Temperature Boston"].dtype == np.int64
+            assert result["Dew Point Boston"].dtype == np.int64
+            assert result["High Temperature Hartford"].dtype == np.int64
+            assert result["Dew Point Hartford"].dtype == np.int64
+            assert result["Generating Capacity Position"].dtype == object
```
WillKoehrsen, line 856: Not sure about this being an object as opposed to an integer. Did you see any non-null values? I pulled a year of the most recent data and all the values were null.

gridstatus/tests/source_specific/test_isone_api.py
```diff
@@ -808,3 +808,108 @@ def test_get_lmp_real_time_5_min_final_date_range(self, date: str, end: str):
...
+            assert result["Import at Time of Peak"].dtype == np.int64
+            assert result["Total Available Generation and Imports"].dtype == np.int64
+            assert result["Projected Peak Load"].dtype == np.int64
+            assert result["Replacement Reserve Requirement"].dtype == np.int64
+            assert result["Required Reserve"].dtype == np.int64
+            assert result["Required Reserve Including Replacement"].dtype == np.int64
+            assert result["Total Load Plus Required Reserve"].dtype == np.int64
+            assert result["Projected Surplus or Deficiency"].dtype == np.int64
+            assert result["Available Demand Response Resources"].dtype == np.int64
+            assert result["Available Realtime Emergency Generation"].dtype == object
```
WillKoehrsen, line 871: This also seems like it could be an integer, although again, all the values for the past year were null so I couldn't confirm.

gridstatus/tests/source_specific/test_isone_api.py
```diff
@@ -808,3 +808,108 @@ def test_get_lmp_real_time_5_min_final_date_range(self, date: str, end: str):
...
+            assert result["Load Relief Actions Anticipated"].dtype == object
+            assert result["Power Watch"].dtype == object
+            assert result["Power Warning"].dtype == object
+            assert result["Cold Weather Watch"].dtype == object
+            assert result["Cold Weather Warning"].dtype == object
+            assert result["Cold Weather Event"].dtype == object
```
WillKoehrsen, line 915: Could we put these checks in a function to avoid needing to repeat them?

gridstatus/isone_api/isone_api.py
```diff
@@ -982,3 +982,127 @@ def _handle_lmp_real_time(
...
+        )
+        df["Dew Point Hartford"] = df["Weather.CityWeather"].apply(
+            lambda x: next(
+                (city["DewPointF"] for city in x if city["CityName"] == "Hartford"),
+                None,
+            ),
+        )
+
+        return df[
+            [
```
WillKoehrsen, line 1077: Nit: we could put these columns in a constant to use here and in the tests.

gridstatus/tests/source_specific/test_isone_api.py
Kladar: it seems all the values are null. did a local
```
$ gsetl backfill -s isone -d capacity_forecast_7_day -st 2018-01-01 -end 2025-04-08 --freq 1D
```
to backfill the whole thing and then a 
```
select * from gridstatus.public.isone_capacity_forecast_7_day where generating_capacity_position != null;
```
and got nothing. 

Same for `available_realtime_emergency_generation`. Removed the checks.

gridstatus/tests/source_specific/test_isone_api.py
Kladar: yeah, datasets should probably be some sort of class like a pydantic model that we can define and evaluate eventually

gridstatus/isone_api/isone_api.py

gridstatus/tests/source_specific/test_isone_api.py
WillKoehrsen: Got it, thanks for checking.

PR #573, opened 2025-04-14
feat(ENG 1045): IESO Resource Adequacy Retry Logic

Kladar:
## Summary
This adds some retry logic to the `get_resource_adequacy_report`, namely the underlying `_get_all_resource_adequacy_jsons()` method.

### Details
The error seems to be due to either a connection sensitivity on lots of requests, since we are multithreading the report pull in `vintage="all"` mode and it can pull a few dozen reports in rapid fashion, or a timing issue that the report title and link is posted by the ISO but there is no underlying report to grab (yet). It usually fails on only one file, and often the last file of the found set, so this retry logic should handle both cases moderately well without being too aggressive. Can adjust as needed. I believe I got the specific error type but I catch all and we can adjust that too, if needed.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -1318,6 +1318,9 @@ class IESO(ISOBase):
             )
 
         json_data_with_times = []
+        max_retries = 3
+        retry_delay = 2
+
         with ThreadPoolExecutor(max_workers=min(10, len(filtered_files))) as executor:
             future_to_file = {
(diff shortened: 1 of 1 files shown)


gridstatus/ieso.py
Kladar, line 1: Tests for IESO are not in great shape (independent of this PR) since they were done in our very first pass of VCR and some things need adjusting, but I will leave that for another time.

WillKoehrsen (approved):

PR #575, opened 2025-04-16
feat(ENG 1496): PJM LMP Real Time Hourly Unverified

Kladar:
## Summary
Adds the `pjm_lmp_real_time_unverified_hourly` dataset. Also finish up VCR-ification of PJM test fixtures

```
pjm = PJM()
df = pjm.get_lmp_real_time_unverified_hourly(start="2025-04-10", end="2025-04-11")
df2 = pjm.get_lmp_real_time_unverified_hourly(start="2025-04-10", end="2025-04-11", location_type="HUB")
df3 = pjm.get_lmp_real_time_unverified_hourly(date="latest")
df4 = pjm.get_lmp_real_time_unverified_hourly(date="2025-04-10", locations="ALL")

print(df)
print(df2)
print(df3)
print(df4)
```

## Details
This bypasses the existing PJM `get_lmp()` method in favor of keeping it separate, while also taking some things from the general method like location filtering. 

### Tests
Not much of a speed up, since there are some tests that take a long time even with fixtures, but now everything is consistent.

**Before**
CleanShot 2025-04-16 at 14 16 33@2x

**After**
CleanShot 2025-04-16 at 14 33 51@2x


gridstatus/pjm.py
```diff
@@ -976,6 +603,79 @@ def _add_pnode_info_to_lmp_data(self, data: pd.DataFrame) -> pd.DataFrame:
...
+    ) -> pd.DataFrame:
+        """Get real-time unverified hourly LMPs"""
+
+        if date == "latest":
+            date = "today"
+
+        params = {
+            "fields": "datetime_beginning_utc, datetime_beginning_ept, pnode_name, type, total_lmp_rt, congestion_price_rt, marginal_loss_price_rt",  # noqa: E501
+        }
+        if location_type:
```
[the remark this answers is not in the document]
Kladar: yeah we should, fixed. I also changed it from "Location Name" to "Location"

gridstatus/pjm.py
```diff
@@ -976,6 +603,80 @@ def _add_pnode_info_to_lmp_data(self, data: pd.DataFrame) -> pd.DataFrame:
...
+                "Interval Start",
+                "Interval End",
+                "Location",
+                "Location Type",
+                "LMP",
+                "Energy",
+                "Congestion",
+                "Loss",
+            ]
+        ]
```
Kladar, line 678: `Location Id` and `Location Short Name` come from the `self._add_pnode_info_to_lmp_data` method, which relies on the `pnode_id` coming from source. However, this dataset does not have a `pnode_id` field from which to populate those columns, only `pnode_name`. 

https://dataminer2.pjm.com/feed/rt_unverified_hrl_lmps/definition

gridstatus/pjm.py
```diff
@@ -34,387 +38,10 @@ class PJM(ISOBase):
...
-        "5021665",
-        "5021666",
-        "5021667",
-        "2156111847",
-        "93353961",
-        "93353963",
-        "93353965",
-    ]
-
+    location_types = LOCATION_TYPES
```
WillKoehrsen, line 41: Since these constants are mutable types, if we were to make a change to these attributes in this class, that would change the constant itself. It's safer to make copies of these constants:

```python
location_types = LOCATION_TYPES[:]
hub_node_ids = HUB_NODE_IDS[:]
...
```

WillKoehrsen: Thanks, that seems like a good solution.

WillKoehrsen (approved):
👍🏽

PR #576, opened 2025-04-17
IESO HOEP Real Time Fix Duplicates

WillKoehrsen:
## Summary

- Fix duplicate issue in IESO HOEP Real Time Hourly that occurred on 2025-04-16


gridstatus/ieso.py
```diff
@@ -1045,6 +1045,10 @@ def get_hoep_real_time_hourly(
                 )
             raise
 
+        # On this date, IESO published duplicates for hour ending instead of going 1-24
```
WillKoehrsen, line 1048: Duplicates published on 2025-04-16

Screenshot 2025-04-17 at 07 35 41

https://reports-public.ieso.ca/public/DispUnconsHOEP/PUB_DispUnconsHOEP_20250416.xml

WillKoehrsen: Assuming the rows are in order, we can just use the index as the correct hour endings.

kmax12: I actually don't think this is correct assumption if you look at the generated time of the report.

given the report stops at 19:00 when the day as completed is suspect as well

PR #577, opened 2025-04-18
Use params for caiso load forecast

kmax12:
I was looking at oasis and noticed we can use API parameters rather than filtering in python code, which feels a lot cleaner

## Testing

```python
In [1]: import gridstatus
In [2]: iso = gridstatus.CAISO()
In [3]: df_5_min = iso.get_load_forecast_5_min("today")
In [4]: df_15_min = iso.get_load_forecast_15_min("today")
In [5]: (df_5_min["Interval End"] - df_5_min["Interval Start"]).unique()
Out[5]: 
<TimedeltaArray>
['0 days 00:05:00']
Length: 1, dtype: timedelta64[ns]

In [6]: (df_15_min["Interval End"] - df_15_min["Interval Start"]).unique()
Out[6]: 
<TimedeltaArray>
['0 days 00:15:00']
Length: 1, dtype: timedelta64[ns]
```

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -565,11 +565,9 @@ class CAISO(ISOBase):
             raw_data=False,
             verbose=verbose,
             sleep=sleep,
-            params={"market_run_id": "RTM"},
+            params={"market_run_id": "RTM", "execution_type": "RTD"},
         )
-        df = df[
-            ((df["Interval End"] - df["Interval Start"]).dt.total_seconds() / 60) == 5
-        ]
+
         df = df.rename(
             columns={"MW": "Load Forecast", "TAC_AREA_NAME": "TAC Area Name"},
         )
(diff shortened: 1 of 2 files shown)


kmax12:
@Kladar I think my local env isn't set up correctly, would you mind taking this PR across the finish line?

also, not sure if there are updates to VCR/tests. This should give same results as previous implementation, but we are calling the API differently

Kladar:
> @Kladar I think my local env isn't set up correctly, would you mind taking this PR across the finish line?
> 
> also, not sure if there are updates to VCR/tests. This should give same results as previous implementation, but we are getting a different response from the API

Sounds good, I will take a look

WillKoehrsen (approved):
Tests pass locally with 

```bash
VCR_RECORD_MODE=all poetry run pytest gridstatus/tests/source_specific/test_caiso.py -k test_get_load_forecast -vvv
```

Kladar (approved):
Tests passing for me locally as well. Might have been a VCR re-record issue (that VCR_RECORD_MODE=ALL seems to be a nice fallback for that I use occasionally as well). I think we're good to merge here.

PR #579, opened 2025-04-18
feat(ENG 984): IESO Forecast Surplus Baseload

Kladar:
## Summary
Adds the `ieso_forecast_surplus_baseload_generation`. 

```
ieso = IESO()
df = ieso.get_forecast_surplus_baseload_generation(date="2025-04-17")
df2 = ieso.get_forecast_surplus_baseload_generation(date="latest")
print(df)
print(df2)
```

## Details
This gets a little funky since the URL is asking for publish time. It's simpler than resource adequacy in that they just publish one report per day and it does not get updated, but If you wanted all the forecasts for a given interval, (say, `2025-04-18 01:00`) you'd have to query the 10 days of preceding reports and filter them. Not the end of the world, but might be a bit unintuitive. 

According to the spec, the columns are up to the developer a bit, so I take a crack at it but could use @cwaldoch 's final opinion on it. 

They are:

```
"Interval Start",
 "Interval End",
 "Publish Time",
 "Surplus Baseload MW",
 "Surplus State",
 "Action",
 "Export Forecast MW",
 "Minimum Generation Status",
```
Example report is here: https://www.ieso.ca/-/media/Files/IESO/uploaded/sbg/PUB_SurplusBaseloadGen_20250417_v1
And I've attached a json version here, which converts the xml to a json [surplus_forecast.json](https://github.com/user-attachments/files/19815216/surplus_forecast.json)

Let me know if there's a different approach I should use. There are probably a fair number of ways to make this work.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -2171,7 +2171,9 @@ class IESO(ISOBase):
                 else:
                     surplus_state = SurplusState.NO_SURPLUS.value
 
-                interval_start = date_forecast + pd.Timedelta(hours=hour - 1)
+                interval_start = (
+                    date_forecast + pd.Timedelta(hours=hour - 1)
+                ).tz_localize(self.default_timezone)
                 interval_end = interval_start + pd.Timedelta(hours=1)
 
                 data.append(
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -5,6 +5,7 @@ import re
 import time
 import xml.etree.ElementTree as ET
 from concurrent.futures import ThreadPoolExecutor, as_completed
+from enum import Enum
 from typing import Literal
 from urllib.error import HTTPError
 
@@ -104,6 +105,22 @@ IESO_ZONE_MAPPING = {
     "PQQC": "Quebec Q4C",
     "PQXY": "Quebec X2Y",
 }
+
+
+class SurplusState(str, Enum):
+    """Enum for surplus baseload generation states.
+
+    The state is determined by the Action field in the report:
+    - No Action (empty/null) -> No Surplus (white)
+    - Other -> Managed with exports/curtailments/VG dispatch (green)
+    - Manoeuvre -> Potential to dispatch nuclear units (yellow)
+    - Shutdown -> Potential to shutdown nuclear units (red)
+    """
+
+    NO_SURPLUS = "No Surplus"
+    MANAGED_WITH_EXPORTS = "Managed with Exports"
+    NUCLEAR_DISPATCH = "Nuclear Dispatch"
+    NUCLEAR_SHUTDOWN = "Nuclear Shutdown"
 
 
 class IESO(ISOBase):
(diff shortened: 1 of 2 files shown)


gridstatus/ieso.py
```diff
@@ -2079,3 +2096,110 @@ def _extract_hourly_values(
...
+                "Interval Start",
+                "Interval End",
+                "Publish Time",
+                "Surplus Baseload MW",
+                "Surplus State",
+                "Action",
+                "Export Forecast MW",
+                "Minimum Generation Status",
+            ]
+        ]
```
Kladar, line 2205: Just asking @cwaldoch to review the column choices here (`Action` being requested originally, and `Surplus State` falling out of the report a bit).

gridstatus/ieso.py
```diff
@@ -106,6 +107,22 @@
...
 
 
+class SurplusState(str, Enum):
+    """Enum for surplus baseload generation states.
+
+    The state is determined by the Action field in the report:
+    - No Action (empty/null) -> No Surplus (white)
+    - Other -> Managed with exports/curtailments/VG dispatch (green)
+    - Manoeuvre -> Potential to dispatch nuclear units (yellow)
+    - Shutdown -> Potential to shutdown nuclear units (red)
```
WillKoehrsen, line 117: How were you able to figure these out? I don't see that mapping in an [example report](https://www.ieso.ca/-/media/Files/IESO/uploaded/sbg/PUB_SurplusBaseloadGen_20250420_v1).

gridstatus/ieso.py
```diff
@@ -2079,3 +2096,110 @@ def _extract_hourly_values(
...
+
+        df = pd.DataFrame(data)
+        df.sort_values("Interval Start", inplace=True)
+        df.reset_index(drop=True, inplace=True)
+
+        return df[
+            [
+                "Interval Start",
+                "Interval End",
+                "Publish Time",
```
WillKoehrsen, line 2198: In the returned dataframe, `Interval Start` and `Interval End` don't have a timezone, but `Publish Time` does. I think they should all have the local timezone.

```ipython
In [8]: df.dtypes
Out[8]: 
Interval Start                    datetime64[ns]
Interval End                      datetime64[ns]
Publish Time                 datetime64[ns, EST]
Surplus Baseload MW                      float64
Surplus State                             object
Action                                    object
Export Forecast MW                       float64
Minimum Generation Status                 object
dtype: object
```

gridstatus/ieso.py
Kladar: This came from some investigation Connor did on the XML report, in the Data Spec attached to the ticket.
> The colors are mapped corresponding Action values to in the XML via the XSL (snip below), which are Manoeuvre, Shutdown, and Other. 

CleanShot 2025-04-21 at 10 28 01@2x

And so color is linked to the `Action` field, and then `Action` links to description of `Surplus State` in the report.

gridstatus/ieso.py
Kladar: nice, agreed, added a localization.

gridstatus/ieso.py
WillKoehrsen: Nice. That's some solid investigation.

WillKoehrsen (approved):
Looks good. May want Connor to give approval on columns before merging.

PR #583, opened 2025-04-22
Reduce Error Logging from CAISO

WillKoehrsen:
## Summary

- We currently log two errors every single time we are rate limited by CAISO. This is excessive and we don't need to log an error in this situation because we are aware of the problem
- This PR changes the error to an info log and adds a warning if we exceed our retries and are rate limited

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -331,6 +331,7 @@ class CAISO(ISOBase):
         raw_data: bool = False,
         verbose: bool = False,
         sleep: int = 5,
+        max_retries: int = 3,
     ) -> pd.DataFrame | None:
         start, end = _caiso_handle_start_end(start, end)
         config = copy.deepcopy(config)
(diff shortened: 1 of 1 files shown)


gridstatus/caiso/caiso.py
```diff
@@ -346,16 +347,23 @@ def _get_oasis(
...
-        while retry_num < 3:
+        while retry_num < max_retries:
             r = requests.get(url)
 
             if r.status_code == 200:
                 break
 
             retry_num += 1
-            logger.error(f"Failed to get data from CAISO. Error: {r.status_code}")
-            logger.error(f"Retrying {retry_num}...")
```
WillKoehrsen, line 357: This doesn't need to be an error, and we definitely don't need to emit two errors every time we are rate limited. Sentry interprets error logs as errors which causes us to quickly exhaust our error budget.

Kladar (approved):

WillKoehrsen:
I haven't merged yet because I want to make sure this works. The problem is that I can't recreate the 429 errors we see.

PR #585, opened 2025-04-24
[ENG-1551] IESO Transmission Outages

WillKoehrsen:
## Summary

- Adds `IESO().get_transmission_outages_planned`

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -1,3 +1,5 @@
+# flake8: noqa: E501
+
 import datetime
 import http.client
 import os
@@ -15,7 +17,7 @@ import xmltodict
 from bs4 import BeautifulSoup
 
 from gridstatus import utils
-from gridstatus.base import ISOBase, NotSupported
+from gridstatus.base import ISOBase, NoDataFoundException, NotSupported
 from gridstatus.decorators import support_date_range
 from gridstatus.gs_logging import logger
 from gridstatus.ieso_constants import (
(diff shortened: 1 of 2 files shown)


gridstatus/ieso.py
```diff
@@ -2615,3 +2615,105 @@ def get_lmp_day_ahead_hourly_ontario_zonal(
...
+        # There will be overlap between the reports so we need to drop duplicates,
+        # keeping the latest publish time
+        data = data.sort_values(["Publish Time"], ascending=[False])
+        data = data.drop_duplicates(
+            subset=[c for c in data.columns if c != "Publish Time"],
+        )
+
+        data = data.sort_values(
+            ["Interval Start", "Publish Time", "Outage ID"],
+        ).reset_index(drop=True)
```
Kladar, line 2717: Been looking at this for a bit since it feels weird to sort twice, have list generator in subset, `ascending` taking a list with one value, etc haha. But it works.

could probably do a 

```suggestion
        data = data.drop_duplicates(
            subset=[c for c in data.columns if c != "Publish Time"], keep="last"
        )

        data = data.sort_values(
            ["Interval Start", "Publish Time", "Outage ID"],
        ).reset_index(drop=True)
```
to remove the extra sort, but yeah, it works as-is.

Kladar (approved):
Data flows and tests pass. One pythonicity comment only because I looked at that snippet for so long lol

gridstatus/ieso.py
WillKoehrsen: Thanks, I'll use this. One problem is we don't know if there will be duplicates when the actual values go live. The current sandbox data has many duplicates which is why we drop them here. We'll have to monitor once the data is live.

PR #586, opened 2025-04-24
fix(ENG 983): Explicit columns for IESO Intertie Actuals

Kladar:
## Summary
Fixes the columns in the output df


gridstatus/ieso.py
```diff
@@ -2284,7 +2284,58 @@ def get_intertie_actual_schedule_flow_hourly(
             target_date = pd.Timestamp(date).date()
             df = df[df["Interval Start"].dt.date == target_date]
 
-        return df
+        return df[
```
WillKoehrsen, line 2287: Should we create an `ieso_constants` file and put this there? Would make the code more readable.

Kladar: agreed, added constants file and moved some things out

WillKoehrsen: Thanks.

WillKoehrsen (approved):
IESO tests passing locally.

PR #587, opened 2025-04-27
[ENG-1550] IESO Transmission Limits

WillKoehrsen:
## Summary

- Adds 
  - `IESO().get_in_service_transmission_limits`
  - `IESO().get_outage_transmission_liimts`

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -15,6 +15,7 @@ import pandas as pd
 import requests
 import xmltodict
 from bs4 import BeautifulSoup
+from lxml import etree as lxml_etree
 
 from gridstatus import utils
 from gridstatus.base import ISOBase, NoDataFoundException, NotSupported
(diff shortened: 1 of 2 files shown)


gridstatus/ieso.py
```diff
@@ -2722,3 +2723,128 @@ def get_transmission_outages_planned(
...
+            url = f"{PUBLIC_REPORTS_URL_PREFIX}/TxLimitsAllInService0to34Days/PUB_TxLimitsAllInService0to34Days.xml"
+            date = pd.Timestamp.now(tz=self.default_timezone)
+        else:
+            url = f"{PUBLIC_REPORTS_URL_PREFIX}/TxLimitsAllInService0to34Days/PUB_TxLimitsAllInService0to34Days_{date.strftime('%Y%m%d')}.xml"
+
+        xml_content = self._request(url, verbose).text
+
+        return self._process_transmission_limits(
+            xml_content,
+        )
```
Kladar, line 2744: It looks like there is a versioning system for these files. Do we need to pick up the versioned files at all? The current versioned and unversioned files are identical though not sure that always holds true. We're probably good, but just noting that. 

CleanShot 2025-04-28 at 08 10 10@2x

https://reports-public-sandbox.ieso.ca/public/TxLimitsAllInService0to34Days/

WillKoehrsen: We'll have to wait and see if the versioned reports differ. I'm assuming this report is like other IESOs reports where the latest versioned file is identical to the unversioned file.

gridstatus/ieso.py
```diff
@@ -2722,3 +2723,128 @@ def get_transmission_outages_planned(
...
+            str(tree.xpath("//ns:CreatedAt/text()", namespaces=ns)[0]),
+        ).tz_localize(
+            self.default_timezone,
+        )
+
+        data = []
+
+        facility_types = {"Internal": "Internal", "Intertie": "Intertie"}
+
+        for xml_type, output_type in facility_types.items():
```
Kladar, line 2795: do we need a dictionary with identical keys and values? 

```suggestion

        for facility_type in ["Internal", "Intertie"]:
```

and then all the references in the loop can be simply `facility_type`.

Kladar (approved):
All a bit WIP until the rubber meets the road on Thursday

gridstatus/ieso.py
WillKoehrsen: Whoops, these used to have different keys and values. Updated to use a list.

PR #588, opened 2025-04-28
feat(ENG 1548): IESO Zonal Load

Kladar:
## Summary
Adds the `ieso_get_load_zonal_hourly` and the `get_load_zonal_5_min` datasets. These are daily published datasets, so we will likely add some more real-time datasets as they become available. 

```
ieso = IESO()

df = ieso.get_load_zonal_hourly(date="latest")
df2 = ieso.get_load_zonal_hourly(date="2025-04-21")
print(df) 
print(df2)

df3 = ieso.get_load_zonal_5_min(date="latest")
df4 = ieso.get_load_zonal_5_min(date="2025-04-21")
print(df3)
print(df4)
```

### Details
I don't grab versioned files since these are actuals. Since the files return the full year of data, I do some filtering down to what is relevant for the query. We could return the whole year. I might also have the `@support_date_range` frequency wrong here.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -2745,7 +2745,7 @@ class IESO(ISOBase):
             df["Date"]
             + pd.to_timedelta(df["Hour"] - 1, unit="h")
             + pd.to_timedelta((df["Interval"] - 1) * 5, unit="m")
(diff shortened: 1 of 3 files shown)


WillKoehrsen (approved):
Looks good and tests pass.

PR #592, opened 2025-04-30
IESO Market Launch URL

WillKoehrsen:
## Summary

- Update the IESO public reports URL for the market launch


WillKoehrsen:
Tests are not going to pass since the new data hasn't been released yet.

WillKoehrsen:
Some tests are still failing, probably because IESO hasn't started releasing all the new data to production. I will merge and we can fix issues if they arise.

PR #593, opened 2025-05-01
IESO Real Time Totals

WillKoehrsen:
## Summary

- Adds `IESO().get_real_time_totals`
- Adds a warning to datasets that have been retired by IESO

### DIFF
--- a/gridstatus/base.py
+++ b/gridstatus/base.py
@@ -22,6 +22,10 @@ class NoDataFoundException(Exception):
     pass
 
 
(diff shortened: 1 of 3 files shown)


Kladar (approved):
`latest` works as does the earliest available timestamp "2025-05-01 04:00:00"


FWIW I do get a not found exception on this
```
ieso = IESO()
df = ieso.get_real_time_totals("2025-05-01")
```
which I think is just the hour start date range decorator behavior with an incomplete first day of data.

PR #595, opened 2025-05-01
feat(ENG 1549): IESO Renewable forecasts

Kladar:
## Summary
Adds the variable generation datasets: `solar_embedded_forecast`, `solar_market_participant_forecast`, `wind_embedded_forecast`, and `wind_market_participant_forecast` following the data modeling discussion. 

```
df = ieso.get_solar_embedded_forecast("2025-05-01", vintage="all")
df2 = ieso.get_solar_market_participant_forecast("2025-05-01", vintage="all")
df3 = ieso.get_wind_embedded_forecast("2025-05-01", vintage="all")
df4 = ieso.get_wind_market_participant_forecast("2025-05-01", vintage="all")

print(df)
print(df.columns)

print(df2)
print(df2.columns)

print(df3)
print(df3.columns)

print(df4)
print(df4.columns)
```

### Details
[Here's an example report](https://reports-public.ieso.ca/public/VGForecastSummary/PUB_VGForecastSummary_20250501_v15.xml).

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -3069,7 +3069,9 @@ class IESO(ISOBase):
         ]
         df = pd.concat(dfs).reset_index(drop=True)
         df.drop_duplicates(inplace=True)
-        df = df[df["Organization Type"] == "Embedded" & df["Type"] == "Solar"]
+        df = df[
+            (df["Organization Type"] == "Embedded") & (df["Type"] == "Solar")
+        ].reset_index(drop=True)
         df.drop(columns=["Organization Type", "Type"], inplace=True)
         return df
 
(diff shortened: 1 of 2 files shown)


WillKoehrsen (approved):
Looks good and runs as expected locally. I do wonder if we need both "Publish Time" and "Last Modified". The times seem very similar and I think they may cause confusion.

Kladar:
> Looks good and runs as expected locally. I do wonder if we need both "Publish Time" and "Last Modified". The times seem very similar and I think they may cause confusion.

Yeah makes sense; I go back and forth on this as well. I tend to keep it for paper trail reasons, since I think of "Publish Time" as useful for forecast analysis in figuring out what information the forecast generation process had when it made the forecast, and "Last Modified" as useful to know which report the data came from, in case there is a reason to go chase down an issue. Forecasts stretch my brain.

PR #596, opened 2025-05-04
[ENG-1584] IESO Predispatch Prices

WillKoehrsen:
## Summary

- Adds IESO Predispatch prices
  - `IESO().get_lmp_predispatch_hourly`
  - `IESO().get_lmp_predispatch_hourly_intertie`
  - `IESO().get_lmp_predispatch_hourly_ontario_zonal`
  - `IESO().get_lmp_predispatch_hourly_virtual_zonal`

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -2080,7 +2080,7 @@ class IESO(ISOBase):
 
             file_data = self._get_lmp_csv_data(
                 url,
-                delivery_date=delivery_date,
+                base_date=delivery_date,
                 minutes_per_interval=60,
                 verbose=verbose,
             )
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -14,6 +14,7 @@ from xml.etree import ElementTree
 
 import pandas as pd
 import requests
+import tqdm
 import xmltodict
 from bs4 import BeautifulSoup
 from lxml import etree as lxml_etree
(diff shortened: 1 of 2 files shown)


WillKoehrsen:
Note the "Publish Time" in these datasets because there are multiple versions of the prices published each day. I'm not sure if we want to call this something else like "Dispatch Time". 

Note also that this does not use the "Last modified" time from the directory. Instead, it uses the CreatedAt time from the file itself to be more accurate. (The "Last modified" times in the directory are rounded times and I think we should keep the seconds in case we need to differentiate between two different files published at the same minute but different seconds).

Finally, I tried to reduce code duplication by re-using methods from the day ahead LMPs. However, there were subtle differences between the day-ahead and predispatch datasets that required the use of flags to change function behavior.

WillKoehrsen:
The IESO tests are failing because there's not yet enough production data for historical tests.

WillKoehrsen:
Tests are now passing with the production URL since there is enough past data.

gridstatus/ieso.py
```diff
@@ -2003,42 +2006,145 @@ def get_lmp_day_ahead_hourly(
...
+
+        return data
+
+    def _get_lmp_csv_data(
         self,
         url: str,
-        date: pd.Timestamp,
-        interval_duration: str = "hour",
+        base_date: pd.Timestamp,
         minutes_per_interval: int = 60,
```
Kladar, line 2126: since only valid values are 5 and 60 I would probably call this a
```suggestion
        minutes_per_interval: Literal[5, 60] = 60,
```

gridstatus/ieso.py
```diff
@@ -2610,18 +2872,92 @@ def get_lmp_day_ahead_hourly_ontario_zonal(
...
+                subset=["Interval Start", "Location", "Publish Time"],
+                keep="last",
+            )
+            .sort_values(["Interval Start", "Location", "Publish Time"])
+            .reset_index(drop=True)
+        )
+
+        return data
+
+    def _process_lmp_hourly_ontario_zonal(
```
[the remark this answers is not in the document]
WillKoehrsen: Yeah. We still add a `location` column for consistency with our other LMP datasets.

gridstatus/ieso.py
WillKoehrsen: Good idea, updated.

WillKoehrsen:
> we're doing several different xml parsing approaches in this file; might be something to simplify in the future.

This PR re-used parsing when feasible. I agree there's more code reuse that could be done for these XML reports.

PR #597, opened 2025-05-05
feat(ENG 1545): IESO Constraint Shadow Prices

Kladar:
## Summary
[DRAFT UNTIL SITE LINK IS LIVE]
Adds the `ieso_shadow_prices_real_time_5_min` dataset. 

```
ieso = IESO()

df = ieso.get_shadow_prices_real_time_5_min(date="latest")
df2 = ieso.get_shadow_prices_real_time_5_min(date="2025-04-02")

print(df)
print(df2)
```

### Details
Currently these are not updating on the `public/` reports site, but they should hit any time now since they are on a 6 day lag. 

Prod: https://reports-public.ieso.ca/public/RealtimeConstrShadowPrices/
Sandbox: https://reports-public-sandbox.ieso.ca/public/RealtimeConstrShadowPrices/

Bit of a heady report since each report contains only one hour of 5 minute data, with the version number being seemingly synonymous with the hour (I don't actually use the version number for anything, but it could be that each hour gets more than one version/each day gets more than 24 versions). It could be that it has a more complex updating regime but we'll have to see how it actually updates once it goes live. This should get us 95% of the way there though. 

The code follows a similar pattern as adequacy report, getting the version files (w/ multiprocessing and retry logic), parsing, etc. For now it repeats that logic, but definitely starting to see a more general pattern that we could apply to some of these types of report getter methods.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -3049,3 +3049,186 @@ class IESO(ISOBase):
         )
 
         return data
+
+    @support_date_range(frequency="HOUR_START")
+    def get_shadow_prices_real_time_5_min(
+        self,
+        date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
+        end: pd.Timestamp | None = None,
+        verbose: bool = False,
+        last_modified: str | pd.Timestamp | None = None,
+    ) -> pd.DataFrame:
+        if last_modified:
+            last_modified = utils._handle_date(last_modified, tz=self.default_timezone)
+        if date == "latest":
+            # TODO(kladar): Remove this once the link is available
+            # base_url = f"{PUBLIC_REPORTS_URL_PREFIX}/RealtimeConstrShadowPrices"
+            base_url = "https://reports-public-sandbox.ieso.ca/public/RealtimeConstrShadowPrices"
+            file = "PUB_RealtimeConstrShadowPrices.xml"
+            r = self._request(base_url)
+            file_last_modified = pd.Timestamp(
+                re.search(
+                    r'<a href="PUB_RealtimeConstrShadowPrices\.xml">.*?</a>\s+(\d{2}-\w{3}-\d{4} \d{2}:\d{2})',
+                    r.text,
+                ).group(1),
+                tz=self.default_timezone,
+            )
+            json_data = self._fetch_and_parse_shadow_prices_file(base_url, file)
+            df = self._parse_shadow_prices_report(json_data)
+            df["Last Modified"] = file_last_modified
+            df.sort_values(
+                ["Interval Start", "Publish Time", "Last Modified", "Constraint"],
+                inplace=True,
+            )
+            return df[
+                [
+                    "Interval Start",
+                    "Interval End",
+                    "Publish Time",
+                    "Last Modified",
+                    "Constraint",
+                    "Shadow Price",
+                ]
+            ].reset_index(drop=True)
+
+        json_data_with_times = self._get_all_shadow_prices_jsons(date, last_modified)
+        dfs = []
+        for json_data, file_last_modified in json_data_with_times:
+            df = self._parse_shadow_prices_report(json_data)
+            df["Last Modified"] = file_last_modified
+            dfs.append(df)
+        df = pd.concat(dfs)
+        df = utils.move_cols_to_front(
+            df,
+            ["Interval Start", "Interval End", "Publish Time", "Last Modified"],
+        )
+        df.sort_values(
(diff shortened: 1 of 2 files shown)


gridstatus/tests/source_specific/test_ieso.py
```diff
@@ -1819,3 +1819,42 @@ def test_get_real_time_totals_historical_date_range(self):
...
+        assert self._check_is_datetime_type(df["Last Modified"])
+        assert df["Shadow Price"].dtype == "float64"
+
+    def test_get_shadow_prices_real_time_5_min_latest(self):
+        with file_vcr.use_cassette(
+            "test_get_shadow_prices_real_time_5_min_latest.yaml",
+        ):
+            df = self.iso.get_shadow_prices_real_time_5_min("latest")
+            self._check_shadow_prices(df)
+            assert df["Interval Start"].is_monotonic_increasing
```
Kladar, line 1846: i like `is_monotonic_increasing`

gridstatus/tests/source_specific/test_ieso.py
```diff
@@ -1819,3 +1819,42 @@ def test_get_real_time_totals_historical_date_range(self):
...
+
+    @pytest.mark.parametrize(
+        "date, end",
+        [
+            ("2025-05-01", "2025-05-03"),
+        ],
+    )
+    def test_get_shadow_prices_real_time_5_min_historical_range(self, date, end):
```
Kladar, line 1854: This will fail until Friday when there's enough data. I have another PR forthcoming to fix IESO tests more generally

gridstatus/ieso.py
```diff
@@ -3049,3 +3049,180 @@ def get_real_time_totals(
...
+            df.sort_values(
+                ["Interval Start", "Publish Time", "Last Modified", "Constraint"],
+                inplace=True,
+            )
+            return df[
+                [
+                    "Interval Start",
+                    "Interval End",
+                    "Publish Time",
+                    "Last Modified",
```
WillKoehrsen, line 3086: Do we need both the publish time and last modified? For the files I've seen, they seem very close to each other, generally within 30 minutes.

gridstatus/ieso.py
```diff
@@ -3049,3 +3049,180 @@ def get_real_time_totals(
...
+        pattern = '<a href="({}.*?.xml)">.*?</a>\\s+(\\d{{2}}-\\w{{3}}-\\d{{4}} \\d{{2}}:\\d{{2}})'
+        file_rows = re.findall(pattern.format(file_prefix), r.text)
+        if not file_rows:
+            raise FileNotFoundError(f"No shadow price files found for date {date_str}")
+        if last_modified:
+            if last_modified.tz is None:
+                last_modified = utils._handle_date(
+                    last_modified,
+                    tz=self.default_timezone,
+                )
```
WillKoehrsen, line 3151: Don't we already put `last_modified` through `_handle_date` in the calling function?

https://github.com/gridstatus/gridstatus/pull/597/files#diff-c8d15e6c82b2b163c7c78770947c46f0f4ad0733f2a98db9b190617f3e97c21cR3061-R3062

gridstatus/ieso.py
Kladar: yeah since this is not a forecast per se I'm good with dropping it

gridstatus/ieso.py
Kladar: alas we do. Updated.

gridstatus/ieso.py
```diff
@@ -3702,3 +3702,194 @@ def _parse_variable_generation_forecast(
...
+            base_url = f"{PUBLIC_REPORTS_URL_PREFIX}/RealtimeConstrShadowPrices"
+            file = "PUB_RealtimeConstrShadowPrices.xml"
+            r = self._request(base_url)
+            file_last_modified = pd.Timestamp(
+                re.search(
+                    r'<a href="PUB_RealtimeConstrShadowPrices\.xml">.*?</a>\s+(\d{2}-\w{3}-\d{4} \d{2}:\d{2})',
+                    r.text,
+                ).group(1),
+                tz=self.default_timezone,
+            )
```
WillKoehrsen, line 3726: When `date == "latest"`, can't we just request this file: https://reports-public.ieso.ca/public/RealtimeConstrShadowPrices/PUB_RealtimeConstrShadowPrices.xml?

Kladar: ah yeah since we don't need last modified. updated.

WillKoehrsen (approved):
👍🏽

PR #600, opened 2025-05-10
[ENG-1613] IESO Intertie Schedule and Flows Updated

WillKoehrsen:
## Summary

- Renames the existing `get_intertie_actual_schedule_flow_hourly` to `get_yearly_intertie_actual_schedule_flow_hourly`
- Adds an updated `get_intertie_actual_schedule_flow_hourly` that uses a more frequently updated XML file
- Adds a new `get_intertie_flow_5_min` that uses the same XML file

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -1797,7 +1797,7 @@ class IESO(ISOBase):
             target_date = pd.Timestamp(date).date()
             df = df[df["Interval Start"].dt.date == target_date]
 
-        return df[INTERTIE_ACTUAL_SCHEDULE_FLOW_HOURLY_COLUMNS]
+        return df[INTERTIE_ACTUAL_SCHEDULE_FLOW_HOURLY_COLUMNS].reset_index(drop=True)
 
     def _get_intertie_schedule_flow_data(
         self,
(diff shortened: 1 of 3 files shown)


WillKoehrsen:
> One comment about reset index

I might have just missed this, but I didn't see a comment on this?

gridstatus/ieso.py
```diff
@@ -1712,15 +1713,17 @@ def get_forecast_surplus_baseload_generation(
         ]
 
     @support_date_range(frequency="YEAR_START")
-    def get_intertie_actual_schedule_flow_hourly(
+    def get_yearly_intertie_actual_schedule_flow_hourly(
```
Kladar, line 1716: Ope, not sure why this didn't get picked up in the review. 

Might be worth adding a `reset_index()` to the df return here 
CleanShot 2025-05-12 at 10 38 56

Kladar:
> > One comment about reset index
> 
> I might have just missed this, but I didn't see a comment on this?

Oh weird, review didn't bring it in. Added

gridstatus/ieso.py
WillKoehrsen: Good catch, updated.

PR #601, opened 2025-05-12
feat(ENG 1593): PJM Load Forecast 5 Min

Kladar:
## Summary
Adds the `pjm_load_forecast_5_min` dataset

```
pjm = PJM()
df = pjm.get_load_forecast_5_min("2025-05-12")
print(df)
print(df.dtypes)
```

### Details
Pretty straighforward, going for the same columns as the hourly forecast. 

Also added a make command to make running single source-specific tests easier. Could probably go about that in all kinds of ways but bout time I simplified that a bit. 

```
$ make test-one-off market=pjm test=test_get_load_forecast_5_min_historical_range
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,20 @@ test-unit:
 	uv pip install vcrpy
 	$(PYTEST_CMD) $(UNIT_ONLY)
 
+.PHONY: test-one-off
+test-one-off:
+ifndef market
+	$(error market parameter is required. Usage: make test-one-off market=MARKET test=TEST_NAME)
+endif
+ifndef test
+	$(error test parameter is required. Usage: make test-one-off market=MARKET test=TEST_NAME)
+endif
+ifeq ($(market),ercot)
+	uv run pytest gridstatus/tests/source_specific/test_$(market).py::TestErcot::$(test)
+else
+	uv run pytest gridstatus/tests/source_specific/test_$(market).py::Test$(shell echo $(market) | tr '[:lower:]' '[:upper:]')::$(test)
+endif
+
 .PHONY: installdeps-dev
 installdeps-dev:
 	uv sync
--- a/Makefile
+++ b/Makefile
@@ -68,18 +68,25 @@ test-unit:
 	uv pip install vcrpy
 	$(PYTEST_CMD) $(UNIT_ONLY)
 
+#TODO(kladar): Maybe gridstatus gets a CLI in the future for things like this.
+#NB: This is for running a single test without having to fully specify the
+#    path and the pytest command.
+#    Usage: make test-one-off market=MARKET test=TEST_NAME
+#    Example: make test-one-off market=pjm test=test_get_load_forecast_5_min_latest
 .PHONY: test-one-off
 test-one-off:
+#NB: First two blocks are for error checking.
 ifndef market
 	$(error market parameter is required. Usage: make test-one-off market=MARKET test=TEST_NAME)
 endif
 ifndef test
 	$(error test parameter is required. Usage: make test-one-off market=MARKET test=TEST_NAME)
 endif
+#NB: This puts the market in the right capitalization for the path. Ercot is not ALL CAPS like the other market test classes: TestErcot vs TestCAISO, etc.
 ifeq ($(market),ercot)
-	uv run pytest gridstatus/tests/source_specific/test_$(market).py::TestErcot::$(test)
+	uv run pytest -vvv gridstatus/tests/source_specific/test_$(market).py::TestErcot::$(test)
 else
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -68,6 +68,20 @@ test-unit:
...
+	$(error market parameter is required. Usage: make test-one-off market=MARKET test=TEST_NAME)
+endif
+ifndef test
+	$(error test parameter is required. Usage: make test-one-off market=MARKET test=TEST_NAME)
+endif
+ifeq ($(market),ercot)
+	uv run pytest gridstatus/tests/source_specific/test_$(market).py::TestErcot::$(test)
+else
+	uv run pytest gridstatus/tests/source_specific/test_$(market).py::Test$(shell echo $(market) | tr '[:lower:]' '[:upper:]')::$(test)
+endif
```
WillKoehrsen, line 83: Could you maybe add a few comments to explain the logic here and why Ercot is treated differently.

gridstatus/pjm.py
```diff
@@ -273,6 +274,42 @@ def get_load_forecast_historical(
...
+                verbose=verbose,
+            )
+
+        params = {
+            "fields": (
+                "evaluated_at_utc,forecast_datetime_beginning_utc,forecast_datetime_ending_utc,forecast_area,forecast_load_mw"
+            ),
+        }
+
+        filter_timestamp_name = "forecast_datetime_beginning"
```
WillKoehrsen, line 300: Something odd is happening with the filtering. In my tests, there is too much data returned at the start of the range.

```python

data =PJM().get_load_forecast_5_min('2025-05-10T00:00:00Z', '2025-05-10T02:00:00Z')
data['Publish Time'].min().tz_convert('UTC')

> Timestamp('2025-05-09 22:05:00+0000', tz='UTC')

data['Publish Time'].max().tz_convert('UTC')

> Timestamp('2025-05-10 02:00:00+0000', tz='UTC')
```

Makefile
Kladar: Added some explainers. We might have other ways to do this, but found myself writing paths to specific tests when I wanted to run them and so this makes that a bit easier. Not meant for CI or anything.

WillKoehrsen: Thanks!

gridstatus/pjm.py
Kladar: investigating

WillKoehrsen: Probably not worth worrying about since we're getting more data than expected and we are passing the correct parameters to the API.

gridstatus/pjm.py
```diff
@@ -273,6 +274,42 @@ def get_load_forecast_historical(
...
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Load forecast made today extending for six days in 5 minute intervals.
+        """
+        if date == "latest":
+            return self.get_load_forecast_5_min(
+                # pd.Timestamp.now(tz=self.default_timezone).floor("5min"),
```
WillKoehrsen, line 289: Nit: remove before merging.

gridstatus/pjm.py
```diff
@@ -273,6 +274,42 @@ def get_load_forecast_historical(
...
 
+    @support_date_range(frequency=None)
+    def get_load_forecast_5_min(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Load forecast made today extending for six days in 5 minute intervals.
```
WillKoehrsen, line 285: Nit: I think this forecast only extends 2 hours.

> This feed contains the Five-minute load forecast data showing forecast amounts every five minutes for a two hour window by area

gridstatus/pjm.py
Kladar: +1, done

gridstatus/pjm.py
Kladar: my copy/paste skills were not on yesterday it seems. done.

gridstatus/pjm.py
Kladar: Oh this is because it is queried on the interval rather than the publish time. So it's returning the correct intervals IMO

WillKoehrsen: Should we be filtering on `evaluated_at`? That's what we do for other PJM forecasts.

Kladar: ah got it, yeah if we do that for the other methods then I'll do that here. updated.

WillKoehrsen (approved):

PR #602, opened 2025-05-12
feat(ENG 1618): IESO Shadow Prices Day Ahead Hourly

Kladar:
## Summary
Adds the `ieso_shadow_prices_day_ahead_hourly` dataset.

```
df = ieso.get_shadow_prices_day_ahead_hourly(date="2025-05-01")
print(df)
print(df.dtypes)
print(df.columns)
```

### Details
Same structure as the real time from last week.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -3965,7 +3965,7 @@ class IESO(ISOBase):
             base_url = f"{PUBLIC_REPORTS_URL_PREFIX}/RealtimeConstrShadowPrices"
             file = "PUB_RealtimeConstrShadowPrices.xml"
             json_data = self._fetch_and_parse_shadow_prices_file(base_url, file)
-            df = self._parse_shadow_prices_report(json_data)
+            df = self._parse_real_time_shadow_prices_report(json_data)
             df.sort_values(
                 ["Interval Start", "Publish Time", "Constraint"],
                 inplace=True,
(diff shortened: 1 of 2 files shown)


gridstatus/ieso.py
```diff
@@ -4010,6 +4010,64 @@ def get_shadow_prices_real_time_5_min(
             ]
         ].reset_index(drop=True)
 
+    @support_date_range(frequency="DAY_START")
+    def get_shadow_prices_day_ahead_hourly(
```
WillKoehrsen, line 4014: Why is the day-ahead method fetching real time files?

Screenshot 2025-05-13 at 11 47 54

Kladar: oh woops, I only modified the base url in the `latest` code path, was thinking they had the same xml structure but it's slightly different.

Working now. I could probably refactor and combine some methods (or even go the other way and keep them fully separate end-to-end), but hopefully this strikes a decent balance.

gridstatus/ieso.py
```diff
@@ -4019,15 +4081,19 @@ def _fetch_and_parse_shadow_prices_file(self, base_url: str, file: str) -> dict:
...
+            base_url = f"{PUBLIC_REPORTS_URL_PREFIX}/DAConstrShadowPrices"
 
         if isinstance(date, (datetime.datetime, datetime.date)):
             date_str = date.strftime("%Y%m%d")
         else:
             date_str = date.replace("-", "")
-        file_prefix = f"PUB_RealtimeConstrShadowPrices_{date_str}"
+        file_prefix = f"PUB_{market}ConstrShadowPrices_{date_str}"
         r = self._request(base_url)
         pattern = '<a href="({}.*?.xml)">.*?</a>\\s+(\\d{{2}}-\\w{{3}}-\\d{{4}} \\d{{2}}:\\d{{2}})'
```
Kladar, line 4098: This also leaves a little bit to be desired, but I think a bigger refactor of this file with some IESO class utilities would obviate that, which will be possible once I get the tests back to good.

gridstatus/ieso.py
```diff
@@ -4087,7 +4153,34 @@ def _get_all_shadow_prices_jsons(
                         break
         return json_data_with_times
 
-    def _parse_shadow_prices_report(self, json_data: dict) -> pd.DataFrame:
```
WillKoehrsen, line 4090: This old method name is still referenced twice in this file.

Kladar: yar, updated

WillKoehrsen (approved):
👍🏽

PR #604, opened 2025-05-21
[ENG-1657] ERCOT DAM Total Energy Purchased and Sold

WillKoehrsen:
## Summary

- Adds `Ercot().get_dam_total_energy_purchased()
- Adds `Ercot().get_dam_total_energy_sold()

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
     hooks:
       - id: validate-pyproject
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.6.8
+    rev: v0.11.5
     hooks:
       - id: ruff
         args: ["--fix"]
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -3543,9 +3543,9 @@ class Ercot(ISOBase):
                 # Sometimes ERCOT handles DST end by putting 25 hours in HourEnding
                 # which makes IntervalStart where HourEnding >= 3 an hour later than
                 # they should be. We correct this by subtracting an hour.
-                assert (
-                    doc["HourEnding"].max() == 25
-                ), f"Time parsing error. Did not find HourEnding = 25. {e}"
+                assert doc["HourEnding"].max() == 25, (
+                    f"Time parsing error. Did not find HourEnding = 25. {e}"
+                )
                 doc.loc[doc["HourEnding"] >= 3, "Interval Start"] = doc.loc[
                     doc["HourEnding"] >= 3,
                     "Interval Start",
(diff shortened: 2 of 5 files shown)


gridstatus/ercot.py
```diff
@@ -3658,3 +3664,89 @@ def _handle_indicative_lmp_by_settlement_point(self, df: pd.DataFrame):
...
+            )
+
+        # DAM data so subtract one from the date
+        doc = self._get_document(
+            report_type_id=DAM_TOTAL_ENERGY_PURCHASED_RTID,
+            date=date - pd.DateOffset(days=1),
+            extension="csv",
+        )
+
+        data = (
```
kmax12, line 3698: i think we are going to want a `_handle_dam_total_energy_purchased_sold(doc)` method, because when we receive the zip of backfill files from ERCOT, we will want be able to call just those steps on them

WillKoehrsen: Makes sense. I'll wait until we have the files to implement.

kmax12 (approved):
LGTM once all tests pass

gridstatus/ercot.py
```diff
@@ -3386,10 +3392,13 @@ def read_doc(
...
-        df = pd.read_csv(
-            io.BytesIO(response), compression="zip", **(read_csv_kwargs or {})
-        )
+        if request_kwargs:
+            response = requests.get(doc.url, **(request_kwargs or {})).content
+            df = pd.read_csv(
+                io.BytesIO(response), compression="zip", **(read_csv_kwargs or {})
+            )
+        else:
+            df = pd.read_csv(doc.url, compression="zip", **(read_csv_kwargs or {}))
```
WillKoehrsen, line 3401: This change is necessary because `doc.url` might be a `StringIO`. `pd.read_csv` can read in from a file buffer, so, unless we need to use `requests` with `request_kwargs`, we can use `pd.read_csv` to handle both urls and `StringIO`.

gridstatus/ercot.py
```diff
@@ -3658,3 +3667,94 @@ def _handle_indicative_lmp_by_settlement_point(self, df: pd.DataFrame):
...
+        Returns:
+            pandas.DataFrame: A DataFrame with DAM total energy purchased data
+        """
+        if date == "latest":
+            return self.get_dam_total_energy_purchased(
+                date="today",
+                verbose=verbose,
+            )
+
+        # DAM data so subtract one from the date
```
WillKoehrsen, line 3694: For example, to get data for May 20, we want to retrieve the file from May 19. The function returns data on the provided `date`.

.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
     hooks:
       - id: validate-pyproject
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.6.8
+    rev: v0.11.5
```
WillKoehrsen, line 25: The `ruff` in `pre-commit` and the `ruff` in the `pyproject.toml` were different versions and were formatting files differently.

WillKoehrsen:
The ERCOT tests were already failing. I've been running the newly added tests with 

```
VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot.py -k test_get_dam_total_energy
```

PR #605, opened 2025-05-22
fix(ENG 1619): Clean up IESO Tests and Methods

Kladar:
## Summary
Cleans up the various out of date IESO methods and their tests now that we're in a stable post-market renewal state. Let me know if I removed or marked `NotSupported` something that shouldn't be. 

### Details
~The `lmp_day_ahead_hourly` latest tests are failing because of the trading day failure in IESO I believe, so they should start passing tomorrow.~ looks like they are passing now

There are some datasets that only maintain 30 days, so for those methods (MCP, HOEP) I called them unsupported since there will be no data in about 9 days. I set their tests to check for the error raise (and pared them down to one test, where applicable). 

For methods where the data is coming from a different source and method (e.g. `load_forecast`), I put where the new data can be found in the error message.

It seems like the wind and solar generation forecasts can be missing intervals. There might be a pattern to this since the `publish time` = `12:38:07-05:00` was the issue on both 2025-05-19 and 2025-05-20, though for different forecast hour intervals (5 and 15, respectively). I'll keep an eye on this, and I handle the error to `continue` and log the error just once per file rather than for each missing data point. Update the relevant `all` test to not check for data length. 

Lastly, I also fixturize the generator report tests and do some minor updates to other tests to clean them up a bit.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -3554,8 +3554,21 @@ class IESO(ISOBase):
                             intervals = [intervals]
 
                         for interval in intervals:
-                            hour = int(interval["ForecastHour"])
-                            output = float(interval["MWOutput"])
+                            try:
+                                hour = int(interval["ForecastHour"])
+                                output = float(interval["MWOutput"])
+                            except KeyError:
+                                # NB: This logs the error once per file, rather than for each element in the file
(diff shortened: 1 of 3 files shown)


WillKoehrsen (approved):
Tests pass locally. Thanks for doing this cleanup work!

PR #606, opened 2025-05-22
[ENG-1658] ERCOT COP Adjustment Period Snapshot

WillKoehrsen:
## Summary

- Adds `Ercot().get_cop_adjustment_period_snapshot_60_day`

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -3807,7 +3807,7 @@ class Ercot(ISOBase):
             .reset_index(drop=True)
         )
 
-        # These columns are not in older data files
+        # Columns not in older data files. The columns first have data on 2024-06-28
         for col in ["Minimum SOC", "Maximum SOC", "Hour Beginning Planned SOC"]:
             if col not in data.columns:
                 data[col] = pd.NA
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -3807,6 +3807,11 @@ class Ercot(ISOBase):
             .reset_index(drop=True)
         )
 
+        # These columns are not in older data files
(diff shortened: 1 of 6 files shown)


gridstatus/ercot_api/pubapi-apim-api.json
WillKoehrsen, line 1: This file has changed and the old version didn't include the COP adjustment period snapshot dataset.

gridstatus/ercot.py
```diff
@@ -3758,3 +3761,60 @@ def get_dam_total_energy_sold(
...
+        data = (
+            data.rename(columns={"QSE Name": "QSE"})
+            .drop(
+                columns=["Time"],
+            )
+            .sort_values(["Interval Start", "Resource Name"])
+            .reset_index(drop=True)
+        )
+
+        # These columns are not in older data files
```
Kladar, line 3810: Do we know when they show up? Could probably memorialize that date here in the comment

Kladar (approved):
working for me. just curious, why implement both in Ercot and ErcotAPI?

WillKoehrsen:
> just curious, why implement both in Ercot and ErcotAPI?

The ErcotAPI has historical data going back to 2010-12-01 compared to a rolling ~425 past days from the Ercot method. However, we've found data is posted sooner to the Ercot method source than to the ErcotAPI source. So we use Ercot for incremental updates and the ErcotAPI for the historical backfill. 

We do this for many datasets include `ercot_as_plan`, `ercot_spp_real_time_15_min`, `ercot_lmp_by_settlement_point`, etc.

gridstatus/ercot.py
WillKoehrsen: Yep, added to the comment.

PR #607, opened 2025-05-23
feat(ENG 1662): PJM 5 minute tie flows

Kladar:
## Summary
Adds the `pjm_tie_flows_5_min` dataset.

### Details
30 day retention dataset. See note about the flooring timestamps to the nearest minute.

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2685,10 +2685,9 @@ class PJM(ISOBase):
             interval_duration_min=5,
             verbose=verbose,
         )
-
+        print(df)
         df = df.rename(
             columns={
-                "datetime_beginning_utc": "Interval Start",
                 "tie_flow_name": "Tie Flow Name",
                 "actual_mw": "Actual",
                 "scheduled_mw": "Scheduled",
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2661,3 +2661,43 @@ class PJM(ISOBase):
                 "Modified Datetime UTC",
             ]
         ]
+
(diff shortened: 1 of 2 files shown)


gridstatus/pjm.py
```diff
@@ -2661,3 +2661,42 @@ def get_regulation_market_monthly(
...
+            columns={
+                "tie_flow_name": "Tie Flow Name",
+                "actual_mw": "Actual",
+                "scheduled_mw": "Scheduled",
+            },
+        )
+        # NB: The data has an extra second on each timestamp like 2025-05-20 12:00:01, so we need to floor to the minute
+        df["Interval Start"] = pd.to_datetime(df["Interval Start"]).dt.floor("min")
+        df["Interval End"] = pd.to_datetime(df["Interval End"]).dt.floor("min")
+
```
Kladar, line 2699: Every interval seems to have an extra second in the source data. We could keep this as is, but I'm flooring it to the nearest minute here. 
CleanShot 2025-05-23 at 11 34 02@2x

WillKoehrsen: I'm fine with this as long as we're consistent across datasets. It looks like we round to the nearest minute in `get_load` so there's precedent but I'm not sure if there are others where there are seconds and we don't round.

https://github.com/gridstatus/gridstatus/blob/919b760e9e49de1ed8be327271bf653d60b6bfe0/gridstatus/pjm.py#L169-L174

Kladar: yeah it's weird cause it's not an instantaneous dataset and the seconds is always just one extra second. seems like an artifact on the PJM side of things

PR #613, opened 2025-05-30
[ENG-1676] ERCOT System Level ESR

WillKoehrsen:
## Summary

- Adds `ErcotAPI().get_system_load_charging_4_seconds()`


gridstatus/ercot_api/ercot_api.py
```diff
@@ -1293,6 +1327,50 @@ def get_cop_adjustment_period_snapshot_60_day(
...
+                verbose=verbose,
+            )
+
+        api_params = {"AGCExecTimeFrom": date, "AGCExecTimeTo": end}
+
+        data = self.hit_ercot_api(
+            endpoint=ESR_ENDPOINT,
+            api=APITypeEnum.ESR_API,
+            **api_params,
+        )
```
WillKoehrsen, line 1349: I haven't got the historical bulk download method working yet for this dataset, but, since this dataset only has data starting yesterday, I think that's fine for now.

gridstatus/ercot.py
```diff
@@ -3430,6 +3430,21 @@ def read_docs(
             )
         return pd.concat(dfs).reset_index(drop=True)
 
+    def ambiguous_based_on_dstflag(self, df: pd.DataFrame) -> pd.Series:
```
WillKoehrsen, line 3433: Moved this to a top-level function in this class so it can be used by `ErcotAPI`.

.env.template
```diff
@@ -4,13 +4,17 @@ EIA_API_KEY=
 # Register at https://apiportal.pjm.com/ for an API key
 PJM_API_KEY=
 
+# NOTE: The ERCOT API is geo-restricted.
```
WillKoehrsen, line 7: If you want to run this, you'll need to use a VPN and set your location to Texas.

gridstatus/ercot_api/ercot_api.py
```diff
@@ -1293,6 +1327,50 @@ def get_cop_adjustment_period_snapshot_60_day(
...
+
+        data = data.rename(
+            columns={
+                "AGCExecTime": "Timestamp",
```
kmax12, line 1353: let's call this just `Time` to be consistent with other datasets

kmax12, line 1353: does this data also return `AGCExecTimeUTC` column? if so, can we use that rather than rely on `AGCExecTime`/`DSTFlag`?

kmax12 (commented):
I haven't run the code yet since im not set up with a texas VPN

PR #615, opened 2025-06-03
feat(ENG 1696): AESO API Init

Kladar:
## Summary
Adds the `AESO()` class, generic API call methods, and supply and demand report based datasets:

- `aeso_supply_and_demand`
- `aeso_fuel_mix`
- `aeso_interchange`
- `aeso_reserves`

Also adds the `get_assets_list` method that will be useful for the next round of datasets, but that is WIP since it might get a bit more complicated as I continue to build it out. 

### API
First, you can create your own API key at the AESO signup site here: https://developer-apim.aeso.ca/signup. From there, subscribe the the AESO Public API product and it will generate a primary and secondary API key (this is exactly analogous to MISO and in fact the website seems the exact same format as the [MISO API site)](https://data-exchange.misoenergy.org/apis). Put this as you `AESO_API_KEY` env variable and you should be good to query data.

I do some moderate error handling on the API, but it has been fully reliable. The rate limit on the API appears to be very generous or nonexistent. Here's some stats - on a single API key, the API can handle with no issues:

- At least 50 concurrent requests to the same endpoint
- 100+ rapid sequential requests
- 20 sequential requests to each of 5 different endpoints (100 total requests), which is good as certain endpoints plug into others and require multiple requests to get all the data (more below)
- A sustained load test of 200 requests over 20 seconds, showing no noticeable degradation or rate limiting.

I can check in the script I used to do API rate limiting if we want it (it's in `/scripts` right now).

### Datasets
The most interesting aspect of the API is that it seems to only give `current` or `latest` data. There doesn't seem to be any way to get historical data from at least the supply and demand endpoint (prices, generation, etc might be a different story). I did try a dozen or so variations on the `/current` endpoint to try to find a hidden date range endpoint and couldn't but it wasn't exhaustive. It is also timestamp samplings rather than discrete intervals, so the dataset updates have to be experimented with and their publish frequency calculated. 

It feels weird to not add a `date` and `end` or `@support_date_range` for these methods, but alas that's where we're at.

For example, I polled the `get_supply_and_demand()` method every 10 seconds for 15 minutes and saw the dataset update every minute on the minute. I would call this a 1 minute dataset, since the last_updated_time spreads from the API are every minute on the minute, with a ~45 second delay between current time and publish time:

```
Started at: 2025-06-03 12:58:03
...
Update detected at 12:58:44
Time since last update: N/A
Time: 2025-06-03 11:57:00-06:00 -> 2025-06-03 11:58:00-06:00
Total Net Generation: 10250 -> 10247
Net Actual Generation: 7967 -> 7966
Net Actual Interchange: 516 -> 530
Alberta Internal Load: 9734 -> 9717
Cogeneration Net Generation: 3640 -> 3637
Combined Cycle Net Generation: 2693 -> 2690
Gas Fired Steam Net Generation: 957 -> 961
Simple Cycle Net Generation: 165 -> 164
Solar Net Generation: 1164 -> 1159
Wind Net Generation: 1095 -> 1100
British Columbia Flow: 454 -> 460
Montana Flow: 62 -> 70
...
Update detected at 12:59:45
Time since last update: 0:01:01.444126
Time: 2025-06-03 11:58:00-06:00 -> 2025-06-03 11:59:00-06:00
Total Net Generation: 10247 -> 10279
Net Actual Generation: 7966 -> 7982
Net Actual Interchange: 530 -> 553
Alberta Internal Load: 9717 -> 9726
Combined Cycle Net Generation: 2690 -> 2691
Gas Fired Steam Net Generation: 961 -> 959
Hydro Net Generation: 332 -> 333
Other Net Generation: 204 -> 200
Solar Net Generation: 1159 -> 1167
Wind Net Generation: 1100 -> 1128
British Columbia Flow: 460 -> 472
Montana Flow: 70 -> 81
...
Update detected at 13:00:47
Time since last update: 0:01:01.458310
Time: 2025-06-03 11:59:00-06:00 -> 2025-06-03 12:00:00-06:00
Total Net Generation: 10279 -> 10298
Net Actual Generation: 7982 -> 7973
Net Actual Interchange: 553 -> 561
Alberta Internal Load: 9726 -> 9737
Gas Fired Steam Net Generation: 959 -> 961
Hydro Net Generation: 333 -> 331
Other Net Generation: 200 -> 203
Solar Net Generation: 1167 -> 1173
Wind Net Generation: 1128 -> 1138
British Columbia Flow: 472 -> 482
Montana Flow: 81 -> 79
```
It returns a single, latest value. So we could construct an `Interval Start` and `Interval End` rather than a time column for all the supply and demand datasets based on this publishing. Would be curious for @cwaldoch 's thoughts on `Time` vs implied `Interval Start` and `Interval End` of 1 minute. 

I can check in the script I used to do endpoint/dataset update polling if we want it (it's in `/scripts` right now).

There is also some arbitrary selection going on in the columns for each dataset, so happy to iterate on those and add/subtract/change them for each dataset. To see the data and columns:

```
from gridstatus.aeso_api.aeso_api import AESO

aeso = AESO()
df = aeso.get_supply_and_demand()
print(df)
print(df.columns)

df2 = aeso.get_fuel_mix()
print(df2)
print(df2.columns)

df3 = aeso.get_interchange()
print(df3)
print(df3.columns)

df4 = aeso.get_reserves()
print(df4)
print(df4.columns)
```

### Other
Also adds some type hints to upstream utils, env template value, pre-commit mypy ignore, etc. 

### Future Work
I'll start in on adding the next batch of datasets now, with asset generation and price data, branching from this feature branch, and pull in changes that happen here from review.

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -233,7 +233,11 @@ class AESO:
             endpoint += "?" + "&".join(params)
 
         data = self._make_request(endpoint)
-        df = pd.DataFrame(data["return"])
+        df = pd.json_normalize(data["return"])
+
+        if df.empty:
+            return pd.DataFrame(columns=list(ASSET_LIST_COLUMN_MAPPING.values()))
+
         df = df.rename(columns=ASSET_LIST_COLUMN_MAPPING)
         df = df[list(ASSET_LIST_COLUMN_MAPPING.values())]
 


Kladar:
Looks like we might be able to programmatically access the historical data from the ets.aeso.ca

http://ets.aeso.ca/ets_web/ip/Market/Reports/ActualForecastWMRQHReportServlet?beginDate=06022025&endDate=06032025&contentType=csv

gridstatus/aeso_api/aeso_api.py
WillKoehrsen, line 1: Nit: should we call this file just `aeso.py` since the class is only `AESO`?

Kladar: yeah, and since it looks like there might be some historical data to be accessed from their ets site, meaning two sources and not all of them API. updated

WillKoehrsen: Do we want to add the historical data in a future PR? I think there's enough here for now.

cwaldoch: from my perspective I'd be pretty psyched to start getting current AESO data in and then run down historicals in the future, unclear to me how useful they are in this market (since it has, as the kids say, hella quirks).

Kladar: Yeah was planning on a future one for that, if we want to go that route. For now we'll get the current (enough for a live page)

gridstatus/aeso/aeso.py
```diff
@@ -0,0 +1,246 @@
...
+class AESO:
+    """
+    API client for Alberta Electric System Operator (AESO) data.
+
+    Handles authentication and provides methods to access various AESO datasets
+    including supply and demand, market data, and operational information.
+    """
+
+    # NB: Data is also typically provided in UTC
+    default_timezone = "America/Edmonton"
```
WillKoehrsen: Is this the same as `US/Mountain`? I see the data comes back from the API with a `_mpt` timezone which I assume is "Mountain Prevailing Time".

If so, could we use `US/Mountain` for the `default_timezone` since that's what AESO uses in their [documentation](https://www.aeso.ca/market/market-updates/2023/daylight-saving-time-notice)?

gridstatus/aeso/aeso.py
```diff
@@ -0,0 +1,246 @@
...
+            Parsed JSON response as dictionary
+
+        Raises:
+            HTTPError: If the API returns an HTTP error status
+            URLError: If there's a network/connection error
+            ValueError: If the response is not valid JSON
+        """
+        url = f"{self.base_url}/{endpoint.lstrip('/')}"
+
+        req = urllib.request.Request(url, headers=self.default_headers)
```
WillKoehrsen: Why `urllib` instead of `requests`? `urllib` seems more complicated.

gridstatus/tests/source_specific/test_aeso.py
WillKoehrsen, line 1: Do we need to update the Makefile to run this in GitHub actions?

gridstatus/aeso/aeso.py
Kladar: ah you're right, I was thinking this was market participant time haha. updated.

gridstatus/aeso/aeso.py
Kladar: ah yeah, a holdover from initial API exploration. updated and simplified.

gridstatus/tests/source_specific/test_aeso.py
Kladar: added the test-aeso command, I'll add the secret here

Kladar: added secret, rerunning action

Kladar: passing now 🦾

gridstatus/aeso/aeso.py
```diff
@@ -0,0 +1,238 @@
...
+            DataFrame containing current supply and demand information
+        """
+        endpoint = "currentsupplydemand-api/v1/csd/summary/current"
+        data = self._make_request(endpoint)
+
+        df = pd.json_normalize(data["return"])
+
+        df["Time"] = pd.to_datetime(
+            df["last_updated_datetime_utc"] + "+0000",
+            format="%Y-%m-%d %H:%M%z",
```
WillKoehrsen: Nit: I think you can just do `pd.to_datetime(df["last_updated_datetime_utc"], utc=True)` instead of concatenating the offset as a string.

WillKoehrsen (approved):
Data looks good and tests pass 👍🏽.

gridstatus/aeso/aeso.py
Kladar: agreed, and spotted a couple other parsing optimizations after looking at it fresh.

PR #618, opened 2025-06-04
feat(ENG 1690): AESO Pool Price, Forecast Pool Price, and System Marginal Price

Kladar:
## Summary
Adds pricing datasets, including:

- `aeso_pool_price`
- `aeso_pool_price_forecast`
- `aeso_system_marginal_price`

### Pool Price and Forecast Pool Price

This is an hourly dataset that pertains to hourly intervals, so this has an `Interval Start` and `Interval End` for the actuals and an `Interval Start`, `Interval End`, and `Publish Time` for forecast data. I think it does make sense to split up actual and forecast into separate datasets just to simplify downstream systems and updates that happen within the same report. 

It looks like the forecast is first published 3 hours ahead of the actual price (e.g. Hour Ending 16 will publish at Hour Ending 13). 
CleanShot 2025-06-04 at 13 16 12@2x

However, it looks like there is a silent update going on in the data between intervals where price forecasts are updating, with not a lot of metadata to show for it:

CleanShot 2025-06-04 at 13 49 29@2x

CleanShot 2025-06-04 at 13 55 04@2x

[According to the docs](http://ets.aeso.ca/Market/Reports/Manual/HelpText/current-actual-forecast-metadata.pdf), they are updating the forecast every 5 minutes. They are basically constantly, silently updating over the course of the next 3 hours. You can see this yourself by hitting "Ok" on the ETS over the course of an hour on the `Actual Forecast` report: http://ets.aeso.ca/ I've added some logic to figure out the publish time dynamically based on the current time. 

- If the interval is in the past, use the 5 minutes before the interval timestamp
- If the interval is in the future, take the request time (now) and floor to the 5 minute mark.

Downstream systems will hopefully be able to handle the rest. Open to many other approaches as well. 

CleanShot 2025-06-04 at 13 58 36@2x

Actual prices themselves are on about a two hour lag from current time.

### System Marginal Price
System Marginal Prices update seemingly whenever the price changes, and thus are actually an instantaneous dataset (rather than an interval dataset). The API returns start and end times, but it just updates the previous record to end and then sets the new record to be until the end of the day. In light of that, I just return the `Time` column.

**API Response Example**
CleanShot 2025-06-04 at 13 31 06@2x

**ETS System Reporting**
CleanShot 2025-06-04 at 13 35 39@2x

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -320,17 +320,15 @@ class AESO:
                 errors="coerce",
             )
             # NB: Publish times are a bit opaque from AESO, so we calculate a best estimate here.
-            # Earliest forecast is 3 hours before the interval.
-            # Then it can be updated continually within the next 2 hours after first publication.
-            # Forecasts are relatively unchanging 1 hour out from the interval.
+            # Forecast pool price is provided for current and next two hours, updated every 5 minutes.
+            # For future intervals: use request time floored to 5 minutes
+            # For past/current intervals: use 5 minutes before interval start
             request_time = pd.Timestamp.now(tz=self.default_timezone)
             df["Publish Time"] = df.apply(
                 lambda row: (
-                    request_time
-                    if row["Interval Start"] - request_time > pd.Timedelta(hours=2)
-                    else row["Interval Start"] - pd.Timedelta(hours=2)
-                    if row["Interval Start"] - request_time > pd.Timedelta(hours=1)
-                    else row["Interval Start"] - pd.Timedelta(hours=1)
+                    request_time.floor("5min")
+                    if row["Interval Start"] > request_time
+                    else row["Interval Start"] - pd.Timedelta(minutes=5)
                 ),
                 axis=1,
             )
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -319,7 +319,7 @@ class AESO:
                 df["Forecast Pool Price"],
                 errors="coerce",
             )
-            df["Publish Time"] = df["Interval Start"] - pd.Timedelta(hours=4)
+            df["Publish Time"] = df["Interval Start"] - pd.Timedelta(hours=3)
             return df[
                 [
                     "Interval Start",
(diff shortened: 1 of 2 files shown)


gridstatus/aeso/aeso.py
```diff
@@ -242,3 +243,140 @@ def get_asset_list(
...
+            return df[["Interval Start", "Interval End", "Pool Price"]]
+        else:
+            df["Forecast Pool Price"] = pd.to_numeric(
+                df["Forecast Pool Price"],
+                errors="coerce",
+            )
+            # NB: Publish times are a bit opaque from AESO, so we calculate a best estimate here.
+            # Forecast pool price is provided for current and next two hours, updated every 5 minutes.
+            # For future intervals: use request time floored to 5 minutes
+            # For past/current intervals: use 5 minutes before interval start
```
WillKoehrsen, line 325: Just wondering, where did you find out how often the forecast is updated? I didn't see it on the API page.

https://developer-apim.aeso.ca/api-details#api=poolprice-api-v1-1&operation=getPoolPriceDateRangeReport

gridstatus/aeso/aeso.py
```diff
@@ -242,3 +243,140 @@ def get_asset_list(
...
+                "system_marginal_price": "System Marginal Price",
+                "volume": "Volume",
+            },
+        )
+        df["System Marginal Price"] = pd.to_numeric(
+            df["System Marginal Price"],
+            errors="coerce",
+        )
+        df["Volume"] = pd.to_numeric(df["Volume"], errors="coerce")
+        return df[["Time", "System Marginal Price", "Volume"]]
```
WillKoehrsen: Could we add a `sort_values('Time')` to the returned data? Currently the data is being returned with the latest values first.

gridstatus/aeso/aeso.py
WillKoehrsen: Sorry, should have read the PR description.

http://ets.aeso.ca/Market/Reports/Manual/HelpText/current-actual-forecast-metadata.pdf

PR #619, opened 2025-06-06
[ENG-1722] MISO Interchange 5 Min

WillKoehrsen:
## Summary

- Adds MISO().get_interchange_5_min()
- Parses data in chart: https://publiccharts.misoenergy.org/charts/interchange

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -1630,3 +1630,92 @@ class MISO(ISOBase):
         )
 
         return final_df
+
+    @support_date_range(frequency=None)
+    def get_interchange_5_min(
+        self,
+        date: str | pd.Timestamp,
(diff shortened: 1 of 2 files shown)


Kladar (approved):
You call this out in the code, but just acking the funky `NA` business going on here. 
CleanShot 2025-06-09 at 08 54 31

WillKoehrsen:
> You call this out in the code, but just acking the funky `NA` business going on here.

Yes, the actual values are 5 minutes ahead of the scheduled values and the scheduled values go back further in time than the actuals. You can see the former on [the chart](https://publiccharts.misoenergy.org/charts/interchange):

Screenshot 2025-06-09 at 12 39 08

I thought about dropping rows where there are NAs, but in general, I think we want to err on the side of keeping data.

PR #622, opened 2025-06-09
feat(ENG-1733): AESO load and load forecast

Kladar:
## Summary
Adds the `aeso_load` and `aeso_load_forecast` datasets. 

### Details
Load is pretty straightforward. `Supply and Demand` has this data, but there is an endpoint `Actual Forecast` that has historical, so I use that instead.

Load Forecast is also included here, and the main thing worth noting is the `Publish Time` calculation (as usual). According to the info, it is published every day at 7am and forecasted for the next 13 days. So we assume that historical forecasts only have the latest available (published the same day as the intervals being forecast), and future forecasts can grab the full 13 days into the future (and this checks out, current latest available is 2025-06-22) with publish time being calculated from the current day. Historical days will have a single forecast, but future days will have 13 individual forecasts (one each day) leading up to the interval. 

See docs: http://ets.aeso.ca/Market/Reports/Manual/HelpText/current-actual-forecast-metadata.pdf

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -142,9 +142,17 @@ class AESO:
             DataFrame containing load forecast data with publish times
         """
         if date == "latest":
-            today = pd.Timestamp.now(tz=self.default_timezone).floor("D")
-            end = today + pd.Timedelta(days=13)
-            return self.get_load_forecast(date=today, end=end)
+            current_time = pd.Timestamp.now(tz=self.default_timezone)
+            today_7am = current_time.floor("D") + pd.Timedelta(hours=7)
+            publish_time = (
+                today_7am
+                if current_time >= today_7am
+                else today_7am - pd.Timedelta(days=1)
+            )
+            end = publish_time + pd.Timedelta(days=13)
+            df = self.get_load_forecast(date=publish_time, end=end)
+            df = df[df["Publish Time"] == publish_time]
+            return df
 
         start_date = pd.Timestamp(date).strftime("%Y-%m-%d")
         end_date = pd.Timestamp(end).strftime("%Y-%m-%d") if end else None


gridstatus/aeso/aeso.py
```diff
@@ -80,6 +80,120 @@ def _make_request(self, endpoint: str, method: str = "GET") -> dict[str, Any]:
...
+            utc=True,
+        ).dt.tz_convert(self.default_timezone)
+        df["Interval End"] = df["Interval Start"] + pd.Timedelta(hours=1)
+        df = df.rename(
+            columns={
+                "alberta_internal_load": "Load",
+            },
+        )
+        df["Load"] = pd.to_numeric(df["Load"], errors="coerce")
+        return df[["Interval Start", "Interval End", "Load"]]
```
WillKoehrsen, line 123: Should we drop the NA rows before returning?

Screenshot 2025-06-09 at 19 35 18

gridstatus/aeso/aeso.py
```diff
@@ -80,6 +80,120 @@ def _make_request(self, endpoint: str, method: str = "GET") -> dict[str, Any]:
...
+        the next 13 days. The publish time is determined as follows:
+        - For historical data: 7am on the day of the interval if interval is after 7am,
+          otherwise 7am the previous day
+        - For future data: 7am today (if after 7am) or 7am yesterday (if before 7am)
+
+        Returns:
+            DataFrame containing load forecast data with publish times
+        """
+        if date == "latest":
+            return self.get_load_forecast(date="today")
```
WillKoehrsen: I'm not sure what to do for "latest". This approach gets the load forecast only for today, but the latest load forecast goes out 13 days. Ideally, "latest" would return all the forecast values from today, going out to the end of the forecast period but since the API doesn't support querying by publish date this approach is probably fine.

gridstatus/aeso/aeso.py
Kladar: ah yeah, updated to remove them

gridstatus/aeso/aeso.py
Kladar: mm I agree, I think it should probably return the next 13 days if we want the latest. Added some manual look ahead logic to the method.

PR #624, opened 2025-06-11
feat(ENG 1683): AESO Unit Status

Kladar:
## Summary
Adds the `aeso_unit_status` dateset. This is another `Time`-based instantaneous sampling dataset like `interchange`, `reserves` etc. 

```
from gridstatus.aeso.aeso import AESO
aeso = AESO()
df = aeso.get_unit_status()
print(df)
print(df.columns)
```

(These AESO datasets all have their quirks so going to push them through via one PR at a time, sorry if it's a lot today. I have 3 more I think). 

### Details
Pretty straightforward, no historical data available, values from the API are updated ~minutely.

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -12,6 +12,7 @@ from gridstatus.aeso.aeso_constants import (
     RESERVES_COLUMN_MAPPING,
     SUPPLY_DEMAND_COLUMN_MAPPING,
 )
+from gridstatus.base import NotSupported
 from gridstatus.decorators import support_date_range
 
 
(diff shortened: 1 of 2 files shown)


gridstatus/aeso/aeso.py
```diff
@@ -531,3 +531,62 @@ def get_system_marginal_price(
         )
         df["Volume"] = pd.to_numeric(df["Volume"], errors="coerce")
         return df[["Time", "System Marginal Price", "Volume"]].sort_values(by="Time")
+
+    def get_unit_status(self) -> pd.DataFrame:
```
WillKoehrsen: When a dataset only has a current version, we still tend to have `date` and `end` parameters and raise an error if `date` is not "latest". 

An example is `MISO().get_fuel_mix("latest")`. https://github.com/gridstatus/gridstatus/blob/0ad1d1d6c5fe410c131f07b3b8ca37870f62a09b/gridstatus/miso.py#L87-L106

Kladar: updated

gridstatus/tests/source_specific/test_aeso.py
```diff
@@ -462,3 +462,37 @@ def test_get_load_forecast_historical_publish_times(
...
+        ]
+        for col in numeric_columns:
+            assert pd.api.types.is_numeric_dtype(df[col]), (
+                f"Column {col} should be numeric"
+            )
+
+    def test_get_unit_status(self):
+        """Test getting current unit status data."""
+        with api_vcr.use_cassette("test_get_unit_status.yaml"):
+            df = self.iso.get_unit_status()
```
WillKoehrsen: Need to add "latest" as the argument here.

PR #625, opened 2025-06-12
feat(ENG 1688): AESO Generator Outages

Kladar:
## Summary
Adds the `aeso_generation_outages_hourly` dataset. There are a lot of ins and outs of this dataset. More below

```
from gridstatus.aeso.aeso import AESO
aeso = AESO()
df = aeso.get_generator_outages_hourly(date="latest")
print(df)
print(df.columns)

df2 = aeso.get_generator_outages_hourly(date="2025-06-10", end="2026-06-10")
print(df2)
```

### Source
[The endpoint can be tried here](https://developer-apim.aeso.ca/api-details#api=aiesgencapacity-api-v1&operation=getAIESControlReport), and there is data available for the range 2011-01-01 to 2027-06-12 (24 months in the future). While this data is hourly, there is no ETS system equivalent, only summary reports such as the [Current Daily Outage Report](http://ets.aeso.ca/). ETS is updated every few minutes it seems. 

More info on some of the calculations, fuel types, updates, etc can be found in [that report's metadata page](http://ets.aeso.ca/Market/Reports/Manual/HelpText/current-daily-outage-metadata.pdf).

### "Latest"
Technically there are 24 months in the future which are "latest". However, that's a lot of hourly data to grab in a "latest" fashion and is quite slow. The daily outage report in ETS gives the next 4 months, which seems more reasonable, and we can make sure the user knows the next 24 months are available if they want them. So `latest` gives current time plus 4 months, but a `date="2025-06-12", end="2027-06-12")` will return data for the whole range.

The latest being mildly complex spills over into the handling of start and end dates as well, which is why this has some nested logic. There might be some ongoing DST issues, though this seems to resolve most of them. 

```
if date == "latest":
            current_time = pd.Timestamp.now()
            return self.get_generator_outages_hourly(
                date=current_time,
                end=current_time + pd.DateOffset(months=4),
            )
        else:
            start_date = pd.Timestamp(date)
            if start_date.tz is None:
                start_date = start_date.tz_localize(self.default_timezone)
            else:
                start_date = start_date.tz_convert(self.default_timezone)

            end_date = pd.Timestamp(end) if end else None
            if end_date is not None:
                if end_date.tz is None:
                    end_date = end_date.tz_localize(self.default_timezone)
                else:
                    end_date = end_date.tz_convert(self.default_timezone)
```

### Columns
There are a lot of options here, since we have `Fuel Type` (e.g. `Gas`, `Hydro`) and `Sub Fuel Type` (e.g. `Simple Cycle`), as well as outage types (`Operational Outage`, what we care about by fuel type, `Mothball Outage`, what we care about summed across fuel types). The subfuel types are all related to the `Gas` Fuel type, so I mostly ignore them, combining all the `Gas` subtypes into the simple `Gas` fuel type. The Daily Report also combines all the Mothball Outage (MBO) into a single column, so I do that here as well, and lastly create a `Total Outage` column that sums the operational outages. 

I could separate out the `Gas` subtypes, like `Gas - Simple Cycle`, `Gas - Cogeneration`, etc (I had this at one point but nixed it). Given that we are doing column headers as Fuel Types, this simpler naming seemed like the way to go. We might just have to explain this in the description of the dataset. @cwaldoch et al if you want to weight in here that would be good.

### "Publish Time"
I've gone back and forth on whether or not this is a forecast that needs a publish time and I'm currently operating on the thought that yes it is, and therefore needs a publish time. A user may want to see how future outages evolve over time. Technically these are updated quite frequently though, and there is no explicit publish time from the API (only in the ETS system). So I set the publish time to be 1 hour before each past interval, since it is an hourly dataset (though the outages are updated quite frequently, they are often unchanging) and the current time floored to the nearest hour for future intervals. This will result in a good middle ground of tracked changes but not constant (e.g. every 5 minutes) forecasts for the next 24 months.

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -618,7 +618,7 @@ class AESO:
             DataFrame containing generator outage data
         """
         if date == "latest":
-            current_time = pd.Timestamp.now()
+            current_time = pd.Timestamp.now(tz=self.default_timezone)
             return self.get_generator_outages_hourly(
                 date=current_time,
                 end=current_time + pd.DateOffset(months=4),
@@ -709,7 +709,7 @@ class AESO:
         # NB: Pivot the data to get the by-fuel type outage.
         df_pivot = df.pivot_table(
             index=["Interval Start", "Interval End", "Publish Time"],
-            columns="Fuel Type",
+            columns=["Sub Fuel Type"],
             values="Operating Outage",
             aggfunc="sum",
         ).reset_index()
@@ -719,7 +719,14 @@ class AESO:
             [
                 col
                 for col in df_pivot.columns
-                if col not in ["Interval Start", "Interval End", "Publish Time"]
+                if col
+                not in [
+                    "Interval Start",
+                    "Interval End",
+                    "Publish Time",
+                    "Mothball Outage",
+                    "Total Outage",
+                ]
             ]
         ].sum(axis=1)
 
@@ -750,7 +757,10 @@ class AESO:
             "Interval End",
             "Publish Time",
             "Total Outage",
-            "Gas",
+            "Simple Cycle",
+            "Combined Cycle",
+            "Cogeneration",
+            "Gas Fired Steam",
             "Coal",
             "Hydro",
             "Wind",
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -644,11 +644,12 @@ class AESO:
 
         fuel_type_mapping = {
             "GAS": "Gas",
+            "COAL": "Coal",
             "HYDRO": "Hydro",
             "WIND": "Wind",
             "SOLAR": "Solar",
             "ENERGY STORAGE": "Energy Storage",
-            "OTHER": "Other",
+            "OTHER": "Biomass and Other",
         }
 
         sub_fuel_type_mapping = {
@@ -656,8 +657,9 @@ class AESO:
             "COMBINED_CYCLE": "Combined Cycle",
             "COGENERATION": "Cogeneration",
             "GAS_FIRED_STEAM": "Gas Fired Steam",
+            "COAL": "Coal",
             "HYDRO": "Hydro",
-            "OTHER": "Other",
+            "OTHER": "Biomass and Other",
             "ENERGY STORAGE": "Energy Storage",
             "SOLAR": "Solar",
             "WIND": "Wind",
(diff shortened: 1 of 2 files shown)


gridstatus/aeso/aeso.py
```diff
@@ -598,3 +598,170 @@ def get_unit_status(
...
+            "SOLAR": "Solar",
+            "ENERGY STORAGE": "Energy Storage",
+            "OTHER": "Biomass and Other",
+        }
+
+        sub_fuel_type_mapping = {
+            "SIMPLE_CYCLE": "Simple Cycle",
+            "COMBINED_CYCLE": "Combined Cycle",
+            "COGENERATION": "Cogeneration",
+            "GAS_FIRED_STEAM": "Gas Fired Steam",
```
WillKoehrsen, line 659: For `AESO().get_fuel_mix()`, we have columns for all of these gas subtypes. I saw you talk about combining these in the description. Given the existing fuel mix and the fact that when we sum these to gas, we are losing information, I'd lean in the direction of making these all separate columns.

Kladar: yeah makes sense, I added them back in by just using the sub fuel type for every column (most fuel types and sub fuels types are the same, e.g. wind to wind. it's just gas that is different so it is an easy switch)

cwaldoch: great, I was gonna chime in here with the same as Will

gridstatus/aeso/aeso.py
```diff
@@ -598,3 +598,180 @@ def get_unit_status(
...
+
+        current_time = pd.Timestamp.now(tz=self.default_timezone).floor("h")
+        df["Publish Time"] = df.apply(
+            lambda row: (
+                current_time
+                if row["Interval Start"] > current_time
+                else row["Interval Start"] - pd.Timedelta(hours=1)
+            ),
+            axis=1,
+        )
```
WillKoehrsen, line 707: I agree with adding a publish time and this calculation seems reasonable given what we know.

gridstatus/aeso/aeso.py
```diff
@@ -598,3 +598,180 @@ def get_unit_status(
...
+        df_pivot["Total Outage"] = df_pivot[
+            [
+                col
+                for col in df_pivot.columns
+                if col
+                not in [
+                    "Interval Start",
+                    "Interval End",
+                    "Publish Time",
+                    "Mothball Outage",
```
WillKoehrsen, line 727: Mothball outages don't count towards the total outages?

WillKoehrsen: I don't really know enough about the details to suggest they should or shouldn't.

Kladar: Yeah technically Total Outage is Total Operational Outage; I can change the column name if helpful. Comes from the metadata sheet and the way the Daily Outage report is presented with the fuel types and then the MBO (Mothball Outage) being shown as a single column across all types http://ets.aeso.ca/ -> current -> Daily Outage Report

WillKoehrsen: Thanks. It wasn't clear to me from the dataset spec:

> Lets add all of the columns together for a total outage dataset without load

WillKoehrsen (approved):
👍🏽

PR #626, opened 2025-06-12
[ENG-1743] MISO Interchange Hourly

WillKoehrsen:
## Summary

- Adds MISOAPI().get_interchange_hourly()
- Using method requires getting a MISO API subscription key to the Load, Generation, and Interchange API https://data-exchange.misoenergy.org/api-details#api=load-generation-and-interchange-api

### DIFF
--- a/gridstatus/miso_api.py
+++ b/gridstatus/miso_api.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 import time
 from itertools import chain
@@ -20,6 +21,9 @@ CERTIFICATES_CHAIN_FILE = os.path.join(
 
 PRICING_PRODUCT = "pricing"
 BASE_PRICING_URL = "https://apim.misoenergy.org/pricing/v1"
+LOAD_GENERATION_AND_INTERCHANGE_PRODUCT = "load_generation_and_interchange"
+BASE_LOAD_GENERATION_AND_INTERCHANGE_URL = "https://apim.misoenergy.org/lgi/v1"
+
 PRELIMINARY_STRING = "Preliminary"
 FINAL_STRING = "Final"
 FIVE_MINUTE_RESOLUTION = "5min"
@@ -34,6 +38,7 @@ class MISOAPI:
     def __init__(
         self,
         pricing_api_key: str = None,
+        load_generation_and_interchange_api_key: str = None,
         initial_sleep_seconds: int = 1,
     ):
         """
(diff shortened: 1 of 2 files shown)


gridstatus/tests/source_specific/test_miso_api.py
```diff
@@ -135,3 +135,69 @@ def test_get_lmp_real_time_5_min_ex_post_final_date_range(self):
...
+        assert data["Interval End"].dtype == "datetime64[ns, EST]"
+
+        for col in data.columns:
+            if col not in ["Interval Start", "Interval End"]:
+                if col in ["Net Scheduled Interchange", "Net Actual Interchange"]:
+                    assert data[col].dtype == "float64"
+                else:
+                    assert data[col].dtype in ["int64", "Int64"]
+
+    @pytest.mark.integration
```
Kladar: since these are using a fixture they don't need to be marked integration

WillKoehrsen: Updated.

Kladar:
Sorry for the delay, will finish review this afternoon

gridstatus/miso_api.py
```diff
@@ -277,6 +281,131 @@ def _process_pricing_data(
...
 
+    @support_date_range(frequency="DAY_START")
+    def get_interchange_hourly(
+        self,
+        date: str | datetime.date | datetime.datetime,
+        end: datetime.date | datetime.datetime | None = None,
+        verbose: bool = False,
+    ):
+        if date == "latest":
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("H")
```
Kladar:
```suggestion
            date = pd.Timestamp.now(tz=self.default_timezone).floor("h")
```
to silence 
```
FutureWarning: 'H' is deprecated and will be removed in a future version, please use 'h' instead.
  date = pd.Timestamp.now(tz=self.default_timezone).floor("H")
```

gridstatus/tests/source_specific/test_miso_api.py
```diff
@@ -135,3 +135,73 @@ def test_get_lmp_real_time_5_min_ex_post_final_date_range(self):
...
+    def test_get_interchange_hourly_date_range(self):
+        start = (
+            self.local_start_of_today()
+            - pd.DateOffset(days=20)
+            + pd.DateOffset(hours=10)
+        )
+        end = start + pd.DateOffset(days=2, hours=4)
+
+        with api_vcr.use_cassette(
+            "test_get_interchange_hourly_{start:%Y%m%d}_{end:%Y%m%d}",
```
Kladar:
```suggestion
            f"test_get_interchange_hourly_{start:%Y%m%d}_{end:%Y%m%d}",
```
Need a f-string here, and a `.yaml` filetype on the fixture cassettes generally in this file. I think it technically works without it but shows up weird in `fixtures/`

CleanShot 2025-06-14 at 20 10 44@2x

Kladar (approved):
Nice, kind of 3 datasets in one + new API and everything. Couple small tweaks. We do get `NaN`s in various columns, assuming those are expected though

gridstatus/tests/source_specific/test_miso_api.py
WillKoehrsen: Good catch, updated.

gridstatus/miso_api.py
WillKoehrsen: Updated.

PR #627, opened 2025-06-16
feat(ENG 1744): Add CAISO Wind and Solar RTD, RTPD Forecasts and Actuals

Kladar:
## Summary
Adds the CAISO `get_rtd_renewable_forecast`, `get_rtpd_renewable_forecast` and `get_actual_renewables_hourly` methods to `CAISO()` and simplifies the handler methods and naming convention a bit across similar reports. 

Open questions:
- Renaming of existing `solar_and_wind_forecast_dam`
- ~Publish Time Offset correctness~

### Organization and Naming
Add the methods for the new datasets, as well as their source report constants and parameters, and places the existing `hasp_renewable_forecast_hourly` method up with the existing `solar_and_wind_forecast_dam` (which has been renamed to reflect the naming convention of this report to `dam_renewable_forecast_hourly` - more below). Also puts their tests together in the test file. Finally, expands but simplifies some of the helper methods, to simply `_process_renewable_hourly()` for the Actuals and DAM, and `_handle_renewable_forecast()` for the others (HASP, RTD, RTPD). Obviously these could probably be simplified further down to a single method since they are all the same response schema, but this is a step in the right direction. Open question on this PR if we want to do that final simplification. 

### Publish Times
Still ironing these out given what I can observe in OASIS, but should be an easily tunable parameter given the helper methods. From what I can tell:

**- RTD Published ~10 minutes ahead of time**
CleanShot 2025-06-15 at 21 47 12@2x

**- RTPD Published ~45 minutes ahead of time**
CleanShot 2025-06-15 at 21 50 02@2x

Actuals don't need a publish time.

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -1160,7 +1160,7 @@ class CAISO(ISOBase):
         )
         return self._handle_renewable_forecast(
             df,
-            publish_time_offset=pd.Timedelta(minutes=90),
+            publish_time_offset=pd.Timedelta(minutes=10),
         )
 
     def get_rtpd_renewable_forecast(
@@ -1193,7 +1193,7 @@ class CAISO(ISOBase):
         )
         return self._handle_renewable_forecast(
             df,
-            publish_time_offset=pd.Timedelta(minutes=90),
+            publish_time_offset=pd.Timedelta(minutes=45),
         )
 
     def get_pnodes(self, verbose: bool = False) -> pd.DataFrame:
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -867,7 +867,7 @@ class CAISO(ISOBase):
 
         return df
 
-    def get_renewable_hourly(
+    def get_renewables_hourly(
         self,
         date: str | pd.Timestamp,
         end: str | pd.Timestamp | None = None,
@@ -884,18 +884,18 @@ class CAISO(ISOBase):
             pandas.DataFrame: A DataFrame of wind and solar hourly actuals
         """
         if date == "latest":
-            return self.get_actual_renewable_hourly("today")
+            return self.get_renewables_hourly("today")
 
         df = self.get_oasis_dataset(
-            dataset="renewable",
+            dataset="renewables",
             date=date,
             end=end,
             verbose=verbose,
             raw_data=False,
         )
-        return self._process_renewable_hourly(df)
+        return self._process_renewables_hourly(df)
 
-    def get_renewable_forecast_dam(
+    def get_renewables_forecast_dam(
         self,
         date: str | pd.Timestamp,
         end: str | pd.Timestamp | None = None,
@@ -907,19 +907,19 @@ class CAISO(ISOBase):
         DAM Renewable Forecast
         """
         if date == "latest":
-            return self.get_dam_renewable_forecast_hourly("today", verbose=verbose)
+            return self.get_renewables_forecast_dam("today", verbose=verbose)
 
         current_time = pd.Timestamp.now(tz=self.default_timezone)
 
         data = self.get_oasis_dataset(
-            dataset="renewable_forecast_dam",
+            dataset="renewables_forecast_dam",
             date=date,
             end=end,
             verbose=verbose,
             raw_data=False,
         )
 
-        return self._process_renewable_hourly(
+        return self._process_renewables_hourly(
             data,
             current_time,
             # Day-ahead hourly wind and solar forecast is published at 7:00 AM according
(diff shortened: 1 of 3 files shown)


gridstatus/caiso/caiso.py
```diff
@@ -2208,74 +2394,6 @@ def _handle_lmp_hasp_15_min(self, df: pd.DataFrame) -> pd.DataFrame:
             ]
         ]
 
-    def get_hasp_renewable_forecast_hourly(
-        self,
-        date: str | pd.Timestamp,
-        end: str | pd.Timestamp | None = None,
-        verbose: bool = False,
-    ) -> pd.DataFrame:
-        """Get solar and wind generation HASP hourly data from CAISO.
```
Kladar, line 2217: This just moved as-is up to underneath the existing `solar_and_wind_forecast`, where the new methods also get added.

gridstatus/caiso/caiso_constants.py
```diff
@@ -109,6 +118,48 @@
...
+            "queryname": "SLD_REN_FCST",
+            "version": 1,
+        },
+        "params": {
+            "market_run_id": "HASP",
+        },
+        "meta": {
+            "max_query_frequency": "1d",
+        },
+    },
```
Kladar: same with this, just moved up to the other forecast dataset entry for legibility. new ones are below it.

gridstatus/caiso/caiso.py
```diff
@@ -867,43 +867,71 @@ def get_load_hourly(
...
+        df = self.get_oasis_dataset(
+            dataset="actual_renewable_hourly",
+            date=date,
+            end=end,
+            verbose=verbose,
+            raw_data=False,
+        )
+        return self._process_renewable_hourly(df)
+
+    def get_dam_renewable_forecast_hourly(
```
Kladar: renaming here feels consistent across datasets, but easy to change it back.

gridstatus/caiso/caiso.py
```diff
@@ -867,43 +867,71 @@ def get_load_hourly(
 
         return df
 
-    def get_solar_and_wind_forecast_dam(
+    def get_actual_renewable_hourly(
```
Kladar: if changing the `solar_and_wind_forecast_dam` back, then this becomes `get_solar_and_wind_hourly`

gridstatus/caiso/caiso.py
```diff
@@ -1010,6 +1055,147 @@ def _add_forecast_publish_time(
...
+        df = self.get_oasis_dataset(
+            dataset="rtpd_renewable_forecast",
+            date=date,
+            end=end,
+            verbose=verbose,
+            raw_data=False,
+        )
+        return self._handle_renewable_forecast(
+            df,
+            publish_time_offset=pd.Timedelta(minutes=45),
```
WillKoehrsen: When possible, I think we should use the official CAISO documentation for publication times. The schedules can be found at ATLAS REFERENCE > Publications > OASIS Publications Schedule.

Screenshot 2025-06-16 at 12 28 41

The OASIS publications schedule says this is published 22.5 minutes ahead of time. 

Screenshot 2025-06-16 at 12 31 30

gridstatus/caiso/caiso.py
```diff
@@ -1010,6 +1055,147 @@ def _add_forecast_publish_time(
...
+        df = self.get_oasis_dataset(
+            dataset="rtd_renewable_forecast",
+            date=date,
+            end=end,
+            verbose=verbose,
+            raw_data=False,
+        )
+        return self._handle_renewable_forecast(
+            df,
+            publish_time_offset=pd.Timedelta(minutes=10),
```
WillKoehrsen: The OASIS publications schedule says this is published 2.5 minutes ahead of time.

Screenshot 2025-06-16 at 12 31 48

gridstatus/caiso/caiso.py
Kladar: yep definitely, updated

gridstatus/caiso/caiso.py
Kladar: updated.

gridstatus/caiso/caiso_constants.py
```diff
@@ -109,6 +118,48 @@
...
+            "path": "SingleZip",
+            "resultformat": 6,
+            "queryname": "SLD_REN_FCST",
+            "version": 1,
+        },
+        "params": {
+            "market_run_id": "RTD",
+        },
+        "meta": {
+            "max_query_frequency": "1h",
```
WillKoehrsen: I'm not sure how we determine what this should be. I changed it to "1d", things worked fine, and it cut down the number of API requests by a factor of 24.

WillKoehrsen: My guess is there's a maximum number of rows the API will return at a time.

Kladar: ah yeah makes sense, I am totally cargo culting these parameters. I can bump it up. Updated to `1d`

WillKoehrsen (approved):

PR #628, opened 2025-06-17
feat(ENG 1686): AESO Tx Outages

Kladar:
## Summary
Adds the `aeso_transmission_outage` dataset. I can already tell this is going to be a fun one to maintain...

```
from gridstatus import AESO
aeso = AESO()
df = aeso.get_transmission_outages(date="latest")
df2 = aeso.get_transmission_outages(date="2025-01-17", end="2025-06-17")
print(df)
print(df.columns)
print(df2)
print(df2.columns)
```

### Details
Lots of fun quirks in this one. `latest` is pretty straighforward since we can navigate to the unchanging link.

Historical data is much more suspect, in that the only way I've found to get historical reports is to navigate to the latest report link and work backwards via the "Previous Version" link in the report. Since the archived reports have links (both html and csv) based on the semi-random publish time, we cannot navigate directly to them and thus much search them backward. This is made more complicated by the fact that some of the links are broken, either pointing to local machine links or to links that are not valid. For the local machine links, I've had some success substituting the normal URL scheme to find files. However, the trail runs cold at `2025-01-22 15:41:59` since the previous version link there doesn't work (likely a typo). So we have historical data back to then. 

The report isn't always consistent and has some problematic rows of one flavor or another. I try to handle these as best I can (but given the variable state of the source, my guess is these files are not heavily used).

I iterate through the files sequentially to find the links we want, and then iterate through them to grab the data. I could probably combine these tasks, but the break logic gets harder to manage then and overall the current approach is fairly efficient.

Also adds logging to AESO.

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -854,6 +854,7 @@ class AESO:
         """
         if date == "latest":
             url = "http://ets.aeso.ca/outage_reports/qryOpPlanTransmissionTable_1.html"
+            logger.info("Fetching latest transmission outages data")
             response = requests.get(url)
             response.raise_for_status()
 
@@ -861,6 +862,7 @@ class AESO:
             csv_link = soup.find("a", href=lambda x: x and "csvData" in x)
             csv_href = csv_link["href"].replace("\\", "/")
             csv_url = f"http://ets.aeso.ca/outage_reports/{csv_href}"
+            logger.info(f"Found CSV link: {csv_url}")
 
             publish_match = re.search(
                 r"(\d{4}-\d{2}-\d{2})_(\d{2}-\d{2}-\d{2})",
@@ -906,6 +908,9 @@ class AESO:
             # as there doesn't seem to be a better way to do it.
             start_date = pd.Timestamp(date)
             end_date = pd.Timestamp(end) if end else start_date
+            logger.info(
+                f"Fetching historical transmission outages from {start_date.date()} to {end_date.date()}",
+            )
 
             # NB: Start from the latest file and navigate backwards
             current_url = (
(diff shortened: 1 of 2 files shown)


gridstatus/aeso/aeso.py
```diff
@@ -831,3 +836,236 @@ def get_generator_outages_hourly(
...
+                                    # Convert local file path to proper web URL
+                                    filename = os.path.basename(csv_href)
+                                    csv_url = f"http://ets.aeso.ca/outage_reports/csvData/{filename}"
+                                    logger.info(
+                                        f"Converted local file path to web URL: {csv_url}",
+                                    )
+                                else:
+                                    csv_url = (
+                                        f"http://ets.aeso.ca/outage_reports/{csv_href}"
+                                    )
```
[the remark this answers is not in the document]
WillKoehrsen: You don't have to listen to this. I just wanted to see what Copilot would say.

gridstatus/aeso/aeso.py
```diff
@@ -831,3 +836,236 @@ def get_generator_outages_hourly(
...
+                ]
+            ]
+
+            return df
+
+        else:
+            # NB: For historical data, we need to navigate backwards through Previous Version links
+            # as there doesn't seem to be a better way to do it.
+            start_date = pd.Timestamp(date)
+            end_date = pd.Timestamp(end) if end else start_date
```
WillKoehrsen, line 917: I'm having trouble getting this to work for a single date.

Screenshot 2025-06-17 at 20 54 49

Kladar: Mm, yeah, it's a little unintuitive what a single date should return, but what makes the most sense is the transmission outages active on that day. I suppose to actually return the intervals (rather than the files that contain the intervals) I'll need to change this up a bit. 

The issue is that the "intervals" right now are just the start and end date of an outage, which can span many years. So I'll either need to return things that are active during that day (and do some logic to figure that out) or generate some continuous timeseries. 

I can already say I hate this dataset for many reasons and it seems rarely used and poorly engineered from source, so while I want to make it usable and intuitive, I want to narrow the scope as much as possible on implementing it.

gridstatus/aeso/aeso.py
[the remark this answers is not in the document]
Kladar: No these are actually useful suggestions.

WillKoehrsen:
@Kladar I wrote a script that enumerated all possible URLs working backwards from 2025-01-22 15:41:59 and found a working one.

* http://ets.aeso.ca/outage_reports/archives/_2025-01-17_14-10-25_qryOpPlanTransmissionTable_1.html

From this URL, we can use the Previous Version to find additional earlier URLs.

* http://ets.aeso.ca/outage_reports/archives/_2025-01-15_16-33-34_qryOpPlanTransmissionTable_1.html
* http://ets.aeso.ca/outage_reports/archives/_2025-01-08_14-49-21_qryOpPlanTransmissionTable_1.html
* http://ets.aeso.ca/outage_reports/archives/_2024-12-11_15-25-22_qryOpPlanTransmissionTable_1.html

I haven't checked exactly how far back these reports go. If we hit another dead end, we can run this script to pick up earlier reports. Or if historical reports aren't important, we could just stop.

[find_urls_aeso_transmission.py.zip](https://github.com/user-attachments/files/20799578/find_urls_aeso_transmission.py.zip)

gridstatus/aeso/aeso.py
```diff
@@ -831,3 +836,236 @@ def get_generator_outages_hourly(
...
+                        )
+                        if publish_match:
+                            publish_date = publish_match.group(1)
+                            publish_time = publish_match.group(2).replace("-", ":")
+                            publish_datetime = pd.to_datetime(
+                                f"{publish_date} {publish_time}",
+                            )
+                            publish_datetime = publish_datetime.tz_localize(
+                                self.default_timezone,
+                            )
```
WillKoehrsen, line 971: Just to make sure I understand: the publish time is from the CSV link which is _not_ the same as the datetime in the report URL? 

It looks like the datetime in the CSV link matches the prepared on datetime so this seems right.

Screenshot 2025-06-18 at 11 27 49

Kladar: yeah not entirely sure what they are going for in the report URL, but it seems like the prepared on/CSV link is what we want. wasn't entirely sure they were always different but yeah. Though we are searching the report URL for the date range args so conceivably this could miss. but so few reports and the two are usually very close so not too worried about it

gridstatus/aeso/aeso.py
Kladar: okay, this is working for single dates in that it looks for the most recent before the date given. so searching by publish time (not any sort of interval start from the actual data)

gridstatus/aeso/aeso.py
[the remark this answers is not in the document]
Kladar: implemented them in some form

gridstatus/aeso/aeso.py
WillKoehrsen: Thanks, works now.

PR #631, opened 2025-06-19
feat(ENG 1761): AESO Renewables Forecasts

Kladar:
## Summary
Adds the `aeso_wind_forecast_7_day`, `aeso_wind_forecast_12_hour`, `aeso_solar_forecast_7_day`, and `aeso_solar_forecast_12_hour`.

```
from gridstatus.aeso import AESO
aeso = AESO()
df = aeso.get_wind_forecast_12_hour(date="latest")
df2 = aeso.get_solar_forecast_12_hour(date="latest")
df3 = aeso.get_wind_forecast_7_day(date="latest")
df4 = aeso.get_solar_forecast_7_day(date="latest")
df5 = aeso.get_wind_forecast_7_day(date="2023-03-01")
df6 = aeso.get_solar_forecast_7_day(date="2023-03-01")
```

Currently the forecasts and actuals seem to be lagging a bit, so not entirely sure how this will perform. [Source here](https://www.aeso.ca/grid/grid-planning/forecasting/wind-and-solar-power-forecasting).

### Latest and Historical
The short term forecasts are only available as a `latest` implementation. 

The long term forecasts have historical data available, but only for the range of March 2023 through March 2025. The schema for the historical values are slightly different in that they need to calculate some values that the latest have already calculated, but this is straightforward.

### Publish Times
The shortterm forecasts are published every 10 minutes (currently they seem to be lagging a bit?). The long term forecasts are published every hour. Thus we can calculate the `Publish Time` for: 
- Past forecasted intervals, as the 10 minutes (short term) or 60 minutes (long term) before the `Interval Start` 
- Future forecasted intervals, as the first `Interval Start` timestamp where there is no `Actual` value.

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -1283,7 +1283,10 @@ class AESO:
             format="%Y-%m-%d %H:%M",
         ).dt.tz_localize(self.default_timezone)
 
-        df["Interval End"] = df["Interval Start"] + pd.Timedelta(minutes=10)
+        interval_length = (
+            pd.Timedelta(minutes=10) if term == "shortterm" else pd.Timedelta(hours=1)
+        )
+        df["Interval End"] = df["Interval Start"] + interval_length
 
         # NB: Since the forecasts are published every 10 minutes for shortterm and every 1 hour for longterm,
         # we can calculate the publish time based on the presence of actuals values.
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -1218,10 +1218,7 @@ class AESO:
             format="%Y-%m-%d %H:%M",
         ).dt.tz_localize(self.default_timezone)
 
-        if term == "shortterm":
-            df["Interval End"] = df["Interval Start"] + pd.Timedelta(minutes=10)
-        elif term == "longterm":
-            df["Interval End"] = df["Interval Start"] + pd.Timedelta(hours=1)
+        df["Interval End"] = df["Interval Start"] + pd.Timedelta(minutes=10)
 
         if date != "latest":
             start_date = (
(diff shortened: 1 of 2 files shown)


gridstatus/aeso/aeso.py
```diff
@@ -1104,3 +1117,336 @@ def _generate_csv_url(self, csv_href: str) -> str:
...
+            raise RequestException(
+                f"Failed to fetch {forecast_type} forecast data: {str(e)}",
+            )
+
+        df["Interval Start"] = pd.to_datetime(
+            df["Forecast Transaction Date"],
+            format="%Y-%m-%d %H:%M",
+        ).dt.tz_localize(self.default_timezone)
+
+        df["Interval End"] = df["Interval Start"] + pd.Timedelta(minutes=10)
```
WillKoehrsen: Should the timedelta here be 1 hour for `longterm`?

Kladar: ope yep, good catch; updated

gridstatus/aeso/aeso.py
```diff
@@ -1104,3 +1117,339 @@ def _generate_csv_url(self, csv_href: str) -> str:
...
+        )
+        df["Interval End"] = df["Interval Start"] + interval_length
+
+        # NB: Since the forecasts are published every 10 minutes for shortterm and every 1 hour for longterm,
+        # we can calculate the publish time based on the presence of actuals values.
+        # For past forecasted intervals (intervals with an actual value), we know the most recent forecast was published just before each interval.
+        # For future forecasted values, the publish time for all of them is set to the first interval that has no actual value, since
+        # that's when the forecast was last published.
+        publish_time_offset = (
+            pd.Timedelta(minutes=10) if term == "shortterm" else pd.Timedelta(hours=1)
```
WillKoehrsen: I think you can use `interval_length` here.

WillKoehrsen (approved):

gridstatus/aeso/aeso.py
Kladar: yep, done

PR #632, opened 2025-06-23
Hardcode SPP List of Hubs and Interfaces

WillKoehrsen:
## Summary

- Hardcode the SPP LMP list of hubs and interfaces so we don't have to rely on an external service to retrieve them


kmax12 (approved):

Kladar (approved):
following along, works against local db

PR #634, opened 2025-06-24
fix(ENG 1813): IESO Predispatch

Kladar:
## Summary
Enforce types and parallelize URL getting for IESO predispatch LMP methods where needed.

Examples to try:
```
df = ieso.get_lmp_predispatch_hourly("2025-06-01 04:00:00", end="2025-06-4 04:05:00")
df2 = ieso.get_lmp_predispatch_hourly_ontario_zonal("2025-06-01 04:00:00", end="2025-06-4 04:05:00")
df3 = ieso.get_lmp_predispatch_hourly_intertie("2025-06-01 04:00:00", end="2025-06-4 04:05:00")
```

### Details
Parallelization speeds these methods up by ~50%-80%. Since the dataframe cleanup code for the 3 predispatch submethods (`virtual_zonal`, `ontario_zonal`, `intertie`) is the same, I pull this parallelization and cleanup code out into a single helper function. 

I also fix some IESO tests that were failing. Since data eventually rolls off the back at the source, it makes these relative historical dates rather than fixed dates. All IESO tests should be back to passing now.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -2257,16 +2257,30 @@ class IESO(ISOBase):
                 f"No Predispatch Hourly Virtual Zonal LMP data found for {date} to {end}",
             )
 
-        data_list = []
-
-        for url in tqdm.tqdm(urls):
-            file_data = self._parse_lmp_hourly_virtual_zonal(
+        def process_url(url: str) -> pd.DataFrame:
+            return self._parse_lmp_hourly_virtual_zonal(
                 url,
                 verbose=verbose,
                 predispatch=True,
             )
 
-            data_list.append(file_data)
+        data_list = []
+        with ThreadPoolExecutor(max_workers=min(10, len(urls))) as executor:
+            future_to_url = {executor.submit(process_url, url): url for url in urls}
+
+            for future in tqdm.tqdm(as_completed(future_to_url), total=len(urls)):
+                url = future_to_url[future]
+                try:
+                    file_data = future.result()
+                    data_list.append(file_data)
+                except Exception as e:
+                    logger.error(f"Error processing {url}: {str(e)}")
+                    continue
+
+        if not data_list:
+            raise NoDataFoundException(
+                f"No valid data found for Predispatch Hourly Virtual Zonal LMP for {date} to {end}",
+            )
 
         data = pd.concat(data_list)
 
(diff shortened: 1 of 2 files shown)


gridstatus/ieso.py
```diff
@@ -1987,7 +1984,25 @@ def get_lmp_predispatch_hourly(
             )
 
             file_data["Publish Time"] = publish_time
-            data_list.append(file_data)
+            return file_data
+
+        data_list = []
+        with ThreadPoolExecutor(max_workers=min(10, len(urls))) as executor:
```
kmax12: why can't this use `_process_urls_with_threadpool`?

gridstatus/ieso.py
```diff
@@ -2906,6 +2896,71 @@ def _process_lmp_hourly_ontario_zonal(
...
         return df
 
+    def _process_urls_with_threadpool(
+        self,
+        urls: list[str],
+        process_func: callable,
+        error_message: str,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """Generic helper to process multiple URLs using ThreadPoolExecutor.
```
WillKoehrsen, line 2869: Does this function work with any IESO dataset or is it LMP specific?

gridstatus/ieso.py
```diff
@@ -2607,7 +2600,15 @@ def _parse_lmp_hourly_intertie(
...
 
                     hour = int(hour_data.find(hour_str, ns).text)
-                    value = float(hour_data.find("LMP", ns).text)
+                    lmp_elem = hour_data.find("LMP", ns)
+                    if (
+                        lmp_elem is None
+                        or lmp_elem.text is None
+                        or lmp_elem.text.strip() == ""
+                    ):
+                        continue
```
WillKoehrsen, line 2572: What's the reason behind this change?

Kladar: There are occasional LMP elements missing that fail the data pull on type conversion

`Error: argument must be str or a real number, not NoneType`

gridstatus/ieso.py
Kladar: currently it is predispatch LMP specific. Not trying to refactor the whole class, just make optimizations where I see them.

gridstatus/ieso.py
Kladar: oh actually I was thinking it was different than the other 3 predispatch methods, but it looks the same. 

the 
```
data["Location"] = data["Location"].str.replace(":LMP", "")
```
line was faking me out. updated.

gridstatus/ieso.py
Kladar: actually I shouldn't say that exactly - it would work with any method that has these columns I think 
```
 "Interval Start",
"Interval End",
"Publish Time",
"Location",
```

PR #636, opened 2025-06-24
fix(ENG 1789): Separate AESO generation actuals from forecasts

Kladar:
## Summary
Separates out the actual generation data from the wind and solar 12 hour and 7 day forecasts in AESO. Creates two new datasets called `aeso_wind` and `aeso_solar`

### Details
Since they call the same file link at first I tried to use more of the existing forecast methods, but this was getting a bit nested so I broke out the helper method for the actuals data. The forecast data methods remain unchanged (except for removing the actual generation columns).

### DIFF
--- a/gridstatus/aeso/aeso.py
+++ b/gridstatus/aeso/aeso.py
@@ -1443,7 +1443,6 @@ class AESO:
                 "Min": "Minimum Generation Forecast",
                 "Most Likely": "Most Likely Generation Forecast",
                 "Max": "Maximum Generation Forecast",
-                "Actual": "Actual Generation",
                 "MCR": f"Total {forecast_type.capitalize()} Capacity",
                 "Pct Min": "Minimum Generation Percentage",
                 "Pct Most Likely": "Most Likely Generation Percentage",
@@ -1455,7 +1454,6 @@ class AESO:
             "Minimum Generation Forecast",
             "Most Likely Generation Forecast",
             "Maximum Generation Forecast",
-            "Actual Generation",
             f"Total {forecast_type.capitalize()} Capacity",
             "Minimum Generation Percentage",
             "Most Likely Generation Percentage",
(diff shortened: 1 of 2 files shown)


Kladar:
Ah damn, realizing I actually do think I need a short term and a long term even for the actuals, since one is hourly and one is 10 minutely. Moving to draft

gridstatus/aeso/aeso.py
```diff
@@ -1579,3 +1573,266 @@ def _get_wind_solar_forecast_historical_data(
...
+        Returns:
+            DataFrame containing actual wind generation data with hourly intervals
+        """
+        if date == "latest":
+            return self._get_wind_solar_actual_latest_data(
+                generation_type="wind",
+                forecast_type="longterm",
+            )
+        else:
+            start_date = self._normalize_timezone(pd.Timestamp(date))
```
WillKoehrsen: Why is this normalization necessary? I thought it was already done by the decorator which calls `utils._handle_date` which looks very similar to `AESO()._normalize_timezone`.

Kladar: ah yeah I knew we had a utility for that I just couldn't remember what it was. updated to `_handle_date`, looks like tests are passing with just dropping it in. updated.

Kladar: oh but you're saying the decorator should handle it without having to drop it in, and I was finding an error I think without specifically calling it. let me see what that was

Kladar: okay, I believe it's working without any explicit normalization (and I added some tests to cover the unsupported historical date ranges)

WillKoehrsen: Yeah, the decorator behavior is a little opaque.

WillKoehrsen (approved):
Code looks good and tests pass locally.

PR #637, opened 2025-07-03
[ENG-1852] PJM Instantaneous Dispatch Rates

WillKoehrsen:
## Summary

- Adds `PJM().get_instantaneous_dispatch_rates()` to get data from https://dataminer2.pjm.com/feed/inst_dispatch_rates/definition

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -888,7 +888,7 @@ class PJM(ISOBase):
         end: str | pd.Timestamp | None = None,
         start_row: int = 1,
         row_count: int = 50000,
-        interval_duration_min: int | None = None,
+        interval_duration_min: int | float | None = None,
         filter_timestamp_name: str = "datetime_beginning",
         verbose: bool = False,
     ):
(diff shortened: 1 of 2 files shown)


gridstatus/pjm.py
```diff
@@ -2700,3 +2700,48 @@ def get_tie_flows_5_min(
         return df[
             ["Interval Start", "Interval End", "Tie Flow Name", "Actual", "Scheduled"]
         ]
+
```
Kladar, line 2703: is support_date_range not needed because the endpoint does the pagination/request splitting for us in this case?

Mine worked with `start=2025-06-20, end=2025-07-03` so doesn't seem required, but not sure what the downstream effects are of not having it.

Kladar (approved):
Looks good, tests pass locally, pretty straightforward. One question about support_date_range decorator.

gridstatus/pjm.py
WillKoehrsen: I think it's technically not needed because `_get_pjm_json` handles both date normalization (using `utils._handle_date`) and pagination. However, since every other dataset retrieval method uses that decorator, I think it's better to add it for consistency.

PR #638, opened 2025-07-03
IESO Safe Parsing to Fix Missing Values

WillKoehrsen:
## Summary

- Adds safe parsing methods in IESO to handle nulls in elements

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -7,10 +7,11 @@ import re
 import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from enum import Enum
-from typing import Literal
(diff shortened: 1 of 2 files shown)


gridstatus/ieso.py
```diff
@@ -70,6 +70,72 @@ class SurplusState(str, Enum):
     NUCLEAR_SHUTDOWN = "Nuclear Shutdown"
 
 
+def _safe_find_text(element, tag, namespaces=None, default=None):
```
Kladar: type annotations needed for the helper functions

Kladar (approved):
makes sense to me, just need type hints

PR #641, opened 2025-07-16
feat(ENG 1854): PJM List Datasets 

Kladar:
## Summary
Adds `get_pricing_nodes`, `get_reserve_subzone_resources`, `get_reserve_subzone_buses`, `get_actual_operational_statistics`, `get_weight_average_aggregation_definition`

```
pjm = PJM()

df1 = pjm.get_pricing_nodes()
df2 = pjm.get_reserve_subzone_resources()
df3 = pjm.get_reserve_subzone_buses()
df4 = pjm.get_actual_operational_statistics(date="2025-07-10")
df5 = pjm.get_weight_average_aggregation_definition()

print(df1)
print(df2)
print(df3)
print(df4)
print(df5)
```

Also alters the existing `projected_peak_tie_flow` to operate on `Publish Time`, changing the `Generated At` to `Publish Time` and updating the `filter_timestamp_name`.

### Details
I tried a few approaches with these datasets such as a traditional `date` and `end` with `@support_date_range`, but thought it was weird to have to give a date to something like `get_pricing_nodes()`. Most of them are formulated with an `Effective Date` (the date they become active) and a `Termination Date`, the date the are inactivated. So I add an `as_of` field the these list datasets with those, that effectively lets the user query the active rows given a certain date (it defaults to getting all rows). 

We could incorporate that `as_of` time as a sort of `Publish Time`, if we want to capture the lists as they change over time. For example the [list of nodes](https://dataminer2.pjm.com/feed/pnode/definition) is updated daily. However, we could just grab the list as-is each time and let other systems handle updating data as they wish. My guess is that these list data are updated partially and in-place, such as when a new termination date is added to a node. I lean toward not including a "Publish Time" for these data, since it fans out considerably and there are likely better ways to capture updates and track changes over time. 

More in the comments

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2988,10 +2988,10 @@ class PJM(ISOBase):
             },
             interval_duration_min=60,
             verbose=verbose,
-            filter_timestamp_name="projected_peak_datetime",
+            filter_timestamp_name="generated_at_ept",
         )
 
-        df["Generated At"] = pd.to_datetime(df["generated_at_ept"]).dt.tz_localize(
+        df["Publish Time"] = pd.to_datetime(df["generated_at_ept"]).dt.tz_localize(
             self.default_timezone,
         )
         df["Interval Start"] = (
@@ -3000,8 +3000,9 @@ class PJM(ISOBase):
                 "UTC",
             )
             .dt.tz_convert(self.default_timezone)
+            .date()
         )
-        df["Interval End"] = df["Interval Start"] + pd.Timedelta(hours=1)
+        df["Interval End"] = df["Interval Start"] + pd.Timedelta(days=1)
         df = df.rename(
             columns={
                 "interface": "Interface",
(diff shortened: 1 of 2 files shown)


gridstatus/pjm.py
```diff
@@ -3012,8 +3013,306 @@ def get_projected_peak_tie_flow(
...
+        ).dt.tz_localize(
+            self.default_timezone,
+        )
+
+        if as_of is None:
+            return df
+
+        return df[(df["Termination Date"].isna()) | (df["Termination Date"] > as_of)]
+
+    def get_pricing_nodes(
```
Kladar, line 3109: There is an existing `get_pnode_ids()` method used in adding the pnode data to the LMP data. It's similar to this one. It probably makes sense to consolidate but wasn't quite sure.

That existing method also drops duplicates of Effective Date. But I replacing it, modifying inplace, or having a separate method was a question I was going to ask here.

gridstatus/pjm.py
```diff
@@ -3012,8 +3013,306 @@ def get_projected_peak_tie_flow(
...
+
+        df = self._get_pjm_json(
+            "agg_definitions",
+            start=None,
+            end=None,
+            params={
+                "fields": "effective_date_ept,terminate_date_ept,agg_pnode_id,agg_pnode_name,bus_pnode_id,bus_pnode_name,bus_pnode_factor",
+            },
+            verbose=verbose,
+        )
```
Kladar, line 3297: this grabs a lot of data and is a bit slow, but I couldn't seem to get it to work with a `filter_timestamp_name` on `terminate_date` to leave behind terminated data, so I mostly do that filtering (for all list methods) after the data pull. The other methods are quicker so filtering on API call matters less, but here it is a bit annoying.

gridstatus/pjm.py
```diff
@@ -2988,10 +2988,10 @@ def get_projected_peak_tie_flow(
             },
             interval_duration_min=60,
             verbose=verbose,
-            filter_timestamp_name="projected_peak_datetime",
+            filter_timestamp_name="generated_at_ept",
```
WillKoehrsen: We shouldn't need the `_ept` here since it's added in `_get_pjm_json`.

https://github.com/gridstatus/gridstatus/blob/6645f55516f88fa69dc029435ad9ad4370c5a2b9/gridstatus/pjm.py#L915-L917

gridstatus/pjm.py
```diff
@@ -3012,8 +3013,306 @@ def get_projected_peak_tie_flow(
...
+            start=date,
+            end=end,
+            params={
+                "fields": "datetime_beginning_utc,datetime_ending_utc,generated_at_ept,area,area_load_forecast,actual_load,dispatch_rate",
+            },
+            filter_timestamp_name="generated_at",
+            verbose=verbose,
+        )
+
+        df = df.rename(
```
WillKoehrsen, line 3054: At this point, the dataframe doesn't have columns `datetime_beginning_utc` and `datetime_ending_utc`. These have already been converted to Interval Start and Interval End in PJM's timezone by `_get_pjm_json`.

https://github.com/gridstatus/gridstatus/blob/6645f55516f88fa69dc029435ad9ad4370c5a2b9/gridstatus/pjm.py#L959-L973

gridstatus/pjm.py
```diff
@@ -3012,8 +3013,306 @@ def get_projected_peak_tie_flow(
...
+        )
+        df["Termination Date"] = pd.to_datetime(
+            df["Termination Date"],
+            errors="coerce",
+        ).dt.tz_localize(
+            self.default_timezone,
+        )
+
+        if as_of is None:
+            return df
```
WillKoehrsen, line 3105: Why would we call this method is `as_of` is `None`? Wouldn't that mean no filtering is done?

gridstatus/pjm.py
```diff
@@ -3012,8 +3013,306 @@ def get_projected_peak_tie_flow(
...
+    def get_pricing_nodes(
+        self,
+        as_of: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieves the pricing nodes data from:
+        https://dataminer2.pjm.com/feed/pnode/definition
+        """
+        if as_of == "now":
```
WillKoehrsen, line 3119: Can we pass `as_of` through `utils._handle_date` to make sure it's a timestamp before it gets sent to `_filter_active_records`? I also don't know if we need to support "now" or just "today" because that's what we do in other retrieval methods.

WillKoehrsen: Also, I was testing different things in Python and discovered this unexpected result:

```python
import pandas as pd

data = pd.DataFrame(
    {"timestamp": [pd.Timestamp.now(), pd.Timestamp.now() + pd.Timedelta(minutes=10)]}
)

data['timestamp'] > 'now'

0    False
1     True
Name: timestamp, dtype: bool

data['timestamp'] > 'today'

0    False
1     True
Name: timestamp, dtype: bool
```

You can directly compare the strings "now" and "today" against a datetime column which is why passing `as_of="today"` actually already works (but I still think we should pass `as_of` through `utils._handle_date`).

gridstatus/pjm.py
Kladar: ah yeah I always forget that. updated.

gridstatus/pjm.py
Kladar: ah yeah. updated.

gridstatus/pjm.py
Kladar: It is also doing the work of replacing the essentially "No Termination Date" termination dates. A bit of a bigger question of what we want to do with those termination dates. 

I could leave them as-is and operate on them like datetimes, but feel they really are a `None` from the ISO. Like if the ISO returned a `None` for these values I wouldn't fill in `9999-12-31T00:00:00` just so I could operate on them. But not married to that idea, we could potentially leave them. 

```
df["Termination Date"] = df["Termination Date"].replace(
            "9999-12-31T00:00:00",
            None,
        )

File: gridstatus/pjm.py (hunk 6 above)
Comment (Kladar) [reply gridstatus/pjm.py:3119]: TIL. but sounds good, updated.

File: gridstatus/pjm.py (hunk 1 above)
Comment (WillKoehrsen) [reply gridstatus/pjm.py:3109]: It would be nice to consolidate the methods. We don't have to do that in this PR though.

File: gridstatus/pjm.py (hunk 5 above)
Comment (WillKoehrsen) [reply gridstatus/pjm.py:3105]: `None` seems fine to me since that date effectively means these rows have no termination date.

File: gridstatus/pjm.py [hunk 7]
```diff
@@ -3012,8 +3013,308 @@ def get_projected_peak_tie_flow(
...
+    def get_pricing_nodes(
+        self,
+        as_of: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieves the pricing nodes data from:
+        https://dataminer2.pjm.com/feed/pnode/definition
+        """
+        as_of = utils._handle_date(as_of)
```
WillKoehrsen: We probably want to pass `tz=self.default_timezone` in all these `_handle_date` calls so `as_of` has a tz. Otherwise, this comparison `(df["Termination Date"] > as_of)` will error because `df['Termination Date']` has a timezone.

Kladar: definitely, updated

gridstatus/pjm.py
WillKoehrsen: Yeah, the terminate date in the API seems to be an exact match so you can get records terminated on a specific date but not records terminated after a date.

Screenshot 2025-07-17 at 11 16 29

gridstatus/tests/source_specific/test_pjm.py
```diff
@@ -2804,3 +2804,231 @@ def test_get_projected_peak_tie_flow_historical_date_range(self, date, end):
...
+        with pjm_vcr.use_cassette("test_get_actual_operational_statistics_latest.yaml"):
+            df = self.iso.get_actual_operational_statistics("latest")
+            self._check_actual_operational_statistics(df)
+            min_start = df["Interval Start"].min().date()
+            today = self.local_start_of_today().date()
+            yesterday = today - pd.Timedelta(days=1)
+            assert min_start in [today, yesterday]
+
+    @pytest.mark.parametrize("date, end", test_dates)
+    def test_get_actual_operational_statistics_historical_date_range(self, date, end):
```
WillKoehrsen, line 2838: I'm getting two failures on this test. Running 

```
VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_pjm.py -m "not slow" -k test_get_actual_operational_statistics_historical_date_range
```

The errors are:

```
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_actual_operational_statistics_historical_date_range[2023-11-05-2023-11-07] - AssertionError: assert datetime.date(2023, 11, 4) == datetime.date(2023, 11, 5)
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_actual_operational_statistics_historical_date_range[2024-09-02-2024-09-04] - AssertionError: assert datetime.date(2024, 9, 1) == datetime.date(2024, 9, 2)
```

Kladar: Yeah getting these test failures generally right now, investigating...

```
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_load_forecast_today - AssertionError: assert Timestamp('2025-04-17 00:00:00-0400', tz='US/Eastern') == Timestamp('2025-07-17 00:00:00-0400', tz='US/Eastern')
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_scheduled_interchange_real_time_latest - gridstatus.base.NoDataFoundException: No data found for rt_scheduled_interchange
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_projected_peak_tie_flow_latest - AttributeError: 'Series' object has no attribute 'date'
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_projected_peak_tie_flow_historical_date_range[2023-11-05-2023-11-07] - ValueError: No objects to concatenate
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_actual_operational_statistics_historical_date_range[2023-11-05-2023-11-07] - AssertionError: assert datetime.date(2023, 11, 4) == datetime.date(2023, 11, 5)
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_actual_operational_statistics_historical_date_range[2024-09-02-2024-09-04] - AssertionError: assert datetime.date(2024, 9, 1) == datetime.date(2024, 9, 2)
FAILED gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_projected_peak_tie_flow_historical_date_range[2024-09-02-2024-09-04] - ValueError: No objects to concatenate
```

Kladar: okay, I think this is working now while still doing what we want for those methods.

Issue #647, opened 2025-07-31
MISO Interconnection Queue request returns a 403 error


NoahV-SC:
Executing the MISO interconnection queue data request returns a 403 (Forbidden) error

miso = gridstatus.MISO()
miso.get_interconnection_queue()

The URL (https://www.misoenergy.org/api/giqueue/getprojects) works fine through browser and when requested through R, so it seems to be problem with how Python is received by the website. Adding a blank header to the request seems to fix the issue: 

    def get_raw_interconnection_queue(self, verbose: bool = False) -> BinaryIO:
        url = "https://www.misoenergy.org/api/giqueue/getprojects"

        msg = f"Downloading interconnection queue from {url}"
        logger.info(msg)

        response = requests.get(url, headers = "")
        return utils.get_response_blob(response)

Kladar:
Noting that the 403 error still occures on a VPN, so be sure to turn that off before proceeding with getting data.

alexsandberg-st:
@Kladar We're seeing this issue as well. Is there any timetable for when the fix will be released? Thank you!

Kladar:
@alexsandberg-st fix was added in https://github.com/gridstatus/gridstatus/pull/648 if you want to pull that latest commit. 

I will cut a release with the changes soon.

PR #648, opened 2025-07-31
fix: add blank header to MISO IxQ `request`

Kladar:
## Summary
Resolves https://github.com/gridstatus/gridstatus/issues/647

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -740,7 +740,7 @@ class MISO(ISOBase):
         msg = f"Downloading interconnection queue from {url}"
         logger.info(msg)
 
-        response = requests.get(url)
+        response = requests.get(url, headers="")
         return utils.get_response_blob(response)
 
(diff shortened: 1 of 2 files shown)


gridstatus/miso.py
```diff
@@ -740,7 +740,7 @@ def get_raw_interconnection_queue(self, verbose: bool = False) -> BinaryIO:
         msg = f"Downloading interconnection queue from {url}"
         logger.info(msg)
 
-        response = requests.get(url)
+        response = requests.get(url, headers="")
```
WillKoehrsen, line 743: This solution works for me, but only when I turn off my VPN. With a VPN on, I still get the 403 (no matter the VPN location).

NoahV-SC: Thanks for the quick fix! It works for me with a VPN on, but that may have to do with the VPN type.

Issue #653, opened 2025-08-12
NYISO Hourly LBMP


iamanonym0us:
Why there is no option for Real time Hourly LBMP for NYISO ? They publish hourly LBMP here https://mis.nyiso.com/public/csv/rtlbmp/20250812rtlbmp_zone.csv and https://mis.nyiso.com/public/csv/rtlbmp/20250812rtlbmp_gen.csv

WillKoehrsen:
Added in https://github.com/gridstatus/gridstatus/pull/656

PR #656, opened 2025-08-25
[ENG-2108] NYISO Hourly Realtime LMPs

WillKoehrsen:
## Summary

- Adds `NYISO().get_lmp(..., markets=gridstatus.REAL_TIME_HOURLY)`
- Addresses https://github.com/gridstatus/gridstatus/issues/653

### DIFF
--- a/gridstatus/base.py
+++ b/gridstatus/base.py
@@ -1,5 +1,5 @@
 import time
-from enum import Enum
+from enum import Enum, StrEnum
 from typing import BinaryIO
 
 import pandas as pd
@@ -26,7 +26,7 @@ class RetiredDataException(Exception):
     pass
 
 
-class Markets(Enum):
+class Markets(StrEnum):
     """Names of LMP Markets"""
 
     REAL_TIME_5_MIN = "REAL_TIME_5_MIN"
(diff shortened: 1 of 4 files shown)


gridstatus/base.py
```diff
@@ -26,7 +26,7 @@ class RetiredDataException(Exception):
     pass
 
 
-class Markets(Enum):
+class Markets(StrEnum):
```
WillKoehrsen, line 29: `StrEnum` makes more sense to me because it means the equivalent string is equal to the Enum value without needing to call `.value`.

```python
# True when using StrEnum, False when using Enum
"REAL_TIME_5_MIN" == Markets.REAL_TIME_5_MIN
```

gridstatus/nyiso.py
```diff
@@ -52,6 +57,7 @@ class DatasetInterval(NamedTuple):
     LOAD_FORECAST_DATASET: DatasetInterval("start", 60),
     DAM_LMP_DATASET: DatasetInterval("start", 60),
     REAL_TIME_LMP_DATASET: DatasetInterval("end", 5),
+    REAL_TIME_HOURLY_LMP_DATASET: DatasetInterval("start", 60),
```
WillKoehrsen, line 60: This must be "start" because the raw data file for a day starts at 00:00 and ends at 23:00.

PR #657, opened 2025-08-26
[ENG-2191] NYISO LMP Hourly Add Reference Bus

WillKoehrsen:

### DIFF
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -23,6 +23,7 @@ from gridstatus.lmp_config import lmp_config
 class NYISOLocationType(StrEnum):
     ZONE = "zone"
     GENERATOR = "generator"
+    REFERENCE_BUS = "reference_bus"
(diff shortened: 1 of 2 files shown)


Kladar (approved):
Working for me, tests pass locally 

Image

gridstatus/nyiso.py
```diff
@@ -625,6 +629,12 @@ def get_lmp(
         else:
             raise ValueError(f"Invalid location_type: {location_type}")
 
+        if REFERENCE_BUS_LOCATION in df["Location"].unique():
+            df.loc[
+                df["Location"] == REFERENCE_BUS_LOCATION,
+                "Location Type",
+            ] = "Reference Bus"
```
WillKoehrsen, line 637: Since the Reference Bus location is in the generator data, I vastly simplified this to just change the location type for the one Reference Bus.

PR #658, opened 2025-08-27
feat(ENG 2185): ERCOT post settlements hourly load

Kladar:
## Summary
Adds `get_load_hourly_post_settlements()` 

```
import gridstatus
ercot = gridstatus.Ercot()

df = ercot.get_hourly_load_post_settlements(date="2004-01-01", end="2025-08-10")
print(df)
```

### Details
Basically all of these files from each year are different in some way, so there is a lot of pandas datetime mashing to get this to work. I believe I handle all the edge cases across filetypes and date formats and different column headers, but hard to look at everything. 

Here's the link to ERCOT: https://www.ercot.com/gridinfo/load/load_hist

Also adds remaining missing typehints and logger to ercot.py while I'm here. No functionality changes beyond adding those.

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -906,7 +906,7 @@ class Ercot(ISOBase):
                 )
                 duplicate_indices = df[date_mask].index
                 second_idx = duplicate_indices[1]
-                logger.info(
+                logger.debug(
                     f"Changing timezone for DST duplicate at {df.loc[second_idx, 'Interval Start']}",
                 )
                 # Change from -06:00 to -05:00 by subtracting 1 hour
@@ -951,6 +951,7 @@ class Ercot(ISOBase):
             "West",
             "ERCOT",
         ]
+        df = df.dropna(subset=["Interval Start", "Interval End"])
         return df[expected_columns].sort_values("Interval Start").reset_index(drop=True)
 
     def _get_supply_demand_json(self) -> dict:


gridstatus/ercot.py
```diff
@@ -717,17 +749,200 @@ def _read_html_display(self, url, verbose=False):
 
         return df
 
-    def _get_supply_demand_json(self, verbose=False):
+    @support_date_range(frequency=None)
```
Kladar: Here is the new logic. Working for me across the whole time range

Image

gridstatus/ercot.py
```diff
@@ -717,17 +749,200 @@ def _read_html_display(self, url, verbose=False):
...
+
+        return df
+
+    def _process_ercot_post_settlements_load_data(
+        self,
+        df: pd.DataFrame,
+    ) -> pd.DataFrame:
+        df.columns = df.columns.str.strip()
+
+        column_mapping = {
```
Kladar, line 828: Not all of these columns are in all of the files, but this process ignores the ones that aren't so we can just maintain one list and not have a bunch of if/else to parse the columns that are there.

gridstatus/ercot.py
```diff
@@ -717,17 +749,200 @@ def _read_html_display(self, url, verbose=False):
 
         return df
 
-    def _get_supply_demand_json(self, verbose=False):
+    @support_date_range(frequency=None)
+    def get_ercot_hourly_load_post_settlements(
```
WillKoehrsen: I don't think we need the `ercot` in the method name.

Kladar: ha woops, updated for main method and hidden methods.

Kladar:
Also realizing I need tests. Sorry, this was such a pandas mash that I was happy to see the light at the end of the tunnel. Could use a `latest` method too.

gridstatus/ercot.py
```diff
@@ -717,17 +749,203 @@ def _read_html_display(self, url, verbose=False):
...
+        end = utils._handle_date(end, self.default_timezone)
+        start_year = date.year
+        end_year = end.year if end else start_year
+
+        logger.info(
+            f"Fetching historical load data for years {start_year} to {end_year}",
+        )
+
+        all_dfs = []
+        for year in range(start_year, end_year + 1):
```
kmax12: does using `@support_date_range(frequency="1 year")` or something like that not work?

gridstatus/ercot.py
```diff
@@ -717,17 +749,203 @@ def _read_html_display(self, url, verbose=False):
...
+            logger.info(f"Processing year {year}...")
+            df = self._download_post_settlements_load_file(year)
+            all_dfs.append(df)
+            logger.info(f"Successfully processed year {year}, got {len(df)} rows")
+        combined_df = pd.concat(all_dfs, ignore_index=True)
+
+        mask = (combined_df["Interval Start"] >= date) & (
```
kmax12: in general, i think it's okay if these methods return more data than requested if that corresponds to how the source data is. the caller can always filter it out if desired

gridstatus/ercot.py
kmax12: oh, i see we already have a `YEAR_START` defined. can we use that?

https://github.com/gridstatus/gridstatus/blob/16c9e1c0298a6d55d7b4764306a12872f8653215/gridstatus/decorators.py#L165-L166

gridstatus/ercot.py
kmax12: might as well leave this type of comment in the code

gridstatus/ercot.py
```diff
@@ -717,17 +749,203 @@ def _read_html_display(self, url, verbose=False):
...
+            df = pd.read_excel(zip_file.open(filename))
+        elif year_link.endswith(".xls") or year_link.endswith(".xlsx"):
+            response = requests.get(year_link)
+            response.raise_for_status()
+            df = pd.read_excel(io.BytesIO(response.content))
+        df = self._process_post_settlements_load_data(df)
+
+        return df
+
+    def _process_post_settlements_load_data(
```
kmax12, line 818: wow, a lot messier than I expected!

gridstatus/ercot.py
Kladar: makes more sense, implemented that.

gridstatus/ercot.py
Kladar: added

gridstatus/ercot.py
```diff
@@ -717,17 +749,203 @@ def _read_html_display(self, url, verbose=False):
...
+            date = pd.Timestamp(f"{current_year}-01-01")
+            end = pd.Timestamp(f"{current_year}-12-31")
+
+        date = utils._handle_date(date, self.default_timezone)
+        end = utils._handle_date(end, self.default_timezone)
+        start_year = date.year
+        end_year = end.year if end else start_year
+
+        logger.info(
+            f"Fetching historical load data for years {start_year} to {end_year}",
```
kmax12: this text might need to be updated since it is always going to be one year

Kladar: updated

gridstatus/ercot.py
```diff
@@ -717,17 +749,201 @@ def _read_html_display(self, url, verbose=False):
...
+        end = utils._handle_date(end, self.default_timezone)
+
+        logger.info(
+            f"Fetching historical load data for year {date.year}",
+        )
+
+        df = self._download_post_settlements_load_file(date.year)
+        logger.info(f"Successfully processed year {date.year}, got {len(df)} rows")
+
+        mask = (df["Interval Start"] >= date) & (df["Interval Start"] < end)
```
kmax12: recopying this comment here since it is on outdated diff now. 

for example, I think this break with `date="lastest"` since end will be dec 31, so the last day would get cut off.

https://github.com/gridstatus/gridstatus/pull/658#discussion_r2305336235

Kladar: I see, yeah I'm end exclusive here when I shouldn't be. 

Feels a bit weird to give a date range like `ercot.get_load_hourly_post_settlements(date='2025-05-01', end='2025-07-01')` and get back a years worth of data though, in that case `2025-01-01` to `2025-07-31`. But I can remove the filter

Kladar: removed

kmax12: I agree it's a bit weird, but the intuition is that we use the supplied dates to query the source endpoint, but won't do any filtering before what the source supports.

this approach is less useful if you are using `gridstatus` standalone, but is nice when used as part of an ETL pipeline where there may be situations where getting extra data is neutral to positive, but never bad.

gridstatus/ercot.py
```diff
@@ -717,17 +749,194 @@ def _read_html_display(self, url, verbose=False):
...
+        Returns:
+            pandas.DataFrame
+        """
+        if date == "latest":
+            # NB: Gets the most recent year available, since they are published as annual files
+            current_year = pd.Timestamp.now().year
+            date = pd.Timestamp(f"{current_year}-01-01")
+            end = pd.Timestamp(f"{current_year + 1}-01-01")
+
+        date = utils._handle_date(date, self.default_timezone)
```
kmax12, line 778: do you need to call `_handle_date` if you are using `support_date_range`?

kmax12 (approved):

PR #664, opened 2025-09-17
ENG-2285: Add PJM Regulation Prices (5 min)

Shiggiddie:
## Summary

Add PJM Regulation Prices (5 min)

### Details

* Add API fields for `reg_prices`
* Rename DF columns according to spec
* Add latest & historic test cases

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -2667,6 +2667,72 @@ class PJM(ISOBase):
         ]
 
     @support_date_range(frequency=None)
+    def get_regulation_prices_5_min(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieves 5-minute regulation pricing data from:
+        https://api.pjm.com/api/v1/reg_prices
+        """
+
+        if date == "latest":
+            return self.get_regulation_prices_5_min(
+                "today",
+                verbose=verbose,
+            )
+
+        params = {
+            "fields": (
+                "datetime_beginning_utc,datetime_beginning_ept,area,"
+                "reserve_quantity,reserve_requirement,market_clearing_price,"
+                "market_capped_clearing_price,capability_clearing_price,"
+                "performance_clearing_price,marginal_benefits_factor"
+            ),
+        }
+
+        df = self._get_pjm_json(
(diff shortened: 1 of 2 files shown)


gridstatus/pjm.py
```diff
@@ -2666,6 +2666,72 @@ def get_regulation_market_monthly(
...
+            start=date,
+            end=end,
+            params=params,
+            verbose=verbose,
+            interval_duration_min=5,
+        )
+
+        df = df.rename(
+            columns={
+                "datetime_beginning_ept": "Interval Start Local",
```
Kladar: You can just return the `datetime_beginning_utc` here and lose the `Interval Start Local` column. Downstream systems can then be left to handle localization/conversion/display in local times.

Edit: I see this was done in the dataset spec that way.

gridstatus/pjm.py
```diff
@@ -2666,6 +2666,72 @@ def get_regulation_market_monthly(
...
+                "area": "Area",
+                "reserve_quantity": "Regulation Quantity MW",
+                "reserve_requirement": "Regulation Requirement MW",
```
Kladar: Not your fault as they are in the dataset spec, and looks like Max commented on this as well already, but we usually leave off the units suffix.

Kladar (commented):
Looks good, couple comments on columns (due to the spec, not your fault, but it's more common that not to have spec be off, given the velocity of dataset additions we do).

gridstatus/tests/source_specific/test_pjm.py
```diff
@@ -2512,6 +2512,45 @@ def test_get_load_forecast_5_min_historical_range(self):
                 past_end_date,
             ) + pd.Timedelta(minutes=175)
 
+    """get_regulation_prices_5_min"""
+
+    regulation_prices_5_min_columns = [
+        "Interval Start",
+        "Interval End",
+        "Interval Start Local",
```
WillKoehrsen, line 2520: I think this needs to be removed.

WillKoehrsen (approved):

gridstatus/pjm.py
Shiggiddie: Updated!

gridstatus/pjm.py
Shiggiddie: Confirmed with markets team that this is not needed, dropped!

PR #678, opened 2025-10-08
Add MISO missing Load, Generation, and Interchange APIs

morindo:
## Summary
This pull request significantly expands the integration test coverage for the MISO API in `test_miso_api.py`. It introduces new tests for various "day ahead" and "real time" cleared and offered demand/generation endpoints, along with robust helper methods to validate DataFrame structure and data types.

The most important changes are:

**New integration tests for MISO API endpoints:**

* Added tests for day-ahead cleared demand (daily and hourly), day-ahead cleared generation (physical and virtual, hourly), and day-ahead offered generation (EcoMax/EcoMin, hourly) endpoints, each verifying correct DataFrame structure and date ranges.
* Added tests for real-time cleared demand (daily and hourly), real-time cleared generation (hourly, currently skipped due to data issues), real-time offered generation EcoMax (hourly), and real-time committed generation EcoMax (hourly), with assertions for columns, types, and date intervals.

**Helper methods for DataFrame validation:**

* Introduced reusable helper methods (e.g., `_check_test_get_day_ahead_cleared_demand`, `_check_day_ahead_cleared_generation_hourly`, etc.) to standardize validation of DataFrame columns and data types across tests.

**Test infrastructure and maintenance:**

* Used `api_vcr` cassettes for all new integration tests to ensure reproducibility and isolation from live API changes.
* Skipped a problematic real-time cleared generation test with a clear comment, documenting an upstream data issue with MISO.

### DIFF
--- a/gridstatus/miso_api.py
+++ b/gridstatus/miso_api.py
@@ -513,7 +513,7 @@ class MISOAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            date = "today"
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         return self._get_day_ahead_cleared_demand(
             date,
@@ -530,7 +530,7 @@ class MISOAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            date = "today"
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         return self._get_day_ahead_cleared_demand(
             date,
@@ -591,7 +591,7 @@ class MISOAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            date = "today"
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         return self._get_day_ahead_cleared_generation_hourly(
             date, end, verbose, generation_type="physical"
@@ -605,7 +605,7 @@ class MISOAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            date = "today"
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         return self._get_day_ahead_cleared_generation_hourly(
             date, end, verbose, generation_type="virtual"
@@ -619,7 +619,7 @@ class MISOAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            date = "today"
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         date_str = date.strftime("%Y-%m-%d")
 
@@ -720,7 +720,7 @@ class MISOAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            date = "today"
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         return self._get_day_ahead_offered_generation_hourly(
             date, end, verbose, ecotype="ecomax"
@@ -734,7 +734,7 @@ class MISOAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            date = "today"
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         return self._get_day_ahead_offered_generation_hourly(
             date, end, verbose, ecotype="ecomin"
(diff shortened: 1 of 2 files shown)


morindo:
Hi @WillKoehrsen, can you just tell me if I'm on the good path?

gridstatus/miso_api.py
```diff
@@ -452,6 +453,598 @@ def get_interchange_hourly(
...
+
+    @support_date_range(frequency="DAY_START")
+    def get_real_time_cleared_generation_hourly(
+        self,
+        date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
+        end: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp] | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        NOTE: This function is not ready for use yet. MISO Real-Time Cleared Generation API returns wrong data for a date.
```
WillKoehrsen: How were you able to figure out the data is wrong?

WillKoehrsen: Ohh, by wrong data do you mean the timestamps are wrong for the requested date? 

Testing the MISO API, I see when you request data for 2025-10-02 it gives you data from 2025-10-02T05:00:00 to 2025-10-03T04:00:00. 

I wonder if for some reason this endpoint is returning data in UTC instead of EST like all the other endpoints.

morindo: Oh, I didn't think about that. It seems to be the case, UTC instead of EST. I already contacted MISO about this, no response yet. Do you think we should convert these dates to EST instead?

WillKoehrsen: I'm not sure about this. It's just a guess based on the data starting at 05:00:00 and the EST offset being 5 hours from UTC. 

If we had a second source of generation we could compare. Maybe we can use the Generation Fuel Type with the assumption that Cleared Generation = sum(Generation Fuel Type) across all fuels? 

I'm a little out of my depth here; this is where I'd defer to @cwaldoch. 

@cwaldoch should the Real Time Cleared Generation be equal to the sum of the Real Time 
Generation Fuel Type? 

https://data-exchange.misoenergy.org/api-details#api=load-generation-and-interchange-api&operation=get-v1-real-time-date-generation-cleared-supply

https://data-exchange.misoenergy.org/api-details#api=load-generation-and-interchange-api&operation=get-v1-real-time-date-generation-fuel-type

morindo: Good point, I'll probably have a response from MISO this week. In the meantime, I'll check if I can validate the data against soon to be retired MISO market reports.

cwaldoch: > @cwaldoch should the Real Time Cleared Generation be equal to the sum of the Real Time Generation Fuel Type?

High level, when it comes to fuel mix reporting from MISO I would be careful assuming 1+1 =2 as they report (sometimes substantially) different fuel mix values for the same interval in different reports. 

In this case though, cleared supply doesn't necessarily have to match what's coming from the state estimator anyways (the real-time fuel type), so I wouldn't be surprised to see discrepancies.

WillKoehrsen: Thanks for the response. 

@morindo I think we can leave the function as is for now, merge this PR, then come back and update the function when we know more.

gridstatus/miso_api.py
```diff
@@ -452,6 +453,691 @@ def get_interchange_hourly(
...
+
+    @support_date_range(frequency="DAY_START")
+    def get_day_ahead_cleared_demand_daily(
+        self,
+        date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
+        end: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp] | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        if date == "latest":
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
```
WillKoehrsen, line 516: Thanks for adding the "latest" handling.

gridstatus/miso_api.py
```diff
@@ -452,6 +453,691 @@ def get_interchange_hourly(
...
+            if col not in ["Interval Start", "Interval End"]:
+                data[col] = data[col].astype(float)
+
+        data = data.sort_values(["Interval Start", "Interval End"])
+
+        return data[
+            ["Interval Start", "Interval End", "Generation Cleared"]
+        ].reset_index(drop=True)
+
+    @support_date_range(frequency="DAY_START")
```
WillKoehrsen, line 936: Added an issue to come back to this: https://github.com/gridstatus/gridstatus/issues/679

PR #686, opened 2025-10-14
Add actual load and medium term load forecast in MISO API.

morindo:
## Summary
This pull request adds support for retrieving actual load and medium-term load forecast data from the MISO API, both hourly and daily, and introduces corresponding integration tests to ensure correctness. The main changes include new methods in `gridstatus/miso_api.py` for fetching and processing this data, as well as comprehensive tests in `test_miso_api.py` to validate the new functionality.

**New Data Retrieval Methods:**

* Added `_get_actual_load`, `get_actual_load_hourly`, and `get_actual_load_daily` methods to fetch actual load data at hourly and daily resolutions from the MISO API.
* Added `_get_medium_term_load_forecast`, `get_medium_term_load_forecast_hourly`, and `get_medium_term_load_forecast_daily` methods to fetch medium-term load forecasts at hourly and daily resolutions, including support for specifying an initialization date for forecasts.

**Testing Enhancements:**

* Introduced integration tests for the new actual load and medium-term load forecast methods, including helper functions to check data structure and types, ensuring the API responses are parsed and formatted correctly.

**Dependency Updates:**

* Imported the `utils` module in `gridstatus/miso_api.py` to support date handling for the new forecast methods.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -942,6 +942,55 @@ class IESO(ISOBase):
         logger.debug(f"DataFrame Shape: {df.shape}")
         return df.sort_values(["Interval Start", "Publish Time", "Last Modified"])
 
+    def get_resource_adequacy_report_by_last_modified(
+        self,
+        last_modified: str | datetime.date | datetime.datetime,
+    ) -> pd.DataFrame:
+        """Retrieve and parse Resource Adequacy Reports modified after last_modified time.
+        This method bypasses date iteration and gets all files across all dates.
+        This is useful for ETL systems that want to get all new files at once.
+
+        Args:
+            last_modified: The last modified time after which to get report(s)
+            vintage: The version of the report to get
+
+        Returns:
+            pd.DataFrame: The Resource Adequacy Report df with all files modified after last_modified
+        """
+        if last_modified:
+            last_modified = utils._handle_date(last_modified, tz=self.default_timezone)
+
+        json_data_with_times = self._get_all_resource_adequacy_jsons_by_last_modified(
+            last_modified,
+        )
+        dfs = []
+        for json_data, file_last_modified in json_data_with_times:
+            try:
+                df = self._parse_resource_adequacy_report(json_data)
+                df["Last Modified"] = file_last_modified
+                dfs.append(df)
+            except Exception as e:
+                logger.warning(
+                    f"Failed to parse resource adequacy report: {str(e)}",
+                )
+                continue
+
+        if not dfs:
+            return pd.DataFrame()
+
+        df = pd.concat(dfs, ignore_index=True) if len(dfs) > 1 else dfs[0]
+        df = utils.move_cols_to_front(
+            df,
+            [
+                "Interval Start",
+                "Interval End",
+                "Publish Time",
+                "Last Modified",
+            ],
+        )
+        logger.debug(f"DataFrame Shape: {df.shape}")
+        return df.sort_values(["Interval Start", "Publish Time", "Last Modified"])
+
     # Note(Kladar): This might be fairly generalizable to other XML reports from IESO
     def _get_latest_resource_adequacy_json(
         self,
@@ -985,9 +1034,10 @@ class IESO(ISOBase):
                     tz=self.default_timezone,
                 )
             filtered_files = [
-                (file, time)
-                for file, time in files_and_times
-                if pd.Timestamp(time, tz=self.default_timezone) >= last_modified
+                (file, last_modified_time)
+                for file, last_modified_time in files_and_times
+                if pd.Timestamp(last_modified_time, tz=self.default_timezone)
+                >= last_modified
             ]
             logger.info(
                 f"Found {len(filtered_files)} files after last modified time {last_modified}",
@@ -1068,9 +1118,10 @@ class IESO(ISOBase):
                     tz=self.default_timezone,
                 )
             filtered_files = [
-                (file, time)
-                for file, time in file_rows
-                if pd.Timestamp(time, tz=self.default_timezone) >= last_modified
+                (file, last_modified_time)
+                for file, last_modified_time in file_rows
+                if pd.Timestamp(last_modified_time, tz=self.default_timezone)
+                >= last_modified
             ]
             logger.info(
                 f"Found {len(filtered_files)} files after last modified time {last_modified}",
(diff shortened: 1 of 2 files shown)


WillKoehrsen:
Hey @morindo, thanks for another contribution and sorry for the delay in reviewing. I will get to this today.

gridstatus/miso_api.py
```diff
@@ -1142,6 +1143,203 @@ def get_real_time_generation_fuel_type_hourly(
...
+        if end is not None:
+            data = data[data["Interval End"] <= end]
+
+        for col in data.columns:
+            if col not in ["Interval Start", "Interval End", "Region"]:
+                data[col] = data[col].astype(float)
+
+        return data[["Interval Start", "Interval End", "Region", "Load"]].reset_index(
+            drop=True
+        )
```
WillKoehrsen, line 1195: Can we sort values by "Interval Start" and "Region" before returning? We generally like to sort by the time first then the "subseries" if the dataset has one.

gridstatus/miso_api.py
```diff
@@ -1142,6 +1143,203 @@ def get_real_time_generation_fuel_type_hourly(
...
+    def _get_medium_term_load_forecast(
+        self,
+        date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
+        end: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp] | None = None,
+        verbose: bool = False,
+        init: str | pd.Timestamp | None = None,
+        time_resolution: str = HOURLY_RESOLUTION,
+    ) -> pd.DataFrame:
+        """
+        init: Optional init date to get forecast from.
```
WillKoehrsen: If I understand correctly, `init` is what we generally call `publish_time` so could we update this parameter to be called `publish_time`?

gridstatus/miso_api.py
```diff
@@ -1142,6 +1143,203 @@ def get_real_time_generation_fuel_type_hourly(
...
+        if date == "latest":
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d") + pd.Timedelta(
+                days=6
+            )  # Forecast date must be within 7 days of the init date.
+
+        date_str = date.strftime("%Y-%m-%d")
+
+        url = f"{BASE_LOAD_GENERATION_AND_INTERCHANGE_URL}/forecast/{date_str}/load?timeResolution={time_resolution}"
+        if init is not None:
+            utils._handle_date(init, self.default_timezone)
```
WillKoehrsen: We need to assign this to `init` right?

gridstatus/miso_api.py
```diff
@@ -1142,6 +1143,203 @@ def get_real_time_generation_fuel_type_hourly(
...
+    ) -> pd.DataFrame:
+        """
+        init: Optional init date to get forecast from.
+            It must be earlier than both:
+                - the current date (now)
+                - and the forecast date (the date for which you're requesting predictions).
+
+            If you don't specify this init date:
+                - When your requested date is before today, it defaults to date - 1 day
+                - When your requested date is today or in the future, it defaults to today - 1 day
```
WillKoehrsen, line 1241: We generally try to provide a "Publish Time" column with our forecast datasets. I think we can use this logic to provide at least an approximate publish time even though MISO doesn't provide it in the API. 

```python
miso_publish_time = min(
            date.normalize() - pd.DateOffset(days=1),
            pd.Timestamp.now(tz=self.default_timezone).normalize()
            - pd.DateOffset(days=1),
        )
df["Publish Time"] = init.normalize() if init is not None else miso_publish_time
```

(using the start of the day since we don't have a specific time).

morindo: Perfect, I will add this column.

gridstatus/miso_api.py
morindo: Yes, I'll change that to publish_time.

gridstatus/miso_api.py
morindo: **init** it optional here. We don't really need to assign-it. In the example text I describe the behaviour.

 If you don't specify this init date:
                - When your requested date is before today, it defaults to date - 1 day
                - When your requested date is today or in the future, it defaults to today - 1 day

- If today = 2025-10-14
                - You request forecasts for 2025-10-10 → init = 2025-10-09
                - You request forecasts for 2025-10-14 (today) → init = 2025-10-13
                - You request forecasts for 2025-10-15 (future) → init = 2025-10-13
                
 I could also use the same logic as MISO to create the Publish Time column.
 
 What do you think @WillKoehrsen ?

gridstatus/miso_api.py
```diff
@@ -1142,6 +1143,216 @@ def get_real_time_generation_fuel_type_hourly(
...
+        if date == "latest":
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d") + pd.Timedelta(
+                days=6
+            )  # Forecast date must be within 7 days of the init date.
+
+        date_str = date.strftime("%Y-%m-%d")
+
+        url = f"{BASE_LOAD_GENERATION_AND_INTERCHANGE_URL}/forecast/{date_str}/load?timeResolution={time_resolution}"
+        if publish_time is not None:
+            utils._handle_date(publish_time, self.default_timezone)
```
WillKoehrsen: This needs to be updated to assign the result to `publish_time`

```python
publish_time = utils._handle_date(publish_time, self.default_timezone)
```

otherwise this statement has no effect and you get the error `'str' object has no attribute 'strftime'` when calling `init` as a date string.

gridstatus/miso_api.py
```diff
@@ -1142,6 +1143,216 @@ def get_real_time_generation_fuel_type_hourly(
...
+            ]
+        ].reset_index(drop=True)
+
+    @support_date_range(frequency="DAY_START")
+    def get_medium_term_load_forecast_hourly(
+        self,
+        date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
+        end: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp] | None = None,
+        verbose: bool = False,
+        init: str | pd.Timestamp | None = None,
```
WillKoehrsen: Can we update this parameter to `publish_time` here and in `get_medium_term_load_forecast_daily` as well?

gridstatus/miso_api.py
```diff
@@ -1142,6 +1143,216 @@ def get_real_time_generation_fuel_type_hourly(
...
+        return data[
+            [
+                "Interval Start",
+                "Interval End",
+                "Region",
+                "Local Resource Zone",
+                "Load Forecast",
+                "Publish Time",
+            ]
+        ].reset_index(drop=True)
```
WillKoehrsen, line 1324: Can we put "Publish Time" as the third column and add `.sort_values(["Interval Start", "Publish Time", "Region"])`?

PR #690, opened 2025-10-16
Fix support date range decorators on miso pricing api

morindo:
## Summary
- Moved the **support_date_range** decorator on public functions instead of private ones. This allows passing additional keyword arguments (**kwargs) such as error="raise".

- Updated all get_lmp_* functions accordingly. I’ll apply similar changes to the get_mcp_* functions next.

Fixed three tests:

- Adjusted for cases where real-time data isn’t yet available at test execution time.

- Updated test_get_interchange_hourly_ to ensure compatibility on Windows.


WillKoehrsen (approved):
I don't remember why the decorator was placed on the private methods in the first place. The tests still pass locally with these changes 👍🏽

PR #691, opened 2025-10-18
[ENG-2516] Fix NYISO LMP 15 Minute not on 15 minute intervals

WillKoehrsen:
## Summary

- `NYISO().get_lmp(market=Markets.REAL_TIME_15_MIN)` was returning data not on 15 minute intervals (i.e. :10, :25, :40, :55). This was caused by upstream processing that treats the mixed data as all 5-minute data.
- This PR addresses the issue and adds a test to ensure the 15 minute data is on 15 minute intervals (:00, :15, :30, :45). The test fails on main and passes on this branch

### DIFF
--- a/gridstatus/nyiso.py
+++ b/gridstatus/nyiso.py
@@ -690,10 +690,15 @@ class NYISO(ISOBase):
             df.loc[~mask_15_min, "Market"] = Markets.REAL_TIME_5_MIN.value
             df.loc[mask_15_min, "Market"] = Markets.REAL_TIME_15_MIN.value
 
-            df.loc[mask_15_min, "Interval End"] = df.loc[
+            # For 15-min data, the original "Interval End" column contains the correct
+            # end time (since the raw data has timestamps as interval END). However,
+            # "Interval Start" was calculated assuming a 5-minute interval, so we need
+            # to recalculate it for 15-minute intervals.
+            # Interval Start = Interval End - 15 minutes
+            df.loc[mask_15_min, "Interval Start"] = df.loc[
                 mask_15_min,
-                "Interval Start",
-            ] + pd.Timedelta(
+                "Interval End",
+            ] - pd.Timedelta(
                 minutes=15,
             )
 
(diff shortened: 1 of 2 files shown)


gridstatus/nyiso.py
```diff
@@ -690,10 +690,15 @@ def _process_lmp_data(
             df.loc[~mask_15_min, "Market"] = Markets.REAL_TIME_5_MIN.value
             df.loc[mask_15_min, "Market"] = Markets.REAL_TIME_15_MIN.value
 
-            df.loc[mask_15_min, "Interval End"] = df.loc[
+            # For 15-min data, the original "Interval End" column contains the correct
```
WillKoehrsen, line 693: Here's a quick way to test. Run 

```bash
uv run python -c "from gridstatus.nyiso import *; df = NYISO().get_lmp('today', market=Markets.REAL_TIME_15_MIN, verbose=True); print(df.tail(20))"
```

and note the `url` as well as the "Interval Start" and "Interval End" in the dataframe. Download the file from the URL and confirm the "Interval End" values line up with the "Time Stamp" in the 15 minute data in the file (at the end).

Image

Image

The "Time" column in the dataframe shows what caused the off 15 minute intervals; earlier in the processing, we subtract 5 minutes from the "Time Stamp" in the file (which is the "Interval End") to get "Interval Start" for the 5 minute data (recall the 5 and 15 minute data is in the same file). This is the right "Interval Start" for 5 minute data but not for 15-minute data. 

The "Time" column is incorrect for the 15-minute data and I think should be dropped. I haven't done that yet because it would drop that column for all NYISO LMP data and I wanted to confirm that is okay.

WillKoehrsen: TL;DR at this point in the code, "Interval End" is correct and "Interval Start" is 10 minutes too late so we subtract 15 minutes from "Interval End" to get the correct "Interval Start".

WillKoehrsen: If you're wondering why don't we fix this earlier in the processing, it's because we don't know what is 5 minute and 15 minute data until a few lines prior to this code.

kmax12: the gift that keeps on giving. thanks for debugging this.

PR #698, opened 2025-10-28
feat(ENG 2613): Additional PJM datasets

Kladar:
## Summary
Adds:

- `pjm_generation_capacity_daily`
- `pjm_cleared_virtuals_daily`
- `pjm_inc_and_dec_bids_day_ahead_hourly`
- `pjm_sync_reserve_events`

### Details
Lots of paper cuts in these.

The `filter_timestamp_name` in `_get_pjm_json` assumes a `_ept` suffix, so it doesn't work for `cleared_virtuals_daily`. I have to go manual there. Probably not perfect but I don't trust modifying `_get_pjm_json` give its prevalence and the state of the tests. 

Sync reserves is a list not date-filterable from the API, and there are only ~1000 records going back to 2002, so I just return the whole list each time.

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -3465,10 +3465,16 @@ class PJM(ISOBase):
         https://dataminer2.pjm.com/feed/day_gen_capacity/definition
         """
         if date == "latest":
-            ten_days_ago = pd.Timestamp.now(tz=self.default_timezone) - pd.DateOffset(
-                days=10,
-            )
-            return self.get_generation_capacity_daily(ten_days_ago.date())
+            try:
+                return self.get_generation_capacity_daily("today")
+            except NoDataFoundException:
+                yesterday = (
+                    pd.Timestamp.now(tz=self.default_timezone).normalize()
+                    - pd.Timedelta(days=1)
+                ).date()
+                return self.get_generation_capacity_daily(
+                    pd.Timestamp(yesterday, tz=self.default_timezone),
+                )
 
         df = self._get_pjm_json(
             "day_gen_capacity",
--- a/gridstatus/tests/source_specific/test_pjm.py
+++ b/gridstatus/tests/source_specific/test_pjm.py
@@ -3167,8 +3167,8 @@ class TestPJM(BaseTestISO):
         with pjm_vcr.use_cassette("test_get_generation_capacity_daily_latest.yaml"):
             df = self.iso.get_generation_capacity_daily("latest")
             self._check_generation_capacity_daily(df)
-            ten_days_ago = self.local_today() - pd.Timedelta(days=10)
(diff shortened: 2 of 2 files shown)


gridstatus/pjm.py
```diff
@@ -3452,3 +3452,265 @@ def get_weight_average_aggregation_definition(
...
+            },
+            filter_timestamp_name="bid_datetime_beginning",
+            verbose=verbose,
+        )
+
+        df["Interval Start"] = (
+            pd.to_datetime(df["bid_datetime_beginning_utc"], format="ISO8601")
+            .dt.tz_localize("UTC")
+            .dt.tz_convert(self.default_timezone)
+        )
```
WillKoehrsen: Don't we already have the local time in "bid_datetime_beginning_ept" at this point?

gridstatus/pjm.py
```diff
@@ -3452,3 +3452,265 @@ def get_weight_average_aggregation_definition(
...
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieves the daily cleared virtual transactions data from:
+        https://dataminer2.pjm.com/feed/day_inc_dec_utc/definition
+        """
+        if date == "latest":
+            return self.get_cleared_virtuals_daily("today")
+
+        date = utils._handle_date(date, tz=self.default_timezone)
```
WillKoehrsen: This is already done by the decorator.

gridstatus/pjm.py
```diff
@@ -3452,3 +3452,265 @@ def get_weight_average_aggregation_definition(
...
+            )
+            delayed_date = delayed_date.replace(day=1)
+            return self.get_inc_and_dec_bids_day_ahead_hourly(delayed_date)
+
+        df = self._get_pjm_json(
+            "hrl_da_incs_decs",
+            start=date,
+            end=end,
+            params={
+                "fields": "bid_datetime_beginning_ept,bid_datetime_beginning_utc,price_point,inc_mw,dec_mw",
```
WillKoehrsen: Do we not want the Modified Datetime UTC to use as a publish time? Seems like useful information to have.

Image

WillKoehrsen: Maybe a question for @cwaldoch.

cwaldoch: I don't see a downside, I would be for that.

gridstatus/pjm.py
Kladar: yep. updated

gridstatus/pjm.py
Kladar: ah yeah. updated.

gridstatus/pjm.py
Kladar: added and renamed to Publish Time

PR #700, opened 2025-10-29
[ENG-2621] ISONE Capacity Forecast DST Fix

WillKoehrsen:
## Summary

- Fix `ISONEAPI().get_capacity_forecast_7_day()` for DST end 2025

### DIFF
--- a/gridstatus/tests/source_specific/test_isone_api.py
+++ b/gridstatus/tests/source_specific/test_isone_api.py
@@ -728,6 +728,9 @@ class TestISONEAPI(TestHelperMixin):
     def _check_capacity_forecast_7_day_columns(self, result: pd.DataFrame) -> None:
         """Validate the DataFrame columns against the Pydantic model fields."""
         assert list(result.columns) == ISONE_CAPACITY_FORECAST_7_DAY_COLUMNS
+        assert result["Interval Start"].dtype == "datetime64[ns, US/Eastern]"
+        assert result["Interval End"].dtype == "datetime64[ns, US/Eastern]"
+        assert result["Publish Time"].dtype == "datetime64[ns, US/Eastern]"
         assert result["High Temperature Boston"].dtype in [np.int64, np.float64]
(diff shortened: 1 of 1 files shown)


gridstatus/isone_api/isone_api.py
```diff
@@ -1073,8 +1073,12 @@ def _handle_capacity_forecast(
             meta=["CreationDate"],
         )
 
-        df["Publish Time"] = pd.to_datetime(df["CreationDate"])
-        df["Interval Start"] = pd.to_datetime(df["MarketDate"])
+        df["Publish Time"] = pd.to_datetime(df["CreationDate"], utc=True).dt.tz_convert(
+            self.default_timezone,
+        )
+        df["Interval Start"] = pd.to_datetime(df["MarketDate"], utc=True).dt.tz_convert(
```
WillKoehrsen, line 1079: Compare the output from:

```python
pd.to_datetime(['2025-11-02T00:00:00.000-04:00', '2025-11-03T00:00:00.000-05:00'])

FutureWarning: In a future version of pandas, parsing datetimes with mixed time zones will raise an error unless `utc=True`. Please specify `utc=True` to opt in to the new behaviour and silence this warning. To create a `Series` with mixed offsets and `object` dtype, please use `apply` and `datetime.datetime.strptime`

Index([2025-11-02 00:00:00-04:00, 2025-11-03 00:00:00-05:00], dtype='object')
```

to 

```python
pd.to_datetime(['2025-11-02T00:00:00.000-04:00', '2025-11-03T00:00:00.000-05:00'], utc=True)

DatetimeIndex(['2025-11-02 04:00:00+00:00', '2025-11-03 05:00:00+00:00'], dtype='datetime64[ns, UTC]', freq=None)
```

As a best practice, when datetime strings have a timezone offset, we should always pass `utc=True` because then Pandas knows how to handle mixed timezones such as happens during DST transitions.

PR #701, opened 2025-10-29
fix(ENG 2613): PJM DST transitions

Kladar:
## Summary
Actually using the utc output from the API seems to handle DST transitions much better. EPT seems to fall down in various ways. 

Correct for various DST transitions discovered in looking at historical data.

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -3649,6 +3649,12 @@ class PJM(ISOBase):
             },
         )
 
+        df["Publish Time"] = (
+            pd.to_datetime(df["Publish Time"], format="ISO8601")
+            .dt.tz_localize("UTC")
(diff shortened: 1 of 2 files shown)


gridstatus/pjm.py
```diff
@@ -3487,9 +3487,11 @@ def get_generation_capacity_daily(
             verbose=verbose,
         )
 
-        df["Interval Start"] = pd.to_datetime(
-            df["bid_datetime_beginning_ept"],
-        ).dt.tz_localize(self.default_timezone)
+        df["Interval Start"] = (
+            pd.to_datetime(df["bid_datetime_beginning_utc"], format="ISO8601")
```
WillKoehrsen, line 3491: Well sorry for my earlier suggestion to use the `ept` column.

PR #703, opened 2025-11-02
[ENG-2643] Fix ERCOT Energy Storage Resources and System Load Charging for DST End 2025

WillKoehrsen:
## Summary

- Fixes two Ercot datasets for DST end 2025
  - `Ercot().get_energy_storage_resources()` by preventing a duplicate timestamp
  - `ErcotAPI().get_system_load_charging_4_seconds()` by preventing duplicate timestamps

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -1357,7 +1357,14 @@ class ErcotAPI:
         # same system demand and charging values and must be duplicates because the odds
         # of these values being the same one hour apart is infinitesimal. So, we drop
         # one of rows 1 and 3 and don't use the UTC time.
-        data = data.drop_duplicates(subset=["AGCExecTime", "DSTFlag", "SystemDemand"])
+        if (
+            # DST transition check
+            pd.to_datetime(data["AGCExecTime"].min()).tz
+            != pd.to_datetime(data["AGCExecTime"].max()).tz
+        ):
+            data = data.drop_duplicates(
+                subset=["AGCExecTime", "DSTFlag", "SystemDemand"],
+            )
 
         data["AGCExecTime"] = pd.to_datetime(data["AGCExecTime"]).dt.tz_localize(
             self.default_timezone,


gridstatus/ercot_api/ercot_api.py
```diff
@@ -1348,23 +1348,32 @@ def get_system_load_charging_4_seconds(
...
+        # During the DST end transition, ERCOT publishes duplicate values like this:
+        # AGCExecTime         DSTFlag AGCExecTimeUTC       SystemDemand ESRChargingMW
+        # 2025-11-02T01:13:33 False   2025-11-02T07:13:33  42476.2188   2192.14233
+        # 2025-11-02T01:13:33 True    2025-11-02T07:13:33  41154.0469   1275.81592
+        # 2025-11-02T01:13:33 False   2025-11-02T06:13:33  42476.2188   2192.14233
+        # Even though the UTC times are different between rows 1 and 3, they have the
+        # same system demand and charging values and must be duplicates because the odds
+        # of these values being the same one hour apart is infinitesimal. So, we drop
+        # one of rows 1 and 3 and don't use the UTC time.
+        data = data.drop_duplicates(subset=["AGCExecTime", "DSTFlag", "SystemDemand"])
```
kmax12: The one hesitation I have is that clearly something is broken with this specific hour of data.

We are now adding a drop duplicates call that will run hundred thousand+ times throughout the year to solve a problem that occurs on a single hour. 

It not only seems wildly inefficient, but also feels like it adds risk to solve a very isolated problem.

Two ideas come to mind

1. do a one off fill of this hour right now and then leave the code as-is.

2. Only run the drop duplicates code when requested date range contains a dst change

gridstatus/ercot.py
```diff
@@ -367,26 +367,23 @@ def get_energy_storage_resources(
...
-            # ERCOT publishes two intervals with 2am timestamp
-            # during CDT to CST transition
-            # but skips the repeated 1am timestamp
-            # let's manually fix this before further timestamp parsing
-            df.loc[
-                (df["timestamp"] == "2024-11-03 02:00:00-0600")
-                & (df["dstFlag"] == "N"),
-                "timestamp",
-            ] = "2024-11-03 01:00:00-0600"
+        for timestamp in ["2024-11-03 02:00:00-0600", "2025-11-02 02:00:00-0600"]:
```
kmax12, line 370: This change looks good

gridstatus/ercot_api/ercot_api.py
kmax12: We do drop_duplicates in other places but two things that stand out as different here

1. This is 4s data, so more voluminous

2. It's specifically solving a dst problem. In other places, the issue could pop up any time of the year

WillKoehrsen: >Two ideas come to mind

> 1.  do a one off fill of this hour right now and then leave the code as-is.

> 2. Only run the drop duplicates code when requested date range contains a dst change

For 1., we are going to have this problem again next year. For 2. we have the overhead of the dst transition check on every run. I don't know how that overhead compares to the overhead of the `drop_duplicates` call.

WillKoehrsen: To test how much overhead the current solution adds, I ran

```python
%timeit -n 10 -r 2 ErcotAPI().get_system_load_charging_4_seconds('2025-11-01T00:00:00Z', '2025-11-01T01:00:00Z')
```

* On `main`: 1.09 s ± 63.4 ms per loop (mean ± std. dev. of 2 runs, 10 loops each)
* On this branch: 1.01 s ± 67 ms per loop (mean ± std. dev. of 2 runs, 10 loops each)

kmax12: on each option I'm proposing 

1. that is true, but that is also the case for the other fix in this PR. there is also a chance ERCOT fixes this bug next year

2. I probably should have focused less on performance and more on the maintainability. if there is a line of code (particularly one that a pretty wide-scope operation like `drop_duplicates`) that we can avoid running, 99.99% of the time, it's a big win for maintability. we have this `is_dst_end` function that is pretty lightweight. 

https://github.com/gridstatus/gridstatus/blob/05905f0e0e2011e5281ddcb8b52caa95025cef21/gridstatus/utils.py#L301-L302

my thought is to do something like: 

```python
if is_dst_end(date) or (end and is_dst_end(end):
   # drop duplicates logic.
```

this also help future reader understand why/when we are dropping duplicates beyond just a comment

kmax12: curious what direction you're leaning towards of the options on the table @WillKoehrsen?

kmax12: > To test how much overhead the current solution adds, I ran

also, you analysis doesn't look at memory, which is more often our bottleneck that causes server crashes.

still, i dont think this is about performance. the goal is to just make this code as safe as possible. I just always worry about unintended consequences of `drop_duplicates`

WillKoehrsen: Alright, I added a DST transition check

https://github.com/gridstatus/gridstatus/pull/703/commits/94b0721a5ccf4eac54466bda0284afe1f507c109

kmax12: Cool. That's a good way to do it! 

thanks

WillKoehrsen: I'm with you on avoiding unnecessary `drop_duplicates`.

PR #705, opened 2025-11-02
fix: ERCOT wind_actual_and forecast, solar_actual_and_forecast, catch ambiguous times

Kladar:
## Summary
Catch the ambiguity and make it explicitly True. For each ambiguous timestamp, pandas treats it as the first occurrence (DST/CDT, -05:00). This is an assumption, but we are sorting by Interval Start then Publish Time, so I think it is an okay assumption. 

Also has the advantage of only running that code path on DST transitions (the rest of days are unchanged). I do add a `reset_index` to the existing code as well to avoid confusion. 

```
from gridstatus.ercot_api.ercot_api import ErcotAPI
ercot = ErcotAPI()
import pandas as pd

# Here to show the wonkiness of the output aside from the DST transition on main
yesterday = pd.Timestamp.now(tz="US/Central") - pd.Timedelta(days=1)
df = ercot.get_wind_actual_and_forecast_hourly(date=yesterday)
print(df)

# Fails on main
df = ercot.get_wind_actual_and_forecast_hourly("today")
print(df)
```

### Details
Noting that these API endpoint themselves have always been slightly problematic in that ERCOT will occasionally return the same files twice when calling for a date range, resulting in duplicates. But that is the API behavior and so would be accounted for in downstream systems if need be.

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -404,6 +404,7 @@ class ErcotAPI:
             end_date=end,
             verbose=verbose,
             add_post_datetime=True,
+            include_source_filename=True,
         )
 
         return self._handle_wind_actual_and_forecast_hourly(
@@ -489,7 +490,7 @@ class ErcotAPI:
                 data,
                 ["Interval Start", "Interval End", "Publish Time"],
             )
-            .drop(columns=["Time", "postDatetime"])
+            .drop(columns=["Time", "postDatetime", "_source_filename"], errors="ignore")
             .sort_values(["Interval Start", "Publish Time"])
             .reset_index(drop=True)
         )
@@ -565,6 +566,7 @@ class ErcotAPI:
             end_date=end,
             verbose=verbose,
             add_post_datetime=True,
+            include_source_filename=True,
         )
 
         return self._handle_solar_actual_and_forecast_hourly(
@@ -601,7 +603,7 @@ class ErcotAPI:
                 data,
                 ["Interval Start", "Interval End", "Publish Time"],
             )
-            .drop(columns=["Time", "postDatetime"])
+            .drop(columns=["Time", "postDatetime", "_source_filename"], errors="ignore")
             .sort_values(["Interval Start", "Publish Time"])
             .reset_index(drop=True)
         )
@@ -1476,6 +1478,7 @@ class ErcotAPI:
         add_post_datetime: bool = False,
         verbose: bool = False,
         bulk_download: bool = True,
+        include_source_filename: bool = False,
         api: APITypeEnum = APITypeEnum.PUBLIC_API,
     ) -> pd.DataFrame:
         """Retrieves historical data from the given emil_id from start to end date.
@@ -1502,6 +1505,9 @@ class ErcotAPI:
             verbose [bool]: if True, will print out status messages
             bulk_download [bool]: if True, will download the data in batches
                 docIds. This is useful for avoiding rate limiting.
+            include_source_filename [bool]: if True, the returned dataframe will
+                include a column with the filename each row came from. Defaults to
+                False.
 
         Returns:
             [pandas.DataFrame]: a dataframe of historical data
@@ -1553,8 +1559,9 @@ class ErcotAPI:
             df = pd.read_csv(bytes_data, compression="zip")
             if add_post_datetime:
                 df["postDatetime"] = posted_datetime
-            # Store filename for xhr detection (prefer filename from zip, fallback to link)
-            df["_source_filename"] = filename if filename else link
+            if include_source_filename:
+                # Store filename for xhr detection (prefer filename from zip, fallback to link)
+                df["_source_filename"] = filename if filename else link
             dfs.append(df)
 
         return pd.concat(dfs)


gridstatus/ercot_api/ercot_api.py
```diff
@@ -416,9 +417,18 @@ def _handle_wind_actual_and_forecast_hourly(self, data, columns, verbose=False):
 
         data.columns = data.columns.str.replace("_", " ")
 
-        data["Publish Time"] = pd.to_datetime(data["postDatetime"]).dt.tz_localize(
-            self.default_timezone,
-        )
+        try:
+            data["Publish Time"] = pd.to_datetime(data["postDatetime"]).dt.tz_localize(
+                self.default_timezone,
+            )
```
Kladar, line 475: confirming that `ambiguous="infer"` does not work. 

We could potentially simply run `ambiguous=True` for all data and not catch the exception.

kmax12: I'm away from my computer, so I can't double check my understanding, but doesn't this dataset publish once an hour? So won't this approach necessarily overwrite data from the first occurrence of the hour during the repeated hour.

Kladar: In this case they are not quite published on the hour, and there is a (one hour and) 3 second difference. In this case that helps because they don't overwrite. But point taken.

```
i.     Interval Start                          Publish Time
2    2025-10-31 01:00:00-05:00 2025-11-02 01:55:10-05:00
3    2025-10-31 01:00:00-05:00 2025-11-02 01:55:13-05:00
```

My assumption is one of these is actually `01:55:XX-0600`, but it is difficult to figure out which one. Working on it..

Kladar: Hmm, actually they seem to have both `01:55:13-0500` and `02:55:13-0600` already represented with this approach. They seem to be separate forecasts. I'm not quite sure how they are handling it. I need to look deeper into the API response itself myself - Cursor doesn't seem to be able to come up with anything better even looking directly at the response. 

```
2025-11-08 19:00:00-06:00 | 2025-11-08 20:00:00-06:00 | 2025-11-02 01:55:13-05:00
2025-11-08 20:00:00-06:00 | 2025-11-08 21:00:00-06:00 | 2025-11-02 01:55:13-05:00
2025-11-08 21:00:00-06:00 | 2025-11-08 22:00:00-06:00 | 2025-11-02 01:55:13-05:00
2025-11-08 22:00:00-06:00 | 2025-11-08 23:00:00-06:00 | 2025-11-02 01:55:13-05:00
2025-11-08 23:00:00-06:00 | 2025-11-09 00:00:00-06:00 | 2025-11-02 01:55:13-05:00
2025-10-31 02:00:00-05:00 | 2025-10-31 03:00:00-05:00 | 2025-11-02 02:55:13-06:00
2025-10-31 03:00:00-05:00 | 2025-10-31 04:00:00-05:00 | 2025-11-02 02:55:13-06:00
2025-10-31 04:00:00-05:00 | 2025-10-31 05:00:00-05:00 | 2025-11-02 02:55:13-06:00
2025-10-31 05:00:00-05:00 | 2025-10-31 06:00:00-05:00 | 2025-11-02 02:55:13-06:00
2025-10-31 06:00:00-05:00 | 2025-10-31 07:00:00-05:00 | 2025-11-02 02:55:13-06:00
2025-10-31 07:00:00-05:00 | 2025-10-31 08:00:00-05:00 | 2025-11-02 02:55:13-06:00
2025-10-31 08:00:00-05:00 | 2025-10-31 09:00:00-05:00 | 2025-11-02 02:55:13-06:00
2025-10-31 09:00:00-05:00 | 2025-10-31 10:00:00-05:00 | 2025-11-02 02:55:13-06:00

```

kmax12: At some point if there is no obvious clean way, I am OK with a manual/one of fix and a note in the code and our postmortem that this data that will likely be an issue again next year

These are data sets that are used by many users and customers, so it is somewhat critical to get them running again

WillKoehrsen: > My assumption is one of these is actually 01:55:XX-0600, but it is difficult to figure out which one. Working on it..

@Kladar This is my understanding as well. Looking at the file list, it seems that 01:55:11 should be -0600 because its file name has xhr in it and from another comment in the codebase

> ERCOT adds xhr to the second set of file names during the repeated hour for DST end

Image

https://www.ercot.com/mp/data-products/data-product-details?id=np4-732-cd

WillKoehrsen: For this PR, I think it's fine if we manually assign -0600 to 01:55:11 and in a follow-up, maybe we can use that `xhr` to do this programmatically.

Kladar: Ok, it seems like the `postDatetime` from the files given by the API are actually naive, and the `DSTFlag` that comes along in those files refers to the `Interval Start` intervals rather than their publish time so it's not super useful for making the Publish Time timezone-aware. 

Instead of assigning them all CDT, I get the unique ambiguous Publish Times and assign one to be CDT and one to be CST. We are lucky this year they have different seconds values. Otherwise it might be very difficult to separate the rows. 

I assign `2025-11-02 01:55:10` to be the CDT and `2025-11-02 01:55:13` to be the CST 
```
 '2025-11-02T19:55:13.000' '2025-11-02T20:55:12.000']
                Interval Start              Interval End              Publish Time  ...  COP HSL LZ NORTH  STWPF LZ NORTH  WGRPP LZ NORTH
0    2025-10-31 00:00:00-05:00 2025-10-31 01:00:00-05:00 2025-11-02 00:55:11-05:00  ...             322.9           161.3           161.3
1    2025-10-31 01:00:00-05:00 2025-10-31 02:00:00-05:00 2025-11-02 00:55:11-05:00  ...             432.3           288.0           288.0
2    2025-10-31 01:00:00-05:00 2025-10-31 02:00:00-05:00 2025-11-02 01:55:10-05:00  ...             432.3           288.0           288.0
3    2025-10-31 01:00:00-05:00 2025-10-31 02:00:00-05:00 2025-11-02 01:55:13-06:00  ...             432.3           288.0           288.0
4    2025-10-31 02:00:00-05:00 2025-10-31 03:00:00-05:00 2025-11-02 00:55:11-05:00  ...             460.5           432.8           432.8
```

WillKoehrsen: Not sure if you saw my previous comment, but we can use the `xhr` in the file name to determine which is CST.

Screenshot 2025-11-02 at 21 42 49

2025-11-02 01:55:11 should be CST. According to Ercot conventions:

> ERCOT has appended an xhr indicator in the file naming convention to create file uniqueness across the repeated hour files generated. The first file generated will post with the expected naming convention. The second file generated will appear with the _xhr designation.

Kladar: I hadn't refreshed my page here before your comments. Yeah I will implement the xhr tomorrow if possible

Kladar: okay, I was able to see the `xhr` files get returned from the API and use that to determine the CDT/CST publish times (turns out it was reversed from my "just pick one" logic)

CleanShot 2025-11-03 at 09 38 35

Data is looking good now, and hopefully that solves these datasets into future transitions.
CleanShot 2025-11-03 at 09 40 36


I removed most of the logging used to discover all of this, back to the originally logging state. But I can add back in any that we find useful.

gridstatus/ercot_api/ercot_api.py
```diff
@@ -1462,10 +1542,19 @@ def get_historical_data(
...
+                bytes_data, filename = file_data
+            else:
+                bytes_data = file_data
+                filename = None
+
+            df = pd.read_csv(bytes_data, compression="zip")
             if add_post_datetime:
                 df["postDatetime"] = posted_datetime
+            # Store filename for xhr detection (prefer filename from zip, fallback to link)
+            df["_source_filename"] = filename if filename else link
```
WillKoehrsen: This new column breaks several existing tests where we check the exact set of columns.

Kladar: ah yeah not good. I corrected and made it an option to `get_historical_data()` with default false. I believe those that it broke are now passing.

PR #706, opened 2025-11-03
feat(ENG 2526): SPP Day Ahead and Real Time Binding Constraints

Kladar:
## Summary
Adds `get_binding_constraints_day_ahead_hourly()` and `get_binding_constraints_real_time_5_min()` to `SPP()`

```
from gridstatus.spp import SPP

spp = SPP()

df = spp.get_binding_constraints_day_ahead_hourly(date="latest")
print(df)

df2 = spp.get_binding_constraints_day_ahead_hourly(date="2025-11-01", end="2025-11-03")
print(df2)

df3 = spp.get_binding_constraints_real_time_5_min(date="latest")
print(df3)

df4 = spp.get_binding_constraints_real_time_5_min(date="2025-11-01", end="2025-11-02")
print(df4)
```

This also triggered "New Source Review" on `spp.py` where if we are adding new methods, I go back and update the old ones with non-functional changes, such as function typehints and using the `gs_logging.logger`. I also fixturize the existing tests with VCR.

### Details
The Real Time data has 3 types of files available: `Latest`, `By_Day` and `By_Interval`. The `latest` file is always named the same and conveniently covers the "latest" pathway for this dataset. `By_Interval` files are the ones that have the latest ~day's worth of data (the latest is the most recent interval itself) and are persistent, such that the latest and preceding intervals can be found here. 

However, the `By_Interval` are slow to query if you want multiple day's worth of data, so if you are querying anything other than today, the `By_Day` files get used. Essentially the top level function will figure out which files to choose, and the underlying methods are what use the `support_date_range` decorator for 5MIN and DAY START respectively. 

Day Ahead has neither the `Latest` nor `By_Interval` file trees, so that exclusively uses `By_Day` to get data. It is Day Ahead data published sometime "today", starting with the 0100 interval (thus the filenames all have `0100`). So `tomorrow` is typically available for `latest`, it could conceivably not be during the first hour of the day (thus the `latest` pathway falls back to using "today" if it encounters no data). 

### New Source Review
For the uninitiated (@Shiggiddie), "New Source Review" is the process by which the EPA evaluates upgrades to power plants in the US. If new regulations are passed, not all plants need to immediately upgrade (they get grandfathered in to the regulations). But, if the owner of the plant wants to make substantial upgrades to the plant, doing so can trigger New Source Review, where the plant is evaluated such that it is a new plant, and must meet all existing regulations. 

In the case here, if I'm adding new `get` methods to market file (e.g. `spp.py:SPP()`), then I tend to go through and update the whole file for python quality of life, which includes function typehints, using a `logging`-based logger, VCR fixturize existing tests, and attempt to get all tests passing. I take this approach because 
- we'd probably never really go back and do this upgrading/test passing as a separate task, especially across all market files since that's hundreds of methods and tests.
- it doesn't take all that long to tab through it in a single file + point cursor at the tests
- it only has to be done once per source file (and I've done most of them already)
- new methods going forward have these features, so it's nice to have a homogenized code style within a file, rather than sporadic quality changes in the same file

Evidently I hadn't done SPP yet. So now I have. I'll call out the dataset-specific changes in the comments. 

**Before Fixtures** `make test-spp (not slow)` ~2 minutes 
CleanShot 2025-11-03 at 15 48 13

**After VCR** `make test-spp (not slow)` ~10 seconds, 90%+ reduction on second run
CleanShot 2025-11-03 at 15 54 14

I make a couple test modifications that are checking against latest data latency of < 10 minutes, which we definitely aren't using in CI here and don't play well fixturized. I change these to be just that data is in the last 24 hours. When in doubt re-run your fixtures with env var `RECORD_MODE="all"`

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -1880,14 +1880,7 @@ class SPP(ISOBase):
             "Contingency Name",
         ]
 
-        return (
-            df[[c for c in cols_to_keep if c in df]]
-            .drop(
-                columns=["Time", "Interval"],
-                errors="ignore",
-            )
-            .sort_values(["Interval Start", "Constraint Name"])
-        )
+        return df[cols_to_keep].sort_values(["Interval Start", "Constraint Name"])
 
     @support_date_range("DAY_START")
     def get_binding_constraints_real_time_5_min(
@@ -2006,14 +1999,7 @@ class SPP(ISOBase):
             "Contingent Facility",
         ]
 
-        return (
-            df[[c for c in cols_to_keep if c in df]]
-            .drop(
-                columns=["Time", "Interval"],
-                errors="ignore",
-            )
-            .sort_values(["Interval Start", "Constraint Name"])
-        )
+        return df[cols_to_keep].sort_values(["Interval Start", "Constraint Name"])
 
 
 def process_gen_mix(df: pd.DataFrame, detailed: bool = False) -> pd.DataFrame:


gridstatus/spp.py
```diff
@@ -1732,8 +1811,205 @@ def _process_market_clearing(self, df: pd.DataFrame, interval_minutes: int):
 
         return df.sort_values("Interval Start")
 
+    @support_date_range("DAY_START")
```
Kladar, line 1814: New methods that this PR is actually about

gridstatus/tests/source_specific/test_spp.py
```diff
@@ -118,75 +118,89 @@ def _check_lmp_real_time_5_min_by_location(
...
         self._check_lmp_real_time_5_min_by_location(df)
 
         # Latest data should have one interval
-
         assert df["Interval Start"].nunique() == 1
-        assert df["Interval Start"].max() >= (self.now() - pd.DateOffset(minutes=10))
+        # Check that the max interval is relatively recent (within last 24 hours)
+        max_interval = df["Interval Start"].max()
+        assert max_interval >= self.local_start_of_today() - pd.Timedelta(days=1)
+        assert max_interval <= self.local_now() + pd.Timedelta(hours=1)
```
Kladar, line 134: Checking 10 minute latency in a test suite we don't look at much is a bit of putting the cart before the horse. Downstream systems can handle data latency detection.

gridstatus/tests/source_specific/test_spp.py
```diff
@@ -1447,3 +1576,98 @@ def test_market_clearing_day_ahead_date_range(self):
...
+                end=three_days_ago_0215,
+            )
+
+        self._check_binding_constraints_real_time(df)
+        assert df["Interval Start"].min() == three_days_ago
+        # Daily files return full day, so max is end of day
+        assert df["Interval Start"].max() == three_days_ago + pd.Timedelta(
+            hours=23,
+            minutes=55,
+        )
```
Kladar, line 1673: Actually new tests we care about

Kladar:
Need to iron out one thing regarding the real time interval file pathway

Shiggiddie (approved):

gridstatus/tests/source_specific/test_spp.py
WillKoehrsen: Makes sense, thanks for the cleanup.

gridstatus/spp.py
```diff
@@ -1732,8 +1811,203 @@ def _process_market_clearing(self, df: pd.DataFrame, interval_minutes: int):
...
+    def get_binding_constraints_day_ahead_hourly(
```
WillKoehrsen, line 1815: In the day ahead data, `NERC ID` is a float but in the real time data `NERC ID` is an integer. I believe this is because `NERC ID` is `NaN` for some values in the day ahead data and you need to use a special data type in Pandas to allow null values in an integer column, [`pd.Int64Dtype()`](https://pandas.pydata.org/docs/user_guide/integer_na.html).

Image

Kladar: Ah I see. Coercing it to that pd type. I see the `<NA>`s coming through, after that type forcing

CleanShot 2025-11-05 at 09 56 42

WillKoehrsen: Looks good.

gridstatus/spp.py
```diff
@@ -1732,8 +1811,212 @@ def _process_market_clearing(self, df: pd.DataFrame, interval_minutes: int):
...
+            "NERC ID",
+            "TLR Level",
+            "State",
+            "Shadow Price",
+            "Monitored Facility",
+            "Contingent Facility",
+        ]
+
+        return (
+            df[[c for c in cols_to_keep if c in df]]
```
WillKoehrsen: Are there sometimes different columns in the real time data? I think our approach in cases like that is to always return the same columns, adding missing ones as all nulls.

WillKoehrsen (approved):
One comment about columns in the real time data, otherwise, looks good.

gridstatus/spp.py
Kladar: ah not to my knowledge, this should probably just be an explicit return of columns to avoid that. Cursor likes to do a lot of `if column exists in df, do x` it seems that I have to correct or remove manually. And we typically want `do x, throw error if no column`.  I will update

Kladar: updated, running tests now...

PR #712, opened 2025-11-12
fix: NERC ID type enforcement

Kladar:
## Summary
Forces NERC ID in `spp_binding_constraints_real_time_5_min` to be an integer


gridstatus/spp.py
```diff
@@ -1985,6 +1985,7 @@ def _process_binding_constraints_real_time(self, df: pd.DataFrame) -> pd.DataFra
         )
 
         df = df.rename(columns={"NERCID": "NERC ID"})
+        df["NERC ID"] = pd.to_numeric(df["NERC ID"], errors="coerce").astype("Int64")
```
WillKoehrsen, line 1988: Any tests to update with this change?

Kladar: added

PR #714, opened 2025-11-13
fix: CAISO curtailment DST

Kladar:
## Summary
Couple issues from DST transition. 

```
df = caiso.get_curtailment(date="2025-11-01", end="2025-11-06")
print(df)
```

## Description
I'm not as familiar with this implementation (looks like a pretty manual parsing of an HTML report), so did the best to get past the issues causing scraper to fail. Hopefully not overboard in terms of handling of various cases. 

- Actual DST was running into an ambiguous error, so use some DST-safe interval padding in get_caiso_renewables_report by using concrete pd.Timedelta values for minute/hour/day series. Also generates timestamp ranges in UTC and converts back to the CAISO timezone to avoid AmbiguousTimeError during the fall-back hour.
- 11-03 had some oversized renewables arrays, padded with year-to-date zeroes, so trim them 
```
2025-11-03 curt_daily_econ_local_mw length: 337
first 5 values: ['0', '0', '0', '0', '0']
last 5 values: ['0', '0', '0', '0', '18.18']
```
- 11-05 seems to have no data and returns a 404

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -2798,10 +2798,16 @@ class CAISO(ISOBase):
         Fetches the CAISO daily renewable report for a given date and extracts data from
         all the charts into wide dataframes.
         """
-        report_url = f"https://www.caiso.com/documents/daily-renewable-report-{date.strftime('%b-%d-%Y').lower()}.html"
-
-        response = requests.get(report_url)
-
+        slug = date.strftime("%b-%d-%Y").lower()
+        primary_url = (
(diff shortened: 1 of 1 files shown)


WillKoehrsen:
> 11-05 seems to have no data and returns a 404

Not sure how far we want to go to get this data, but the 11-05 report has a different URL: 

https://www.caiso.com/documents/daily-renewable-report-nov-05-2025-corrected.html

WillKoehrsen (approved):
👍🏽 Thanks for fixing!

PR #716, opened 2025-11-16
[ENG-2773], [ENG-2776] SPP Load and Solar and Wind Forecasts Short Term Fix for DST End 2025

WillKoehrsen:
## Summary

- Fixes `SPP().get_load_forecast_short_term()` and `SPP().get_solar_and_wind_forecast_short_term()` for DST end
- Check with:
  - `SPP().get_solar_and_wind_forecast_short_term('2025-11-02 01:00:00-0500', '2025-11-02 02:00:00-0600')`
  - `SPP().get_load_forecast_short_term('2025-11-02 01:00:00-0500', '2025-11-02 02:00:00-0600')`
- Run specific tests with 
  - `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_spp.py -k "test_get_load_forecast_short_term_dst_ending or test_get_solar_and_wind_forecast_short_term_dst_ending"`

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -358,7 +358,7 @@ class SPP(ISOBase):
         self,
         date: pd.Timestamp,
         freq: str = "5min",
-    ) -> tuple[pd.Timestamp, bool]:
+    ) -> pd.Timestamp:
         """Handle DST transition when flooring a date.
 
         Args:
@@ -366,16 +366,14 @@ class SPP(ISOBase):
             freq: The frequency to floor to (e.g., "5min", "h")
 
         Returns:
-            tuple: (floored_date, is_repeated_hour)
+            Timestamp floored to the specified frequency
         """
-        is_repeated_hour = False
         try:
             floored_date = date.floor(freq)
         except pytz.AmbiguousTimeError:
-            is_repeated_hour = True
             floored_date = self.safe_for_dst_transition_floor(date, freq)
 
-        return floored_date, is_repeated_hour
+        return floored_date
 
     def _get_short_term_forecast_data(
         self,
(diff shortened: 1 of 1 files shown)


gridstatus/spp.py
```diff
@@ -355,6 +354,73 @@ def get_load_forecast_mid_term(
...
+            return None
+
+        floored_date, is_repeated_hour = self._handle_dst_floor_date(date, "5min")
+
+        hour = floored_date.hour
+        padded_hour = str(hour).zfill(2)
+        padded_hour_plus_one = str((hour + 1) % 24).zfill(2)
+
+        # The 100 file does not have a "d" but 105d through 155d do.
+        add_d = is_repeated_hour and not floored_date.minute == 0
```
WillKoehrsen: Here are the load forecast and solar and wind short term forecasts during the repeated hour. Note the "d" added to file names except for the one ending 100.csv

Image
Image

gridstatus/spp.py
```diff
@@ -369,8 +435,12 @@ def _post_process_load_forecast(
         # Assume the publish time is in the name of the file. There are different
         # times on the webpage, but these could be the posting time.
         df["Publish Time"] = pd.Timestamp(
-            url.split("-")[-1].split(".")[0],
+            re.search(r"[0-9]{12}", url).group(0),
```
WillKoehrsen, line 443: Switched to this regex to account for the "d" in the repeated file names.

gridstatus/spp.py
WillKoehrsen: Note that SPP does not publish two sets of reports during the 0100 hour so there is a gap in publish times. We would expect to see both 0105d, 0110d, 0115d, etc. and 0105, 0110, 0115, etc. if the repeated hour were handled correctly but we only see the reports with "d" appended.

gridstatus/spp.py
```diff
@@ -491,8 +558,12 @@ def _post_process_solar_and_wind_forecast(
...
+        ).tz_localize(
             tz=self.default_timezone,
+            # Assume the "d" file occurs during CST and Pandas wants ambiguous=True
```
WillKoehrsen, line 569: The "d" means CST assumption is based on an SPP document.

> CDT to CST occurs in the fall. During this transition there will be a repeated HE02 and all intervals within the repeated HE02 are also repeated. Files representing the repeated hour or interval because of the fall DST transition are identified by file names containing a “d” in the date/time portion of the file name. For example, the RTBM LMP by BUS data for the first interval 0105 during the fall DST transition is RTBM-LMP-B-201411020105.csv, while the data for the repeated 0105 interval is RTBM-LMP-B-201411020105d.cs

[spp weis public data guide v1.pdf](https://github.com/user-attachments/files/23571142/spp.weis.public.data.guide.v1.pdf)

gridstatus/spp.py
```diff
@@ -355,6 +354,78 @@ def get_load_forecast_mid_term(
...
+
+        # NOTE: this needs to be updated for DST every year
+        # 0105d through 0155d have a "d" on 2025-11-02
+        add_d = (
+            floored_date.year == 2025
+            and floored_date.month == 11
+            and floored_date.day == 2
+            and floored_date.hour == 1
+            and floored_date.minute != 0
+        )
```
WillKoehrsen, line 417: I updated this to just use a hardcoded date to reduce complexity. It does mean we have to add to this logic next year but as discussed, I think we are fine with this for some datasets.

PR #719, opened 2025-11-18
fix(ENG 2687): SPP Binding Constraint Column Formatting

Kladar:
## Summary
Normally the columns are Title Case, but a few times a year the SPP Binding Constraint Real Time 5 Min columns will be formatted differently (e.g. ALL CAPS). This handles those cases. 

CleanShot 2025-11-18 at 09 59 31@2x

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -2034,13 +2034,13 @@ class SPP(ISOBase):
 
         # Convert to title case to handle nonstandard input
         df.columns = df.columns.str.title().str.replace("_", " ")
-
-        if "Gmtintervalend" in df.columns:
-            df = df.rename(columns={"Gmtintervalend": "GMTIntervalEnd"})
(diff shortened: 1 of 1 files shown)


gridstatus/spp.py
```diff
@@ -2032,13 +2032,22 @@ def _get_binding_constraints_real_time_5_min_from_intervals(
     def _process_binding_constraints_real_time(self, df: pd.DataFrame) -> pd.DataFrame:
         df.columns = df.columns.str.strip()
 
+        # Convert to title case to handle nonstandard input
+        df.columns = df.columns.str.title().str.replace("_", " ")
+
+        if "Gmtintervalend" in df.columns:
```
WillKoehrsen: Nit: the `if` guard isn't necessary because `rename` by default does not error when keys in the dict are not columns in the dataframe.

PR #728, opened 2025-12-01
fix: miso lmp mapping instead of merge

Kladar:
## Summary
The columns are named slightly differently across the MISO and MISOAPI methods, but I believe they now have the same location parsing logic.

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -620,17 +620,16 @@ class MISO(ISOBase):
             data["Interval Start"] = pd.to_datetime(data["INTERVAL"]).dt.tz_localize(
                 self.default_timezone,
             )
-            print(data)
-            print(data.dtypes)
+
             node_to_type_mapping = (
                 MISO()
                 ._get_node_to_type_mapping()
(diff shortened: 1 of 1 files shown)


gridstatus/miso.py
```diff
@@ -621,14 +621,14 @@ def get_lmp(
...
-                right_on="Node",
-                how="left",
+            node_to_type_mapping = (
+                MISO()
+                ._get_node_to_type_mapping()
+                .set_index("Node")["Location Type"]
+                .to_dict()
             )
+            data["Location Type"] = data["CPNODE"].map(node_to_type_mapping)
+            data.rename(columns={"CPNODE": "Node"}, inplace=True)
```
Kladar, line 631: Need to rename to `Node` somewhere such that the rest of the method is unchanged.

Kladar: https://github.com/gridstatus/gridstatus/blob/68c422ce9f6a69b4e31d7ac2c9ac6ba3487a23dc/gridstatus/miso_api.py#L301-L308

PR #732, opened 2025-12-02
[ENG-2862] Stage RTC+B Changes

WillKoehrsen:
## Summary

- Stages the changes in preparation for RTC+B to go live on December 5
- Primarily for now removes the `_rtc_b_trial` from the methods
- We don't know exactly what the source will look like (for example, will the URLs change from the trial URLs) so expect there to be changes to this PR on December 5

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -145,8 +145,8 @@ SHORT_TERM_SYSTEM_ADEQUACY_REPORT_RTID = 12315
 # https://www.ercot.com/mp/data-products/data-product-details?id=NP6-323-CD
 REAL_TIME_ADDERS_AND_RESERVES_RTID = 13221
 
-# RTC Market Trials Real-Time ORDC and Reliability Deployment Price Adders
-REAL_TIME_ADDERS_RTC_B_TRIAL_RTID = 4108
+# Real-Time ORDC and Reliability Deployment Price Adders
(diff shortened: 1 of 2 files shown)


gridstatus/ercot.py
```diff
@@ -4339,65 +4337,15 @@ def _process_cop_adjustment_period_snapshot_60_day_data(
 
         return data
 
-    ### RTC + B Trials ###
-
     # Published every SCED interval
     @support_date_range(frequency=None)
-    def get_lmp_by_settlement_point_rtc_b_trial(
```
WillKoehrsen, line 4346: We already have a `get_lmp` method that can be called with `location_type = "Settlement Point"` to get the LMPs by settlement point. I don't know if that EMIL Report ID will stay the same once RTC+B goes live.

Shiggiddie: If this method is sufficiently different from `get_lmp`, does it make sense to leave this code in this PR (with a modification to the function name to `get_lmp_by_settlement_point`)?

I'm wondering if it is worth potentially prematurely deleting this code if we may need it come Thurs night?

WillKoehrsen: I think we should minimize the changes in this PR. We already use `get_lmp` with `location_type = "Settlement Point"` to get the LMP by settlement point dataset and since the logic to get this data isn't changing, I don't think we should use a different function.

PR #733, opened 2025-12-02
Add pricing nodes in MISO api

### LINKED ISSUE #682: Deprecation Risk: MISO()._get_node_to_type_mapping() Will Break After MISO Market Report Retirement
I recently discovered that **miso_api.py** relies on the internal function miso.py **MISO()._get_node_to_type_mapping()**, which downloads the daily YYYYMMDD__da_expost_lmp.csv market report each time a pricing API function is called to get the node Location Type.

After December 12, 2025, this approach **will no longer work**, as MISO is retiring all pricing market reports (including *_da_expost_lmp.csv).

At this point, it’s unclear how to replace this functionality, since the Data Exchange API does not appear to include node Location Type information in its available datasets.

https://github.com/gridstatus/gridstatus/blob/edc79f2ccd06acfc2853f0ea5233db62db07d011/gridstatus/miso_api.py#L286

morindo:
## Summary
This pull request adds new functionality to retrieve and process MISO pricing node data, enhances internal mapping logic, and introduces comprehensive integration tests for the new API methods. The main changes are grouped into API enhancements, internal logic improvements, and testing.

Fix #682 

**API Enhancements:**

* Added a new `get_pricing_nodes` method to the `miso_api.py` module, which retrieves MISO pricing node data for a specific date or date range, handling quarterly updates and returning a DataFrame with standardized columns (`Node`, `Location Type`).

* Added the `_miso_quarterly_days` helper function to compute quarterly anchor dates (March 1, June 1, September 1, December 1) within a given date range for accurate data retrieval.

**Internal Logic Improvements:**

* Updated `_get_node_to_type_mapping` to use the new `get_pricing_nodes` method, ensuring the mapping between nodes and location types is always current and based on the correct date range.

**Testing:**

* Introduced integration tests for `get_pricing_nodes`, covering retrieval for the latest date, a specific date, and a date range, with assertions on data shape and column types to ensure API correctness.

### DIFF
--- a/gridstatus/miso_api.py
+++ b/gridstatus/miso_api.py
@@ -1994,7 +1994,9 @@ class MISOAPI:
 
         return df_pivot
 
-    def _miso_quarterly_days(self, start, end):
+    def _miso_quarterly_days(
+        self, start: pd.Timestamp, end: pd.Timestamp
+    ) -> List[pd.Timestamp]:
         """
         MISO pricing nodes are updated quarterly on March 1st, June 1st, September 1st, and December 1st.
         This function generates a list of these dates within the specified start and end range.
@@ -2035,10 +2037,7 @@ class MISOAPI:
                  quarterly updates between date and end.
             verbose: If True, prints additional information during data retrieval.
         """
-        if date is None:
-            date = "latest"
-
-        if date == "latest":
+        if date == "latest" or date is None:
             date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
 
         end = (
--- a/gridstatus/tests/source_specific/test_miso_api.py
+++ b/gridstatus/tests/source_specific/test_miso_api.py
@@ -1149,7 +1149,6 @@ class TestMISOAPI(TestHelperMixin):
         with pytest.raises(NotSupported, match="only available for future dates"):
             self.iso.get_look_ahead_hourly(yesterday)
 
-    @pytest.mark.integration
     def test_get_pricing_nodes(self):
         today = pd.Timestamp(self.local_today())
 
@@ -1164,7 +1163,6 @@ class TestMISOAPI(TestHelperMixin):
             if col not in ["Node", "Location Type"]:
                 assert df[col].dtype == "str"
 
-    @pytest.mark.integration
     def test_get_pricing_nodes_by_date(self):
         today = pd.Timestamp(self.local_today())
         date = "latest"
@@ -1182,7 +1180,6 @@ class TestMISOAPI(TestHelperMixin):
             if col not in ["Node", "Location Type"]:
                 assert df[col].dtype == "str"
 
-    @pytest.mark.integration
     def test_get_pricing_nodes_by_date_range(self):
         today = pd.Timestamp(self.local_today())
         date = today - pd.Timedelta(days=100)


gridstatus/miso_api.py
```diff
@@ -1984,3 +1993,117 @@ def _process_as_mcp_data(
...
+        # If start is before all quarterly dates in range, look at previous year
+        previous_year_dates = [
+            pd.Timestamp(f"{start.year - 1}-{m}", tz=self.default_timezone)
+            for m in anchors
+        ]
+        return [max(previous_year_dates)]
+
+    def get_pricing_nodes(
+        self,
+        date: str | pd.Timestamp | None = None,
```
WillKoehrsen: Since `date` defaults to "latest", can we put that as the default in the function signature?

gridstatus/miso_api.py
```diff
@@ -1984,3 +1993,117 @@ def _process_as_mcp_data(
...
+            verbose: If True, prints additional information during data retrieval.
+        """
+        if date is None:
+            date = "latest"
+
+        if date == "latest":
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
+
+        if isinstance(date, str):
+            raise ValueError("Invalid date format")
```
WillKoehrsen: I think we should support `date` being a string by passing `date` through `utils._handle_date`.

gridstatus/miso_api.py
```diff
@@ -1984,3 +1993,117 @@ def _process_as_mcp_data(
...
+
+    def get_pricing_nodes(
+        self,
+        date: str | pd.Timestamp | None = None,
+        end: pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """
+        Retrieve MISO pricing nodes for a specific date or date range.
+        MISO pricing nodes change quarterly on March 1st, June 1st, September 1st, and December 1st.
```
WillKoehrsen, line 2031: Just wondering where this information came from?

gridstatus/miso_api.py
WillKoehrsen: I think we already do this later on so we may just be able to remove this guard.

gridstatus/miso_api.py
```diff
@@ -1984,3 +1993,117 @@ def _process_as_mcp_data(
         df_pivot.columns.name = None
 
         return df_pivot
+
+    def _miso_quarterly_days(self, start, end):
```
WillKoehrsen: I think this implementation can be simplified a little using Pandas' `date_range`:

```python
def _miso_quarterly_days(self, start, end):
    """
    MISO pricing nodes are updated quarterly on March 1st, June 1st, September 1st, and December 1st.
    This function generates a list of these dates within the specified start and end range.
    If no quarterly date falls within the range, return the most recent quarterly date before start.
    """
    dates = pd.date_range(
        # Default to the previous year so we always have a date 
        # before the start date
        start=f"{start.year - 1}-03-01",
        end=end,
        freq="QS-MAR",
        tz=self.default_timezone
    )
    
    dates_in_range = dates[(dates >= start) & (dates <= end)]
    if len(dates_in_range) > 0:
        return dates_in_range.tolist()
    
    return [dates[dates < start][-1]]
```

morindo: Done!

gridstatus/miso_api.py
morindo: This info is available here:

https://cdn.misoenergy.org/Quarterly%20EMS%20Model%20and%20ICCP%20schedules630916.pdf

image

Also, changes are also available in MISO Extranet in MISO Model Manager. Notice, that we cannot use this data in the public. It is prohibited by MISO.

image

gridstatus/miso_api.py
morindo: @WillKoehrsen let me know if it's okay, because I'm kind of lost with all the date management in the repo. `utlis._handle_date` does not seem to support "latest" keyword, it's support "today", should we use "today" instead? Also, is there a difference between "today" and "latest"?

the `@support_date_range` decorator seem to support "latest" but I'm not sure when to use what.

gridstatus/miso_api.py
WillKoehrsen: Thanks!

gridstatus/miso_api.py
WillKoehrsen: You've got it right. `utils._handle_date` doesn't handle "latest" so you need to add a line like

```python
if date == "latest":
    date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
```

before calling `_handle_date(date)`.

`@support_date_range` doesn't do any transformations of "latest" so it's up to the decorated function to handle "latest".

https://github.com/powerinsight/gridstatus/blob/fcb6ae1f493bc512cab242653ebdf4a3dd06399c/gridstatus/decorators.py#L149-L165

gridstatus/miso_api.py
```diff
@@ -1984,3 +1993,106 @@ def _process_as_mcp_data(
...
```
Kladar: nit: typehints

gridstatus/miso_api.py
```diff
@@ -1984,3 +1993,106 @@ def _process_as_mcp_data(
...
+                  Can be a pd.Timestamp or "latest".
+            end: Optional end date for a date range. If provided, retrieves pricing nodes for all
+                 quarterly updates between date and end.
+            verbose: If True, prints additional information during data retrieval.
+        """
+        if date is None:
+            date = "latest"
+
+        if date == "latest":
+            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
```
Kladar:
```suggestion
        if date == "latest" or date is None:
            date = pd.Timestamp.now(tz=self.default_timezone).floor("d")
```

gridstatus/tests/source_specific/test_miso_api.py
```diff
@@ -1147,3 +1148,55 @@ def test_get_look_ahead_hourly_past_date_raises_error(self):
...
+
+        assert df.shape[0] > 0
+
+        assert df.columns.tolist() == ["Node", "Location Type"]
+
+        for col in df.columns:
+            if col not in ["Node", "Location Type"]:
+                assert df[col].dtype == "str"
+
+    @pytest.mark.integration
```
Kladar: Since these use the vcr fixturization, they don't need to be marked as integration tests. 
```suggestion
```

gridstatus/tests/source_specific/test_miso_api.py
```diff
@@ -1147,3 +1148,55 @@ def test_get_look_ahead_hourly_past_date_raises_error(self):
 
         with pytest.raises(NotSupported, match="only available for future dates"):
             self.iso.get_look_ahead_hourly(yesterday)
+
+    @pytest.mark.integration
```
Kladar:
```suggestion
```

gridstatus/tests/source_specific/test_miso_api.py
```diff
@@ -1147,3 +1148,55 @@ def test_get_look_ahead_hourly_past_date_raises_error(self):
...
+
+        assert df.shape[0] > 0
+
+        assert df.columns.tolist() == ["Node", "Location Type"]
+
+        for col in df.columns:
+            if col not in ["Node", "Location Type"]:
+                assert df[col].dtype == "str"
+
+    @pytest.mark.integration
```
Kladar:
```suggestion
```

Kladar (approved):
LGTM - couple nits but method works and tests pass.

PR #735, opened 2025-12-03
[ENG-2876] ERCOT MCPC DAM

WillKoehrsen:
## Summary

- Adds `ErcotAPI().get_mcpc_dam()` which returns the output of `ErcotAPI().get_as_prices()` in a long format instead of wide (has an `AS Type` column instead of each AS Type being a separate column)
- Run specific tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot_api.py -k get_mcpc_dam`
- Make sure there are no regressions with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot_api.py -k as_prices`

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -640,15 +640,15 @@ class ErcotAPI:
         verbose: bool = False,
     ) -> pd.DataFrame:
         if date == "latest":
-            return self.get_as_prices("today", verbose=verbose)
+            return self.get_mcpc_data("today", verbose=verbose)
 
         end = self._handle_end_date(date, end, days_to_add_if_no_end=1)
 
         if self._should_use_historical(date):
             data = self.get_historical_data(
-                # Need to subtract 1 because we filter by posted date and this data
-                # is day-ahead
                 endpoint=AS_PRICES_ENDPOINT,
+                # Need to subtract 1 day because we filter by posted date in thie
+                # historical api and this data is day-ahead
                 start_date=date - pd.Timedelta(days=1),
                 end_date=end - pd.Timedelta(days=1),
                 verbose=verbose,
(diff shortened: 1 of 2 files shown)


gridstatus/ercot.py
```diff
@@ -1851,6 +1851,39 @@ def get_as_prices(
...
+
+        df["AS Type"] = df["AS Type"].map(
+            {
+                "ERCOT Contingency Reserve Service": "ECRS",
+                "Non-Spinning Reserves": "NSPIN",
+                "Regulation Down": "REGDN",
+                "Regulation Up": "REGUP",
+                "Responsive Reserves": "RRS",
+            },
+        )
```
WillKoehrsen: These are abbreviated to match the `AS Type` values in the other `mcpc` datasets.

gridstatus/ercot.py
```diff
@@ -1851,6 +1851,39 @@ def get_as_prices(
...
+        return dataframe is long instead of wide.
+        """
+        wide_df = self.get_as_prices(date, end, verbose)
+        df = wide_df.melt(
+            id_vars=["Time", "Interval Start", "Interval End", "Market"],
+            var_name="AS Type",
+            value_name="MCPC",
+        )
+
+        df["AS Type"] = df["AS Type"].map(
```
kmax12: I appreciate reusing the existing method, but this feels a bit roundabout given that we convert from the narrow to long format, just to change it back here

https://github.com/gridstatus/gridstatus/blob/b79b5b348fcf638e1b2734e6122ce9ca2371d943/gridstatus/ercot.py#L2257-L2264

kmax12: I believe the raw data is in the schema as the real time MCPC trial data, so I think we can probably reuse a fair amount of code from those methods

gridstatus/ercot.py
```diff
@@ -1851,6 +1851,39 @@ def get_as_prices(
...
+            {
+                "ERCOT Contingency Reserve Service": "ECRS",
+                "Non-Spinning Reserves": "NSPIN",
+                "Regulation Down": "REGDN",
+                "Regulation Up": "REGUP",
+                "Responsive Reserves": "RRS",
```
Shiggiddie: Does it make sense to re-use the rename map from `_finalize_as_price_df`?

PR #736, opened 2025-12-04
[ENG-2868] Handle Renaming of DSTFlag to RepeatedHourFlag

WillKoehrsen:
## Summary

- Handles the renaming of `DSTFlag` to `RepeatedHourFlag` that will occur as part of ERCOT's RTC+B changes


kmax12 (approved):
good catch!

PR #738, opened 2025-12-05
Change ERCOT Real Time Adders Report ID

WillKoehrsen:
## Summary

- The ERCOT real time adders is being published under the same report ID as what was previously adders and reserves. This PR changes the report ID for the adders from the trial ID to the correct report ID


Shiggiddie (approved):

gridstatus/ercot.py
```diff
@@ -142,11 +142,13 @@
 
 # Real-Time ORDC and Reliability Deployment Price Adders and Reserves by SCED Interval
 # (ORDC = Operating Reserve Demand Curve)
+# Adders and reserves stopped being published on December 5
```
WillKoehrsen, line 145: I'll come back later and add a useful error message when the real time adders and reserves method is called with a date on or after December 5. The method errors on this date currently but with a non-informative method.

PR #742, opened 2025-12-06
Fix ERCOT RTC Datasets for Queries that Span Two Days

WillKoehrsen:
## Summary

- Fixes new ERCOT RTC datasets for queries that span two days
- The problem was we passed `published_after`, `published_before`, and `date` to `_get_documents` and the `date` parameter filtered out documents from the second day. This is solved by setting `date=None` when we have `published_before` and `published_after`
- Adds new tests for these cases
  - `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot.py -k "mcpc_sced_date_range or mcpc_real_time_15_min_date_range or indicative_mcpc_rtd_date_range or as_total_capability_date_range or real_time_adders_date_range"`

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -4480,10 +4480,14 @@ class Ercot(ISOBase):
         verbose: bool = False,
     ) -> pd.DataFrame:
         """Get Market Clearing Prices for Capacity by SCED interval"""
-        published_before = None
-        published_after = None
-
-        if date != "latest":
+        if date == "latest":
+            docs = self._get_documents(
+                report_type_id=REAL_TIME_CLEARING_PRICES_FOR_CAPACITY_BY_SCED_INTERVAL_RTID,
+                extension="csv",
+                date=date,
+                verbose=verbose,
+            )
+        else:
             if end is None:
                 # Assume getting data for one day
                 end = date + pd.DateOffset(days=1)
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -4491,6 +4491,9 @@ def get_mcpc_sced(
             published_before = end
             published_after = date
 
+            # Since we are filtering by publish before and end, set date to None
+            date = None
```
kmax12: I'm trying to think if this is the cleanest fix and if there is a fix that would avoid such issues in the future. 

do we ever **_need_** to provide both `date` and `published_before`/`after` to `_get_documents`?

the bug here is because they both do different filtering on the same property and combine in unintuitive ways.

https://github.com/gridstatus/gridstatus/blob/745e2c6faf9dddcc1c6331d7f058f145081e9cc8/gridstatus/ercot.py#L3881-L3886

https://github.com/gridstatus/gridstatus/blob/745e2c6faf9dddcc1c6331d7f058f145081e9cc8/gridstatus/ercot.py#L3901-L3904

it seems like we could raise an error if both are provided together since I can't come up with a reason where that would be necessary

kmax12: I don't know if we want to do this, but to illustrate my point — we could remove `date` and migrate all call to use `published_before` and `published_after` while accomplishing the same functionality

kmax12: I see only 4 places where we provide both `date` and `published_before`/`after` to _get_documents. it's only these new method. i'd think it's safe to raise an error without breaking any existing code

Image

WillKoehrsen: Yeah, I thought about changing `_get_documents` but decided against it because it's called in many other places.

kmax12: I don't see it called with all three parameters anywhere else though. But maybe I missed in my quick scan

WillKoehrsen: Even if we make that change to `_get_documents`, we still have to make the change to this method to set `date=None`.

kmax12: is there a way to get this method to never pass date to get_documents?

kmax12: I feel like that leads to the most predictable behavior

kmax12: Oh, I guess we still need to support latest.

kmax12 (approved):
approving this since it fixes the issue at hand. 

i think there is a long running and pre-existing bad api design of `_get_documents` that made an error like easy/possible, but it's understable that we want to fix that outside the scope of this PR

gridstatus/ercot.py
WillKoehrsen: I updated the flow of all these methods so we are not passing `published_before`, `published_after`, and `date` at the same time to `_get_documents`. I think it's a cleaner approach.

https://github.com/gridstatus/gridstatus/pull/742/commits/97643890cafaa559915a2884e1ecb02378943775

kmax12: Nice. That flow feels a lot safer

PR #745, opened 2025-12-08
[ENG-2920], [ENG-2921] IESO Intertie Limits

WillKoehrsen:
## Summary

- Adds `IESO().get_intertie_limits_real_time()` and `IESO().get_intertie_limits_dam()`
- Run tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ieso.py -k intertie_limits`

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -2240,7 +2240,7 @@ class IESO(ISOBase):
         return df, publish_time
 
     @support_date_range(frequency="HOUR_START")
-    def get_intertie_limits_real_time(
+    def get_intertie_limits_real_time_5_min(
         self,
         date: str | pd.Timestamp | tuple[pd.Timestamp, pd.Timestamp],
         end: pd.Timestamp | None = None,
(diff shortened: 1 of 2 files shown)


gridstatus/ieso.py
```diff
@@ -2113,6 +2114,214 @@ def _get_and_parse_intertie_schedule_flow(
...
+            DataFrame with columns for interval start/end and import/export
+            limits for each intertie zone
+        """
+        directory_path = "DAIntertieSchedLimits2"
+        file_directory = f"{PUBLIC_REPORTS_URL_PREFIX}/{directory_path}"
+
+        if date == "latest":
+            url = f"{file_directory}/PUB_{directory_path}.xml"
+        else:
+            url = f"{file_directory}/PUB_{directory_path}_{date.strftime('%Y%m%d')}.xml"
```
Kladar, line 2312: From what I can tell the `_v1`-style files are the same as the unversioned files, so this is probably just IESO using their same style of file production as other continually updating version files but just with a single version per day. 

https://reports-public.ieso.ca/public/RealtimeIntertieSchedLimits/

WillKoehrsen: Yeah, I compared the v1 files to the non-versioned files and they are the same so we are not using the versioned files. There also are `.csv` files for the real time dataset, but I choose to use the XML to share parsing with the day ahead dataset which does not have csv files.

PR #750, opened 2025-12-10
Update MISO API Retry

WillKoehrsen:
## Summary

- Updates the retry mechanism for the MISO API
- Uses `exponential_base` (default 2) for exponential decay instead of `initial_sleep_seconds`
- Allows specifying `max_retries` on initialization
- Only retry on 429 and 5xx responses
- Adds retries for the first query
- Adds testing for the retry approach

### DIFF
--- a/gridstatus/miso_api.py
+++ b/gridstatus/miso_api.py
@@ -1729,7 +1729,7 @@ class MISOAPI:
     ) -> requests.Response:
         """Make a request with exponential backoff retry logic.
 
-        Only retries on 429 (Too Many Requests) and 503 (Service Unavailable) errors.
+        Only retries on 429 (Too Many Requests) and 5xx server errors.
         Other HTTP errors are raised immediately without retry.
 
         Arguments:
@@ -1745,7 +1745,6 @@ class MISOAPI:
             requests.HTTPError: If a non-retryable error occurs or all retries
                 are exhausted.
         """
-        retryable_status_codes = {429, 503}
         last_exception: Exception | None = None
 
         for attempt in range(self.max_retries + 1):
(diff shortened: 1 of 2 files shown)


gridstatus/miso_api.py
```diff
@@ -1694,38 +1705,80 @@ def _get_url(
...
+            verbose: Whether to log verbose output.
+
+        Returns:
+            The successful response.
+
+        Raises:
+            requests.HTTPError: If a non-retryable error occurs or all retries
+                are exhausted.
+        """
+        retryable_status_codes = {429, 503}
```
WillKoehrsen: We should only retry on 429 (rate limit) and 503 (service unavailable) responses. A 400 response indicates there's no data available in which case a retry is not useful.

gridstatus/miso_api.py
```diff
@@ -41,15 +41,23 @@ def __init__(
...
+        exponential_base: int = 2,
     ) -> None:
         """
         Class for querying the MISO API. Currently supports only pricing data.
 
         Arguments:
         pricing_api_key (str): The API key for the pricing API. Can be a comma-separated
-        list of keys if you have multiple keys.
+            list of keys if you have multiple keys.
         initial_sleep_seconds (int): The number of seconds to wait between each request.
```
WillKoehrsen, line 53: `initial_sleep_seconds` isn't the best name for this parameter. However, I didn't want to introduce a breaking change to the API.

gridstatus/miso_api.py
kmax12: good addition. are there are 500s we need to catch?

WillKoehrsen: I've only seen 503s but we could include all 50x.

PR #754, opened 2025-12-12
[ENG-2934] ERCOT Highest Price as Offer Selected DAM and SCED

WillKoehrsen:
## Summary

- Adds `Ercot().get_highest_price_as_offer_selected_dam()` and `Ercot().get_highest_price_as_offer_selected_sced()`
- Run tests with ` VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot.py -k highest_price_as_offer_selected -s`

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -266,10 +266,6 @@ UNPLANNED_RESOURCE_OUTAGES_REPORT_RTID = 22912
 
 # 3-Day Highest Price AS Offer Selected
 # https://www.ercot.com/mp/data-products/data-product-details?id=NP3-915-EX
-THREE_DAY_HIGHEST_PRICE_AS_OFFER_SELECTED_RTID = 13018
-
-# 3-Day DAM Highest Price AS Offer Selected
-# https://www.ercot.com/mp/data-products/data-product-details?id=NP3-915-EX
 THREE_DAY_DAM_HIGHEST_PRICE_AS_OFFER_SELECTED_RTID = 13018
 
 # 3-Day SCED Highest Price AS Offer Selected
@@ -3687,7 +3683,7 @@ class Ercot(ISOBase):
         report_date = date.normalize() + pd.DateOffset(days=3)
 
         doc = self._get_document(
-            report_type_id=THREE_DAY_HIGHEST_PRICE_AS_OFFER_SELECTED_RTID,
+            report_type_id=THREE_DAY_DAM_HIGHEST_PRICE_AS_OFFER_SELECTED_RTID,
             date=report_date,
             verbose=verbose,
         )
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -268,6 +268,14 @@ class ERCOTSevenDayLoadForecastReport(Enum):
 # https://www.ercot.com/mp/data-products/data-product-details?id=NP3-915-EX
 THREE_DAY_HIGHEST_PRICE_AS_OFFER_SELECTED_RTID = 13018
 
+# 3-Day DAM Highest Price AS Offer Selected
+# https://www.ercot.com/mp/data-products/data-product-details?id=NP3-915-EX
+THREE_DAY_DAM_HIGHEST_PRICE_AS_OFFER_SELECTED_RTID = 13018
```
Shiggiddie: It looks like this ID has already been hardcoded in `THREE_DAY_HIGHEST_PRICE_AS_OFFER_SELECTED_RTID`

Confirming it is desirable to add a new ID instead of reuse?

gridstatus/ercot.py
```diff
@@ -3719,25 +3883,63 @@ def _handle_offers(df: pd.DataFrame) -> pd.Series:
...
-        df = (
-            df.groupby(
-                [
-                    "Time",
+        if is_sced_format:
+            # SCED format: Parse SCED Timestamp directly
+            # SCED Timestamp format is like "12/07/2025 00:00:17"
+            df["SCED Timestamp"] = pd.to_datetime(df["SCED Timestamp"]).dt.tz_localize(
+                self.default_timezone,
+            )
```
Shiggiddie, line 3888: presumably this allows us to not have to concern ourselves with DST?

gridstatus/ercot.py
```diff
@@ -3672,44 +3692,188 @@ def get_highest_price_as_offer_selected(
...
+                    "Offered Price",
+                    "Total Offered Quantity",
+                    "Offered Quantities",
+                ]
+            ]
+            .sort_values(["SCED Timestamp", "AS Type", "Resource Name"])
+            .reset_index(drop=True)
+        )
+
+    def _handle_highest_price_as_offer_selected_file(
```
Shiggiddie, line 3843: Given the different handling between sced & dam, this could be split into separate functions entirely

gridstatus/ercot.py
WillKoehrsen: It's a little confusing having this defined twice so I will remove one.

gridstatus/ercot.py
WillKoehrsen: This dataset is new and has no example data during a DST transition (either start or end). I've added the `ambiguous` argument to hopefully avoid issues during DST end.

PR #757, opened 2025-12-16
feat(ENG 2941, ENG 2942, ENG 2261): PJM Retention Datasets

Kladar:
## Summary
Adds
```
PJM().get_marginal_emission_rates_5_min()
PJM().get_pai_intervals_5_min()
PJM().get_voltage_limits()
```

Test with
```
VCR_RECORD_MODE=all uv run pytest -vv gridstatus/tests/source_specific/test_pjm.py -k "pai_intervals_5_min or marginal_emission_rates_5_min"
```
The voltage limit test will get throttled if you run both the test and the command; See more below.

### Details
The data miner datasets are straightforward, but the voltage limits CSV had a bunch of pitfalls that were interesting to work around.

First, when attempting to get the file manually, I discovered PJM is blocking most network requests and I had to eventually `curl --interface en0` to get around my VPN and other browser network interface settings to actually get through to the file. 

Next, the source URL (https://edart.pjm.com/reports/voltagelimits.csv) actually returns a `.zip` file. So I use `utils.get_zip_folder` even though it has a `.csv` url.

Finally, PJM does an interesting throttling where if you request the file from the same IP address within the 30 minutes, it will still return a zip with a CSV, but that CSV will contain the throttling message:
```
voltagelimits.csv file requested by this IP Address has not been updated since your last successful attempt on 12/16/2025 13:14:51 (throttle 1800 seconds).  For any questions, please contact eDARTHelp@pjm.com.
```

So there is no way to make this method work on subsequent calls within the 30 minute limit without doing some pretty serious IP rotation and network interface spoofing, which I would imagine is against PJM terms of use. I just send the user a warning if subsequent calls are made. This also poses a problem for running the method and the test, given that they each make a call to the endpoint. Technically you need to run the test first to get a fixture, after which you can avoid the call to source on the test.

Run these spaced out by 30 minutes if you feel so inclined:
```
$ VCR_RECORD_MODE=all uv run pytest -vv gridstatus/tests/source_specific/test_pjm.py::TestPJM::test_get_voltage_limits_latest
```
```
df = PJM().get_voltage_limits()
```

### DIFF
--- a/gridstatus/tests/source_specific/test_pjm.py
+++ b/gridstatus/tests/source_specific/test_pjm.py
@@ -3410,16 +3410,15 @@ class TestPJM(BaseTestISO):
             today = pd.Timestamp.now(tz=self.iso.default_timezone).date()
             assert df["Interval Start"].max().date() >= today - pd.Timedelta(days=1)
 
-    @pytest.mark.slow
     def test_get_marginal_emission_rates_5_min_historical_date_range(
         self,
     ):
-        date = self.local_today() - pd.Timedelta(days=10)
+        date = self.local_start_of_today() - pd.Timedelta(days=10)
         end = date + pd.Timedelta(hours=1)
         with pjm_vcr.use_cassette(
             f"test_get_marginal_emission_rates_5_min_historical_{date.strftime('%Y-%m-%d')}_{end.strftime('%Y-%m-%d')}.yaml",
         ):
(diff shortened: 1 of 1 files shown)


WillKoehrsen (approved):
Code looks good and tests pass locally. Thanks for explaining the rate limits on the voltage file.

PR #759, opened 2025-12-18
fix(ENG 2764): Get all FCM Auctions

Kladar:
## Summary
Previously it would only get ARA1. Now gets all 3 auctions for a commitment period.

2019 and 2020 auctions are not available from the new endpoint however, so that falls back to the previous method

```
df = ISONEAPI().get_fcm_reconfiguration_annual("latest")
df = ISONEAPI().get_fcm_reconfiguration_annual("2025-01-01")
df = ISONEAPI().get_fcm_reconfiguration_annual("2020-01-01")
```

### DIFF
--- a/gridstatus/isone_api/isone_api.py
+++ b/gridstatus/isone_api/isone_api.py
@@ -2028,6 +2028,9 @@ class ISONEAPI:
             if column in df.columns:
                 df[column] = pd.to_numeric(df[column], errors="coerce")
 
+        if "ARA" in df.columns:
+            df["ARA"] = pd.to_numeric(df["ARA"], errors="coerce").astype("Int64")
+
         df = df.reindex(columns=ISONE_FCM_RECONFIGURATION_COLUMNS)
         df = df.sort_values(
             ["Interval Start", "ARA", "Location ID"],


gridstatus/isone_api/isone_api.py
```diff
@@ -1887,6 +1967,12 @@ def _parse_fcm_reconfiguration_dataframe(
         }
 
         for auction, interval_start, interval_end in auction_records:
+            auction_type = auction.get("Type", "")
+            ara_value = (
+                auction_type.replace("ARA", "")
+                if auction_type.startswith("ARA")
+                else ""
+            )
```
WillKoehrsen, line 1975: Should we convert ARA to an integer here if it can only take on values "1", "2", "3".

WillKoehrsen (approved):

gridstatus/isone_api/isone_api.py
```diff
@@ -1807,49 +1807,129 @@ def get_fcm_reconfiguration_annual(
...
+
+                            if begin and end_date:
+                                interval_start = pd.Timestamp(begin)
+                                interval_end = pd.Timestamp(end_date)
+                            else:
+                                description = auction.get("Description")
+                                interval_start = (
+                                    pd.Timestamp(description)
+                                    if description
+                                    else pd.Timestamp(auction.get("ApprovalDate"))
```
Kladar, line 1900: Just realized how hideous this is lol

PR #760, opened 2025-12-19
fix: only add and sort by ARA if annual

Kladar:
## Summary
Fix FCM Monthly Auction parsing. Kind of a bandaid in lieu of refactoring the parsing method more significantly

### DIFF
--- a/gridstatus/isone_api/isone_api.py
+++ b/gridstatus/isone_api/isone_api.py
@@ -1797,7 +1797,10 @@ class ISONEAPI:
             interval_end = interval_start + pd.DateOffset(months=1)
             annotated_auctions.append((auction, interval_start, interval_end))
 
-        return self._parse_fcm_reconfiguration_dataframe(annotated_auctions)
+        return self._parse_fcm_reconfiguration_dataframe(
+            annotated_auctions,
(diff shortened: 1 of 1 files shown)


gridstatus/isone_api/isone_api.py
```diff
@@ -2029,12 +2029,22 @@ def _parse_fcm_reconfiguration_dataframe(
                 df[column] = pd.to_numeric(df[column], errors="coerce")
 
         if "ARA" in df.columns:
-            df["ARA"] = pd.to_numeric(df["ARA"], errors="coerce").astype("Int64")
+            if df["ARA"].notna().any() and (df["ARA"] != "").any():
```
WillKoehrsen: Does only the annual data have the ARA column? Would it be cleaner to add an `auction_type` parameter to the function and only add the ARA column if that value is "annual" instead of adding the column to the monthly data and then having to drop it later?

Kladar: okay I do that

WillKoehrsen: Thanks. I'm not sure I was right that it would be cleaner though.

PR #763, opened 2025-12-22
Support Python 3.13 / 3.14 

tylernisonoff:
We realized we couldn't use the latest version of the library as it explicitly disallows python 3.13. Opening this to see if we can have a release that allows it!

-- Friends from Tyba


gridstatus/ercot.py
WillKoehrsen, line 6: These changes have already been merged right? I think we just need to merge main into this branch so these changes don't show up in the diff.

pyproject.toml
WillKoehrsen, line 1: Can we add `"rpds-py >= 0.30.0; python_version >= '3.14'"` to the dependencies? This was the only change I had to make for Python 3.14.

gridstatus/ercot.py
tylernisonoff: Sorry about that! messed up my fork

WillKoehrsen:
@Kladar @kmax12 Have you been able to verify my findings from https://github.com/gridstatus/gridstatus/pull/763#issuecomment-3683095352? I would like at least one confirmation.

WillKoehrsen (approved):
👍🏽

PR #764, opened 2025-12-23
[ENG-3022] Add SCED2 Offer Curve to ERCOT SCED Gen Resource

WillKoehrsen:
## Summary

- Adds "SCED1 Offer Curve" column to `Ercot().get_60_day_sced_disclosure` and `ErcotAPI().get_60_day_sced_disclosure`
- Fixes `ErcotAPI` 60 day methods that were broken as the result of a return type change from `get_historical_data`

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -1465,7 +1465,7 @@ class ErcotAPI:
         df_list = []
 
         # Get data once since both endpoints return the same zipfile
-        data_bytes_and_file_names = self.get_historical_data(
+        data_bytes = self.get_historical_data(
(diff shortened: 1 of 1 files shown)


gridstatus/ercot_api/ercot_api.py
```diff
@@ -1715,7 +1716,8 @@ def get_historical_data(
             files = self._individually_download_documents(links=links, verbose=verbose)
 
         if not read_as_csv:
-            return files
+            # Only return the bytes (not the filenames)
+            return [f[0] for f in files]
```
WillKoehrsen, line 1720: The `get_60_day_sced_disclosure` and `get_60_day_dam_disclosure` methods on `ErcotAPI` were broken and this change gets them working again. Those are the only methods that call `get_historical_data` with `read_as_csv=False` so were the only methods impacted.

PR #766, opened 2026-01-02
Fix ERCOT COP Adjustment for DST End (2025)

WillKoehrsen:
## Summary

- Fixes ERCOT COP Adjustment 60 day for DST End 2025
- Make sure `Ercot().get_cop_adjustment_period_snapshot_60_day('2025-11-02')` runs

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -4713,6 +4713,17 @@ class Ercot(ISOBase):
                     ambiguous=ambiguous,
                 ) - pd.Timedelta(hours=1)
             except pytz.AmbiguousTimeError as e:
+                # Handle datasets where HourEnding is like 01:00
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -4713,6 +4713,17 @@ def parse_doc(
                     ambiguous=ambiguous,
                 ) - pd.Timedelta(hours=1)
             except pytz.AmbiguousTimeError as e:
+                # Handle datasets where HourEnding is like 01:00
```
kmax12, line 4716: Took me a second to understand how just this diff this solved problem,  I think a comment like this would have help me get it a lot faster

```
# ERCOT inconsistently formats HourEnding as either "XX:00" (e.g., "01:00")
# or as an integer (e.g., 1). When in string format, normalize to integer
```

PR #767, opened 2026-01-04
fix(ENG 3044): IESO Load Year Transition

Kladar:
## Summary
Running into an issue where the year change (2025 -> 2026) causing an issue with how we retrieve data, since the year is used in the filename, making it difficult to get data across the year transition. Changes the frequency to `YEAR_START`

```
ieso = IESO()
df = ieso.get_load_zonal_5_min("2025-12-29", end="2026-01-02")
print(df)

df = ieso.get_load_zonal_hourly("2025-12-29", end="2026-01-02")
print(df)
```

### Details
Started running into a `FutureWarning` for pandas 3.0 on the chain assignment. We do this kind of columnar chain assignment basically everywhere so would be a much larger undertaking to actually be 3.0 ready, but figure I would try it out with the recommended `assign`.

### DIFF
--- a/gridstatus/ieso.py
+++ b/gridstatus/ieso.py
@@ -3651,9 +3651,9 @@ class IESO(ISOBase):
     ) -> pd.DataFrame:
         if date == "latest":
             url = f"{PUBLIC_REPORTS_URL_PREFIX}/RealtimeDemandZonal/PUB_RealtimeDemandZonal.csv"
-            return self._parse_load_zonal_data(url, date, end)
+        else:
+            url = f"{PUBLIC_REPORTS_URL_PREFIX}/RealtimeDemandZonal/PUB_RealtimeDemandZonal_{date.year}.csv"
 
-        url = f"{PUBLIC_REPORTS_URL_PREFIX}/RealtimeDemandZonal/PUB_RealtimeDemandZonal_{date.year}.csv"
         return self._parse_load_zonal_data(url, date, end)
 
     @support_date_range(frequency="YEAR_START")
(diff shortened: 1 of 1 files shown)


gridstatus/ieso.py
```diff
@@ -3678,17 +3679,27 @@ def _parse_load_zonal_data(
...
                 + pd.to_timedelta(df["Hour"] - 1, unit="h")
                 + pd.to_timedelta((df["Interval"] - 1) * 5, unit="m")
             ).dt.tz_localize(self.default_timezone)
-            df["Interval End"] = df["Interval Start"] + pd.Timedelta(minutes=5)
+            df = df.assign(
+                **{
+                    "Interval Start": interval_start,
+                    "Interval End": interval_start + pd.Timedelta(minutes=5),
+                },
+            )
```
Kladar, line 3692: Here's the pandas `FutureWarning`:

```
FutureWarning: ChainedAssignmentError: behaviour will change in pandas 3.0!
You are setting values through chained assignment. Currently this works in certain cases, but when using Copy-on-Write (which will become the default behaviour in pandas 3.0) this will never work to update the original DataFrame or Series, because the intermediate object on which we are setting values will behave as a copy.
A typical example is when you are setting values in a column of a DataFrame, like:

df["col"][row_indexer] = value

Use `df.loc[row_indexer, "col"] = values` instead, to perform the assignment in a single step and ensure this keeps updating the original `df`.

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy

```
good discussion point in State of the Python

WillKoehrsen: Good fix. I've tried to address warnings like this as they come up.

PR #769, opened 2026-01-08
[ENG-3079] Remove 'latest' Option from NYISO Where Acceptable and Fix NYISO Tests

WillKoehrsen:
## Summary

- Removes "latest" option from NYISO methods except for methods where there is a dedicated latest file
  - `NYISO().get_lmp()` and `NYISO().get_limiting_constraints_real_time()` have dedicated latest files and continue to work
- Removes tests that reference now non-existent "latest" support
- Fixes other tests to get the NYISO build passing?

### Details

- Run all the NYISO tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_nyiso.py -vvv`


gridstatus/nyiso.py
Shiggiddie, line 1: I don't know how to evaluate if the "latest" was removed properly for these various methods.

On this branch w/in `nyiso.py` I still see `"latest"` support for:
* get_status
* get_fuel_mix
* get_load
* get_btm_solar
* get_btm_solar_forecast
* get_interface_limits_and_flows_5_min
* get_lake_erie_circulation_real_time
* get_lake_erie_circulation_day_ahead
* get_lmp // Based on your description, this should be OK
* get_as_prices_day_ahead_hourly
* get_as_prices_real_time_5_min
* get_limiting_constraints_real_time // Based on your description, this should be OK
* get_limiting_constraints_day_ahead

gridstatus/tests/source_specific/test_nyiso.py
```diff
@@ -491,6 +495,10 @@ def test_get_loads(self):
 
     """get_status"""
 
+    @pytest.mark.skip("Not Supported")
+    def test_get_status_latest(self):
+        super().test_get_status_latest()
+
```
Shiggiddie, line 501: I'm not sure why this is skipped vs removed?

gridstatus/nyiso.py
WillKoehrsen: When you say "see "latest" support" what exactly do you mean? If you call any of these methods with "latest" except for `get_lmp` and `get_limiting_constraints_real_time`, you get an error.

gridstatus/tests/source_specific/test_nyiso.py
WillKoehrsen: This test and several others is defined on `BaseTestISO`. `TestNYISO` inherits from `BaseTestISO` so we have to define it here and skip it otherwise it will run and fail due to lack of support in `NYISO`.

gridstatus/nyiso.py
Shiggiddie: This was my bad...somehow I must have messed up the `git checkout...` because when I double checked I was on the `remove-latest-nyiso` I see the problems I was seeing were addressed already, I had to have been on the wrong branch locally 🤦

gridstatus/tests/source_specific/test_nyiso.py
Shiggiddie: This makes sense!  Thank you for the explanation!

Kladar (approved):
Very nice, the decluttering continues! 

Starting now, I will also refrain from adding latest support by default in new methods as well.

PR #777, opened 2026-01-26
Add Timeout for ERCOT API Calls

WillKoehrsen:
## Summary

- Adds a timeout to calls to the ERCOT API so these calls don't hang indefinitely

### DIFF
--- a/gridstatus/ercot_api/ercot_api.py
+++ b/gridstatus/ercot_api/ercot_api.py
@@ -306,6 +306,7 @@ class ErcotAPI:
         # Connect timeouts raise immediately.
         retries = 0
         delay = self.initial_delay
+        reason = ""
         while retries <= self.max_retries:
             try:
                 if method == "POST":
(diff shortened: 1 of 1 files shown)


gridstatus/ercot_api/ercot_api.py
```diff
@@ -301,12 +308,14 @@ def make_api_call(
                     url,
                     headers=self.headers(api=api),
                     json=api_params,
+                    timeout=REQUEST_TIMEOUT_SECONDS,
```
WillKoehrsen: I'm not sure if we should just raise on a timeout (current implementation) or catch the exception and try again up to `max_retries`. The choice depends on if the API will succeed on a retry or if it will time out again.

Shiggiddie: `requests` library allows for a [tuple](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts) to be set for the `timeout` kward for connect vs read timeouts.

Furthermore, requests [defines](https://github.com/psf/requests/blob/70298332899f25826e35e42f8d83425124f755a5/src/requests/exceptions.py#L80-L88) a `ConnectTimeout` vs `ReadTimeout` error.

I think we should:
1. Define a connect, read timeout tuple
2. If the error is due to a `ConnectTimeout`, do not retry and throw an error
3. If the error is due to a `ReadTimeout`, retry up to max retries

WillKoehrsen: Updated to use separate read and connect timeouts. On a read timeout, we retry up to `max_retries`. A connect timeout raises immediately.

gridstatus/ercot_api/ercot_api.py
```diff
@@ -293,21 +302,38 @@ def make_api_call(
         )
 
         # make request with exponential backoff retry strategy
+        # Retries on rate limiting (HTTP 429) and read timeouts.
+        # Connect timeouts raise immediately.
         retries = 0
         delay = self.initial_delay
         while retries <= self.max_retries:
```
Shiggiddie, line 310: I think we can make this code more maintainable by having a single place to sleep & increase exponential backoff, e.g.

```
while retries <= self.max_retries:
    try:
        if method == "POST":
            response = requests.post(
                url,
                headers=self.headers(api=api),
                json=api_params,
                timeout=REQUEST_TIMEOUT,
            )
        else:
            response = requests.get(
                url,
                headers=self.headers(api=api),
                params=api_params,
                timeout=REQUEST_TIMEOUT,
            )

        # If request succeeds, check response
        if response.status_code == status_codes.codes.OK:
            break  # Success
        elif response.status_code == status_codes.codes.TOO_MANY_REQUESTS:
            # We'll retry below
            reason = "Rate-limited"
        else:
            # Non-retriable error
            response.raise_for_status() # This may not be useful to us?

    except requests.exceptions.ReadTimeout:
        reason = "Read timeout"

    retries += 1
    if retries > self.max_retries:
        raise RuntimeError(
            f"Error: Failed after {self.max_retries} retries for {url} with params {api_params}"
        )

    # Centralized backoff
    logger.warning(
        f"Warn: {reason}: waiting {delay:.1f} seconds before retry {retries}/{self.max_retries} "
        f"requesting url: {url} with params: {api_params}"
    )
    time.sleep(delay + random.uniform(0, delay * 0.1))
    delay *= 2
```

WillKoehrsen: Good idea, updated.

PR #778, opened 2026-01-28
[ENG-3152] MISO Multiday Operating Margin Methods

WillKoehrsen:
## Summary

- Adds `MISO().get_multiday_operating_margin` which retrieves data for MISO as a whole
- Adds `MISO().get_multiday_operating_margin_regional` which retrieves data for each region in one dataset

### Validation

- Run specific tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_miso.py -k multiday`

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -1980,16 +1980,27 @@ class MISO(ISOBase):
                     else:
                         row_data[new_name] = None
 
-            # Handle "Additional Emergency Headroom" which appears 3 times
-            # We identify them by finding all rows and using their position
+            # Handle "Additional Emergency Headroom" which appears 3 times:
+            # 1. After RESOURCE COMMITTED -> "Committed Additional Emergency Headroom"
+            # 2. After RESOURCE UNCOMMITTED -> "Uncommitted Additional Emergency Headroom"
+            # 3. After EMERGENCY RESOURCES -> "Emergency Resources Additional Headroom"
             emergency_headroom_rows = data[
                 data.iloc[:, 0].astype(str).str.strip()
                 == "Additional Emergency Headroom"
             ]
             if len(emergency_headroom_rows) >= 1:
-                # First one: after RESOURCE COMMITTED
                 val = emergency_headroom_rows.iloc[0][col]
-                row_data["Additional Emergency Headroom"] = (
+                row_data["Committed Additional Emergency Headroom"] = (
+                    float(val) if pd.notna(val) else None
+                )
+            if len(emergency_headroom_rows) >= 2:
+                val = emergency_headroom_rows.iloc[1][col]
+                row_data["Uncommitted Additional Emergency Headroom"] = (
+                    float(val) if pd.notna(val) else None
+                )
+            if len(emergency_headroom_rows) >= 3:
+                val = emergency_headroom_rows.iloc[2][col]
+                row_data["Emergency Resources Additional Headroom"] = (
                     float(val) if pd.notna(val) else None
                 )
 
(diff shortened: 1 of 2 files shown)


gridstatus/miso.py
```diff
@@ -1748,3 +1748,330 @@ def get_interchange_5_min(
...
+                    float(val) if pd.notna(val) else None
+                )
+            if len(emergency_headroom_rows) >= 2:
+                # Second one: after RESOURCE UNCOMMITTED
+                # Note: The spec lists this as "Additional Emergency Headroom" 3 times
+                # We'll just use the first one as per the spec
+                pass
+            if len(emergency_headroom_rows) >= 3:
+                # Third one: after EMERGENCY RESOURCES
+                pass
```
Shiggiddie: I don't understand why these are here?  It doesn't appear to be doing anything?

gridstatus/miso.py
```diff
@@ -1748,3 +1748,330 @@ def get_interchange_5_min(
...
+            raise NoDataFoundException(
+                f"No data found for any region on {date}",
+            )
+
+        return pd.concat(dfs, ignore_index=True)
+
+    def _get_multiday_operating_margin_data(
+        self,
+        date: str | pd.Timestamp,
+        region: str = "MISO",
```
Shiggiddie, line 1867: Not sure why this defaults to "MISO", can we type off of `regions = ["NORTH", "CENTRAL", "NORTH+CENTRAL", "SOUTH"]`, above?

gridstatus/miso.py
```diff
@@ -1748,3 +1748,330 @@ def get_interchange_5_min(
...
+                col
+                for col in result_df.columns
+                if col not in ["Publish Date", "Peak Hour", "Region"]
+            ]
+            for col in numeric_cols:
+                if col in result_df.columns:
+                    result_df[col] = pd.to_numeric(result_df[col], errors="coerce")
+
+            # Add calculated columns for regional (if not already present)
+            if region != "MISO":
```
Shiggiddie, line 2026: Should always happen? `regions = ["NORTH", "CENTRAL", "NORTH+CENTRAL", "SOUTH"]`

gridstatus/miso.py
```diff
@@ -1748,3 +1748,330 @@ def get_interchange_5_min(
...
+
+                # Max Possible RDT - only add if not already present
+                if "Max Possible RDT" not in result_df.columns:
+                    result_df["Max Possible RDT"] = pd.Series(
+                        dtype="float64",
+                        index=result_df.index,
+                    )
+
+            # Reorder columns
+            if region == "MISO":
```
Shiggiddie, line 2059: Should never happen? `regions = ["NORTH", "CENTRAL", "NORTH+CENTRAL", "SOUTH"]`

gridstatus/miso.py
```diff
@@ -1748,3 +1748,330 @@ def get_interchange_5_min(
...
+            # Read the actual data
+            data = pd.read_excel(
+                excel_file,
+                sheet_name=region,
+                skiprows=3,
+            )
+
+        # Find the date row - it's the first row with date strings in columns
+        date_row_idx = None
+        for idx in range(min(3, len(data))):
```
Shiggiddie: nit `3` here is a bit magical, may be nice to extract to `DEFAULT_CANDIDATE_DATE_ROWS` or something like that

gridstatus/miso.py
WillKoehrsen: No, this can be "MISO" and is called with `region="MISO"` above.

gridstatus/miso.py
WillKoehrsen: Ahh yeah, that was a leftover artifact. It's been removed (along with the block above).

gridstatus/miso.py
WillKoehrsen: Answered above. `region` can be "MISO".

gridstatus/miso.py
WillKoehrsen: Yeah you're right. I've added a comment and a helpful constant name.

gridstatus/miso.py
Shiggiddie: Ah!  Thank you, I misread the other call site.  Ignore

PR #784, opened 2026-02-05
[ENG-3206] [ENG-3205] ERCOT DAM 60 Day ESR Datasets

WillKoehrsen:
## Summary

- Adds ESR and ESR as offers to ERCOT DAM 60 Day data
- Each of these datasets is on one hour intervals

### Validation

```python
from gridstatus.ercot import Ercot
dam = Ercot().get_60_day_dam_disclosure('2025-12-06', process=True)
esr_as_offers = dam['dam_esr_as_offers']
esr_as_offers.head()

esr = dam['dam_esr']
esr.head()
```

- Run tests with
  - `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot.py -k 60_day_dam_disclosure`

### DIFF
--- a/gridstatus/ercot_60d_utils.py
+++ b/gridstatus/ercot_60d_utils.py
@@ -331,6 +331,9 @@ SCED_ESR_COLUMNS = [
     "AS Capability RegDown",
     "AS Capability ECRS",
     "AS Capability NonSpin",
+    "SOC",
+    "Min SOC",
+    "Max SOC",
     "AS Awards NonSpin",
     "AS Awards RRSFFR",
     "AS Awards RRSPFR",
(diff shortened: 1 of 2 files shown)


gridstatus/ercot_60d_utils.py
```diff
@@ -180,6 +182,41 @@
...
+    "RRS MCPC",
+    "ECRSSD Awarded",
+    "ECRS MCPC",
+    "NonSpin Awarded",
+    "NonSpin MCPC",
+    "QSE submitted Curve",
+]
+
+# Same columns as gen/load resource AS offers
+DAM_ESR_AS_OFFERS_COLUMNS = DAM_RESOURCE_AS_OFFERS_COLUMNS
```
Shiggiddie: I'd be a little wary if `DAM_RESOURCE_AS_OFFERS_COLUMNS` change but `DAM_ESR_AS_OFFERS_COLUMNS` don't and we inadvertently break DAM ESR as a result.

Is there high confidence these two will evolve together?

WillKoehrsen: If the columns in one dataset change, we will update this.

PR #787, opened 2026-02-06
Remove Interval Start and Interval End from ERCOT SCED 60 Day Datasets

WillKoehrsen:
## Summary

- Removes "Interval Start" and "Interval End" from ERCOT SCED Gen and Load Resource 60 day
- We are removing these columns because these are SCED datasets and we can rely on the SCED Timestamp column

### Validation

- Run tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot_api.py -k 60_day_sced_disclosure`

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2171,29 +2171,25 @@ class Ercot(ISOBase):
 
             return df
 
-        load_resource = load_resource.rename(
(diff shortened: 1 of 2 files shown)


WillKoehrsen:
Merging this without approval because we told customers we'd update these datasets at 9 am CST and our downstream code needs this update.

PR #788, opened 2026-02-06
[ENG-3201] [ENG-3202] ERCOT SCED Resource AS Offers and AS Offer Updates

WillKoehrsen:
## Summary

- Adds SCED Resource AS Offers and SCED AS Offer Updates in Op Hour to the 60 day SCED disclosures

### Details

- Run tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot_api.py -k 60_day_sced_disclosure`

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2120,6 +2120,7 @@ class Ercot(ISOBase):
         verbose: bool = False,
         skip_esr: bool = False,
     ) -> dict:
+        # TODO: there are other files in the zip folder
         load_resource_file = None
         gen_resource_file = None
         smne_file = None


gridstatus/ercot.py
```diff
@@ -2112,11 +2116,12 @@ def _handle_60_day_sced_disclosure(
         verbose: bool = False,
         skip_esr: bool = False,
     ) -> dict:
-        # TODO: there are other files in the zip folder
```
Shiggiddie, line 2119: Have we now covered all of the files, or was this just a useless TODO?

gridstatus/ercot.py
```diff
@@ -2139,6 +2148,16 @@ def _handle_60_day_sced_disclosure(
         esr = None
         if not skip_esr and esr_file:
             esr = pd.read_csv(z.open(esr_file))
```
Shiggiddie, line 2155: nit: could we stay consistent with a conditional expression style vs standard iff statement with assignment

gridstatus/ercot.py
WillKoehrsen: There are 10 documents in the SCED disclosure file and we have 6 SCED 60 day datasets. I will restore this TODO.

gridstatus/ercot.py
WillKoehrsen: Like this? 

```python
esr = pd.read_csv(z.open(esr_file)) if not skip_esr and esr_file else None
```

I'm not sure that's more readable.

gridstatus/ercot.py
WillKoehrsen: Restored.

gridstatus/ercot.py
Shiggiddie: Seeing that written up I agree it is more readable as implemented and worth keeping as-is despite the switching styles overhead (which is admittedly minimal).

PR #791, opened 2026-02-09
[ENG-3247] Bypass Exponential Retry when Running Tests

WillKoehrsen:
## Summary

- Allows tests to run without the delay from exponential backoff by mocking `time.sleep` to be a non-op in places where exponential backoff is applied

### DIFF
--- a/gridstatus/tests/conftest.py
+++ b/gridstatus/tests/conftest.py
@@ -1,0 +1,31 @@
+from unittest.mock import patch
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def disable_exponential_backoff_sleep():
+    """Disable time.sleep in all modules that use exponential backoff.
+
+    This prevents tests from waiting during retry logic, making them run faster.
+    The fixture is autouse=True so it applies to all tests automatically.
+    """
+    modules_with_backoff = [
+        "gridstatus.pjm",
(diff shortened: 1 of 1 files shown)


gridstatus/tests/conftest.py
Shiggiddie, line 1: This approach relies on us to maintain `modules_with_backoff`.

Are we able to get away with a global sleep mock e.g.
```
@pytest.fixture(autouse=True)
def disable_sleep(monkeypatch):
    monkeypatch.setattr("time.sleep", lambda *_: None)
```

My sense is that we ought not have any tests that rely on real sleep behavior, which is why the patch is acceptable.


Alternatively, if we built out a specific utility for exponential backoff, we could patch only that utility for our test suite and avoid `modules_with_backoff` as well.

WillKoehrsen: I'd prefer to avoid a blanket patching of `time.sleep` because of unintended side effects.

Shiggiddie: That's fair, how about a specific utility for exponential backoff that we patch?

My primary concern is not needing to maintain `modules_with_backoff`

WillKoehrsen: The maintenance overhead seems pretty low and if we don't include a module the impact is minor.

PR #797, opened 2026-02-23
[ENG-3342] Added Columns for ERCOT SCED 60 Day Datasets

WillKoehrsen:
## Summary

- Adds the new columns ERCOT has added to SCED 60 Day Disclosure datasets
  - SCED Load Resource
    - AS Capability RRSPF
    - AS Capability RRSFF
    - AS Capability RRSUF
  - SCED Gen Resource
    - AS Capability RRSPF
    - AS Capability RRSFF
  - SCED ESR
    - AS Capability RRSPF
    - AS Capability RRSFF
- These columns are available starting with data published 2026-02-19 which corresponds to data on 2025-12-21

### Validation

- Run specific tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot.py -k 60_day_sced_disclosure`

PR #799, opened 2026-02-25
[ENG-3359] Use Supplemental Correction Data for ERCOT SCED 60 Day Datasets

WillKoehrsen:
## Summary

- Updates the ERCOT SCED 60 Day datasets to use the supplemental correction ERCOT published for
  - SCED Gen Resource
  - SCED Load Resource
  - SCED ESR
- ERCOT published these corrections because they previously published these datasets missing some columns (see https://github.com/gridstatus/gridstatus/pull/797)

### Validation

- Run tests with `VCR_RECORD_MODE=all uv run pytest -s -vvv gridstatus/tests/source_specific/test_ercot.py -k "test_get_60_day_sced_disclosure"`

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -2059,6 +2059,14 @@ class Ercot(ISOBase):
         pd.Timestamp("2025-12-06").date(),
     ]
 
+    # Data dates for which ESR, Gen Resource, and Load Resource data from the
+    # normal disclosure file is incorrect and should be replaced with data from
+    # the supplemental correction files. Published on Feb 23, 2026.
+    # Data dates Dec 5, 2025 through Dec 20, 2025 correspond to report dates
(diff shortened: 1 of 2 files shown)


gridstatus/ercot.py
```diff
@@ -2359,6 +2384,107 @@ def _get_esr_correction_data(
...
+
+        Arguments:
+            date: The data date (not report date) to fetch supplemental data for
+            process: If True, process the data into standardized format
+            verbose: If True, print verbose output
+
+        Returns:
+            dict with keys for the corrected datasets
+        """
+        # Files in the supplemental zips are named with the report date
```
Shiggiddie, line 2431: Worth a guard clause to ensure date is in range?

WillKoehrsen: We only call this internal function within the guard clause `if use_sced_supplemental:` so I think it's okay.

PR #802, opened 2026-03-01
[ENG-3372] Telemetered Net Output Fix

WillKoehrsen:
## Summary

- Fix the "Telemetered Net Output" column in ERCOT SCED gen 60 day data
- At some point, the column name stopped having an extra space in it. We now strip the columns of spaces before processing

### DIFF
--- a/gridstatus/tests/source_specific/test_ercot.py
+++ b/gridstatus/tests/source_specific/test_ercot.py
@@ -927,6 +927,38 @@ class TestErcot(BaseTestISO):
         smne = df_dict[SCED_SMNE_KEY]
         assert len(smne) > 0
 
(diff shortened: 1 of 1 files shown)


kmax12 (approved):

gridstatus/ercot_60d_utils.py
```diff
@@ -989,7 +991,7 @@ def process_sced_gen(df):
         "LASL",
         "LDL",
         "Base Point",
-        "Telemetered Net Output ",
```
WillKoehrsen, line 992: The trailing space was removed starting 2025-12-05 (the start of ERCOT RTC+B).

PR #803, opened 2026-03-02
[ENG-3385] MISO Binding Constraints Intraday Fixes

WillKoehrsen:
## Summary

- When there are no active constraints, MISO publishes placeholder constraints where all the strings are "None" and the values are 0. These are not real constraints so we should raise a NoDataFoundException in these cases.
- The "Period" provided in the data should be "Interval End", not "Interval Start" as currently implemented
- Correct data types for values other than "Shadow Price" which is a float. Other numeric values are ints

### Validation

- Run specific tests with `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_miso.py -k binding_constraints_real_time_intraday`

### DIFF
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -1626,8 +1626,10 @@ class MISO(ISOBase):
             },
         )
 
-        for col in ["Shadow Price", "Override", "BP1", "PC1", "BP2", "PC2"]:
-            df[col] = pd.to_numeric(df[col], errors="coerce")
+        # Shadow price is a float, all others are integers
+        df["Shadow Price"] = pd.to_numeric(df["Shadow Price"], errors="coerce")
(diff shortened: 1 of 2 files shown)


gridstatus/miso.py
```diff
@@ -1608,7 +1608,12 @@ def _parse_binding_constraints_real_time_intraday(
 
         df = pd.DataFrame(constraints)
 
-        df["Interval Start"] = pd.to_datetime(df["Period"]).dt.tz_localize(
+        # MISO publishes empty binding constraints when there are no active constraints
+        if (df["Name"] == "None").all():
+            raise NoDataFoundException("No real-time binding constraints data found")
+
+        # Period is Interval End
```
WillKoehrsen, line 1615: Updated to change Period to Interval End.

Kladar (approved):

PR #810, opened 2026-03-05
Updated ercot.py to track more projects in queue

ahmedismailxyz:
Expand ERCOT Interconnection Queue to include all project sheets (large gen, small gen, withdrawn, canceled)

## Summary
This PR enhances the ERCOT get_interconnection_queue() method to parse all four sheets from the GIS Report Excel file, not just the "Project Details - Large Gen" sheet. The newly included sheets are:

- Project Details - Small Gen: smaller generation projects that were previously ignored
- Inactive Projects: projects that became inactive before completion
- Cancellation Update: projects that were explicitly cancelled

This is particularly useful for users who want to collect comprehensive ERCOT interconnection data and review withdrawn projects and points of failure in the queue process. Previously, only large active/completed generation projects were returned, which gave an incomplete picture of the interconnection landscape. This allows users to start collecting the data for more projects.

### Details
**What Changed**

**Refactored into modular parsing methods:**
- Extracted the large-gen parsing logic from get_interconnection_queue() into a dedicated _parse_large_gen() method.
- Added _parse_small_gen() to handle the "Project Details - Small Gen" sheet, which has a different structure (different skiprows, uses "Model Ready Date" as the queue date instead of "Screening Study Started").
- Added _parse_inactive_projects() to parse the "Inactive Projects" sheet. These are marked with Status = "Withdrawn" and Withdrawal Comment = "Inactive", with the inactive date captured as the Withdrawn Date.
- Added _parse_cancelled_projects() to parse the "Cancellation Update" sheet. These are marked with Status = "Withdrawn" and Withdrawal Comment = "Cancelled", with the cancel date captured as the Withdrawn Date.

**Promoted lookup maps to class-level constants:**
- fuel_type_map and technology_type_map were moved from local variables inside the method to class-level attributes (_fuel_type_map, _technology_type_map). This avoids duplication since both large-gen and small-gen parsing need them.

**Added a Size Category column:**
- Each row now includes a "Size Category" field ("Large" or "Small") so users can distinguish between large and small generation projects in the combined output.

**Smarter handling of missing columns:**
  - Inactive and cancelled sheets don't have all the same columns as the active project sheets. The parsers add placeholder None columns for missing fields so all DataFrames can be concatenated cleanly.
  - The missing columns list (for format_interconnection_df) now filters out columns that already exist in the data — this prevents overwriting real values like Withdrawn Date and Withdrawal Comment that come from the inactive/cancelled sheets.

**Opened the Excel file once:**
  - The raw data is now loaded into a pd.ExcelFile object once and passed to each parser, rather than re-reading the file for each sheet.

**Cleaned up TODOs and dead comments:**
  - Removed outdated TODO comments (e.g., "other sheets for small projects, inactive, and cancelled" — now implemented) and commented-out column references (Construction Start, Construction End) that were always null.

**Why This Matters**
The ERCOT interconnection queue is a critical dataset for understanding the pipeline of generation projects connecting to the Texas grid. By including inactive and cancelled projects, analysts can now:

- Track attrition rates — understand what fraction of proposed projects drop out and at what stage
- Identify failure patterns — correlate withdrawals with project size, fuel type, technology, or geography
- Get a complete historical picture — see all projects that have ever entered the queue, not just those currently active or completed
- Analyze small vs. large project dynamics — the new Size Category field enables segmentation that wasn't previously possible

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -1494,62 +1494,51 @@ class Ercot(ISOBase):
         response = requests.get(doc_info.url)
         return utils.get_response_blob(response)
 
-    def get_interconnection_queue(self, verbose: bool = False) -> pd.DataFrame:
-        """
-        Get interconnection queue for ERCOT
+    _fuel_type_map = {
+        "BIO": "Biomass",
+        "COA": "Coal",
+        "GAS": "Gas",
+        "GEO": "Geothermal",
+        "HYD": "Hydrogen",
+        "NUC": "Nuclear",
+        "OIL": "Fuel Oil",
+        "OTH": "Other",
+        "PET": "Petcoke",
+        "SOL": "Solar",
+        "WAT": "Water",
+        "WIN": "Wind",
+    }
 
-        Monthly historical data available here:
-            http://mis.ercot.com/misapp/GetReports.do?reportTypeId=15933&reportTitle=GIS%20Report&showHTMLView=&mimicKey
-        """  # noqa
-        raw_data = self.get_raw_interconnection_queue(verbose)
-        # TODO other sheets for small projects, inactive, and cancelled project
-        # TODO see if this data matches up with summaries in excel file
-        # TODO historical data available as well
+    _technology_type_map = {
+        "BA": "Battery Energy Storage",
+        "CC": "Combined-Cycle",
+        "CE": "Compressed Air Energy Storage",
+        "CP": "Concentrated Solar Power",
+        "EN": "Energy Storage",
+        "FC": "Fuel Cell",
+        "GT": "Combustion (gas) Turbine, but not part of a Combined-Cycle",
+        "HY": "Hydroelectric Turbine",
+        "IC": "Internal Combustion Engine, eg. Reciprocating",
+        "OT": "Other",
+        "PV": "Photovoltaic Solar",
+        "ST": "Steam Turbine other than Combined-Cycle",
+        "WT": "Wind Turbine",
+    }
 
-        # skip rows and handle header
+    def _parse_large_gen(self, excel_file: pd.ExcelFile) -> pd.DataFrame:
+        """Parse 'Project Details - Large Gen' sheet."""
         queue = pd.read_excel(
-            raw_data,
+            excel_file,
             sheet_name="Project Details - Large Gen",
             skiprows=30,
         ).iloc[4:]
 
         queue["State"] = "Texas"
         queue["Queue Date"] = queue["Screening Study Started"]
+        queue["Size Category"] = "Large"
 
-        fuel_type_map = {
-            "BIO": "Biomass",
-            "COA": "Coal",
-            "GAS": "Gas",
-            "GEO": "Geothermal",
-            "HYD": "Hydrogen",
-            "NUC": "Nuclear",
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -1565,6 +1587,144 @@ def get_interconnection_queue(self, verbose: bool = False) -> pd.DataFrame:
...
+
+    def _parse_cancelled_projects(self, excel_file: pd.ExcelFile) -> pd.DataFrame:
+        """Parse 'Cancellation Update' sheet."""
+        df = pd.read_excel(
+            excel_file,
+            sheet_name="Cancellation Update",
+            skiprows=7,
+        )
+        # Drop footer notes by requiring non-null Size Category (data rows)
+        df = df.dropna(subset=["INR", "Size Category"])
```
Kladar, line 1691: Thanks for the contribution! 

It seems like projects can show up with the same INR in multiple sheets (e.g. in both Large Gen and Cancelled). Would it be worth de-duplicating these projects to just have one entry returned for them, depending on where they are?

gridstatus/ercot.py
```diff
@@ -1565,6 +1587,144 @@ def get_interconnection_queue(self, verbose: bool = False) -> pd.DataFrame:
...
+                df[col] = None
+
+        return df
+
+    def _parse_cancelled_projects(self, excel_file: pd.ExcelFile) -> pd.DataFrame:
+        """Parse 'Cancellation Update' sheet."""
+        df = pd.read_excel(
+            excel_file,
+            sheet_name="Cancellation Update",
+            skiprows=7,
```
Kladar: I could be wrong here, let me know if your experience is different, but I think this `skiprows` value varies depending on the historical queue being grabbed. They aren't totally consistent in how many rows there are in the header from month to month. Not sure if you want to handle that more dynamically such that historical queues can be grabbed or avoid brittleness to format changes in future queues.

gridstatus/ercot.py
ahmedismailxyz: Got it, what do you think of merging rather than dropping projects? I was thinking could be better so we can keep the cancelled status but preserve the detailed columns from the active sheet because the cancelled ones don't have much information.

gridstatus/ercot.py
ahmedismailxyz: I think you are right about this. I'm going to look into auto-detecting the header row by scanning for known column names.

gridstatus/ercot.py
ahmedismailxyz: I just checked the last 6 months of GIS reports of ERCOT and it seems that the overlaps only exist between active sheets (large/small gen) and the commission update. Furthermore, I used the code to get the data as well to validate. I think the duplication risk would come from adding the commissioning update sheet in the future. However, the code doesn't parse the commissioning update as of now though.

Kladar: got it, thanks for checking! makes sense to me

gridstatus/ercot.py
ahmedismailxyz: I updated it and made a new commit based on your feedback, let me know your thoughts

Kladar (approved):
Looks good, sorry for the delay, thanks!

PR #812, opened 2026-03-06
Fix `timedelta64[ns] to timedelta64[h]` Conversion Error (Issue 227)

brianward92:
This PR fixes the error
```
  ValueError: Cannot convert from timedelta64[ns] to timedelta64[h]. Supported          
  resolutions are 's', 'ms', 'us', 'ns'
```
It should help along the way to supporting `pandas>=2.0`.


WillKoehrsen (approved):
Code looks good and fixes the issue. There is a merge conflict, but, once that is resolved, this should be ready to merge. Sorry for the wait on the review. Next time you can tag me directly and I'll get to it sooner.

PR #814, opened 2026-03-06
fix9ENG 3389): Date, not datetime, for publish date

Kladar:
## Summary
Sets this as just a publish date.

### DIFF
--- a/gridstatus/tests/source_specific/test_ercot.py
+++ b/gridstatus/tests/source_specific/test_ercot.py
@@ -1,3 +1,4 @@
+import datetime
 from io import StringIO
 from typing import Dict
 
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -5075,7 +5075,7 @@ def _get_settlement_points_csv(
                 filename_contains,
             )
             df = df.rename(columns=column_mapping)
-            df["Publish Date"] = publish_date.normalize()
+            df["Publish Date"] = publish_date.date()
```
WillKoehrsen, line 5078: Any tests we need to update with this change?

Kladar: They were all passing still so I didn't bother, but I've added some type checking to the existing tests now so we can make sure it's the right format

PR #816, opened 2026-03-08
Ercot daylight saving time fix

Shiggiddie:
## Summary

RTDMCPCNP6329_20260308_015503_csv includes a interval end of "2:00" am on the standard time to day light savings transition, leading to an error.

### Details

Include nonexistent to properly shift forward any times that don't exist in the daylight savings time transition

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -6331,12 +6331,15 @@ class Ercot(ISOBase):
         return self._handle_indicative_mcpc_rtd(df)
 
     def _handle_indicative_mcpc_rtd(self, df: pd.DataFrame) -> pd.DataFrame:
-        # Parse timestamps with DST handling
+        # Parse timestamps with DST handling. nonexistent= only applies to
+        # spring-forward (gap when 02:00 does not exist); fall-back repeated
+        # hour is handled by ambiguous= from RepeatedHourFlag.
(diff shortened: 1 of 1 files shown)


gridstatus/ercot.py
```diff
@@ -6331,12 +6331,15 @@ def get_indicative_mcpc_rtd(
...
-        # Parse timestamps with DST handling
+        # Parse timestamps with DST handling. nonexistent= only applies to
+        # spring-forward (gap when 02:00 does not exist); fall-back repeated
+        # hour is handled by ambiguous= from RepeatedHourFlag.
         df["Interval End"] = pd.to_datetime(df["IntervalEnding"]).dt.tz_localize(
             self.default_timezone,
             ambiguous=self.ambiguous_based_on_dstflag(
                 df.rename(columns={"IntervalEndingRepeatedHourFlag": "DSTFlag"}),
             ),
+            nonexistent="shift_forward",
```
WillKoehrsen, line 6342: I was wondering if when you shift forward the non-existent time you get duplicates. However, based on what I can see, that is not the case.

kmax12: we should leave that as a comment. 

i had the same question. if there aren't duplicates, im good with this approach.

PR #842, opened 2026-04-07
feat(ENG-995): S3-backed VCR fixtures with GHA cache for CI

Shiggiddie:
## Summary

Moves the gridstatus test suite off flaky live-API calls and onto S3-backed VCR cassette playback for per-PR CI, while preserving a full live-integration run in a new nightly workflow. Uploads ~190 new/refreshed cassettes across 10 ISOs, refactors tests to use absolute dates for cassette stability, and adds safety rails so missing cassettes, cache-buster URLs, and empty-response edge cases don't silently break CI.

## Infrastructure

- **`scripts/fixtures.py`** — new sync tool with `download`, `upload`, `cache-paths`, and `manifest` commands. `ISO_FIXTURE_MAP` is the single source of truth mapping CI matrix ISO names to fixture directories (some ISOs have multiple: `caiso` covers `caiso_save_to`, `ercot` covers `ercot_api`, etc.).
- **Public S3 bucket** `gridstatus-test-fixtures` holds every recorded cassette. Downloads use `--no-sign-request` (no AWS creds needed on CI or dev machines).
- **`.github/workflows/tests-pr.yaml`** runs per-ISO matrix jobs. Each job: restores GHA cache → falls back to S3 download on cache miss → verifies at least one fixture file is present → runs `make test-<iso>` with `VCR_RECORD_MODE=none` → asserts pytest wrote a junit XML with at least one passing test.
- **`.github/workflows/tests-nightly.yaml`** runs the full `test-cov` suite nightly with live integration tests included — this is where `@pytest.mark.integration` tests actually execute.
- **`scripts/fixtures.py::cmd_upload`** has a safety scan that blocks cassettes with `4xx/5xx` responses, with an `EXPECTED_ERROR_PATTERNS` allowlist for intentional error-path tests (`*_no_data*`, `*_raises_error*`, `*_too_far_in_past*`, etc.).
- **`gridstatus/tests/vcr_utils.py`** defaults `record_mode` to `"none"` on CI and `"new_episodes"` locally, and registers a generalized cache-buster matcher (strips `_=<digits>` query params) for all ISOs — needed for CAISO's `outlook/history` CSVs and ERCOT's public API.

## Local dev workflow

- `make fixtures-download` / `make fixtures-download-iso iso=<x>` pulls from S3.
- `make fixtures-upload iso=<x>` uploads after recording (requires AWS creds).
- Default local behavior (no `CI=true`) records new interactions into existing cassettes automatically via `VCR_RECORD_MODE=new_episodes`.
- `CI=true VCR_RECORD_MODE=none uv run pytest …` reproduces the PR-CI playback behavior exactly.

See `CONTRIBUTING.md` for the complete cassette workflow documentation.

## Test refactoring

- **~60 variable renames**: tests with misleadingly-named absolute-date vars (`yesterday`, `tomorrow`, `three_days_ago`, etc.) were renamed to `start`/`end`/`date` so the names no longer imply time-sensitivity.
- **~30 tests marked `@pytest.mark.integration`**: tests that inherently require relative dates, live-only API state, or rolling data windows. These run in nightly, not PR CI.
- **~45 tests marked `@pytest.mark.skip`** with Notion tech-debt links: tests with deeper issues (ERCOT RTC+B migration, library tz bugs, empty-archives handling, parallel-playback flakiness, etc.). See the [Tech Debt - gridstatus](https://www.notion.so/gridstatus/Tech-Debt-gridstatus-33de835f42aa8031b88fddadf1c5d3d6) Notion project.
- **Redundant `use_cassette` wrappers removed** from ~8 tests that only call `pytest.raises(...)` — they make no HTTP requests, so a cassette wrapper served no purpose and caused false missing-cassette skips in playback.
- **Makefile markers consolidated**: `UNIT_NOT_SLOW := -m "not integration and not slow"` (single `-m` to avoid the pytest "last `-m` wins" trap).

## Library / parser fixes discovered along the way

- **`gridstatus/caiso/daily_energy_storage.py`** — raise `NotSupported` for `"latest"`/`"today"` inputs instead of crashing inside `pd.to_datetime`.
- **`gridstatus/miso_api.py`** — drop the new `init` column before the float cast (MISO added it to medium-term load forecast responses).
- **`gridstatus/spp.py`** — accept both `#GMTTIME` and `GMTTIME` column names in the east/west interchange parsers (CSV format changed).

## Cassettes uploaded to S3

| ISO | Cassettes added |
|---|---:|
| ieso | 33 |
| ercot + ercot_api | 88 |
| pjm | 49 |
| aeso | 17 |
| caiso (incl. caiso_save_to) | 3 |
| misc adjustments | ~1 |
| **Total** | **~190** |

## Known residual issues (all tracked in Notion)

- ercot_api parallel playback is flaky under `-n 5` (clean under `-n 1`).
- Several ercot_api assertions drift with ERCOT data shape changes (DST row counts, SCED timing tolerance, archive response schema).
- pjm has 22 tests skipped pending date refresh within PJM's ~30-day retention window.
- 2 spp tests hit a pandas tz-naive/tz-aware comparison bug inside `SPP.get_interchange_real_time` that only surfaces in UTC-based CI.

## Test plan
- [x] PR CI matrix (all 10 test-iso jobs + test-base + lint + build docs + install) passes on this branch.
- [x] S3 bucket contents match what test files reference (verified via sweep on 2026-04-16/17).
- [x] Cassette safety-scan allowlist accepts `_no_data`/`_raises_error` patterns without `--force`.
- [x] Local record mode still works (`VCR_RECORD_MODE=new_episodes make test-<iso>`).
- [x] Cursor bot review addressed — 6 comments fixed in-branch, 5 already resolved by earlier commits, 6 documented as intentionally deferred.


.github/workflows/tests-commit.yaml
```diff
@@ -26,16 +26,3 @@ jobs:
...
-#     needs: [test-base]
-#     runs-on: ubuntu-latest
-#     strategy:
-#       fail-fast: false
-#       matrix:
-#         python-version: ["3.11"]
-#     steps:
-#       - uses: actions/checkout@v3
-#       - name: Run unit tests
-#         run: make test-unit
```
Shiggiddie, line 41: This covers all the same ground as test-iso, but all at once.  Prefer test-iso

.github/workflows/tests-pr.yaml
```diff
@@ -43,9 +43,6 @@ jobs:
       ISONE_API_USERNAME: ${{ secrets.ISONE_API_USERNAME }}
       ISONE_API_PASSWORD: ${{ secrets.ISONE_API_PASSWORD }}
       AESO_API_KEY: ${{ secrets.AESO_API_KEY }}
-      # Proxy for NYISO tests (NYISO blocks GitHub Actions IPs)
-      HTTPS_PROXY: ${{ matrix.iso == 'nyiso' && secrets.WEBSHARE_PROXY_URL || '' }}
-      HTTP_PROXY: ${{ matrix.iso == 'nyiso' && secrets.WEBSHARE_PROXY_URL || '' }}
```
Shiggiddie, line 48: This proxy was causing all nyiso runs to fail because GH Actions was trying to download the repo via proxy in checkout step

.github/workflows/tests-pr.yaml
```diff
@@ -58,7 +55,35 @@ jobs:
         run: uv sync
       - name: Install vcrpy
         run: uv pip install vcrpy
+      - name: Generate per-ISO fixture hash
+        id: fixture-hash
```
Shiggiddie, line 60: Hash created by inspecting S3 metadata for all the ISO's VCR cassettes

.github/workflows/tests-pr.yaml
```diff
@@ -58,7 +55,35 @@ jobs:
...
+        run: |
+          HASH=$(uv run python scripts/fixtures.py manifest --iso ${{ matrix.iso }})
+          echo "hash=$HASH" >> $GITHUB_OUTPUT
+      - name: Get fixture cache paths
+        id: fixture-paths
```
Shiggiddie, line 65: For the ISO, get the paths where VCR cassettes live

.github/workflows/tests-pr.yaml
```diff
@@ -58,7 +55,35 @@ jobs:
...
+        run: |
+          PATHS=$(uv run python scripts/fixtures.py cache-paths --iso ${{ matrix.iso }})
+          echo "paths<<EOF" >> $GITHUB_OUTPUT
+          echo "$PATHS" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+      - name: Cache VCR fixtures
```
Shiggiddie, line 71: Determine if cache hit from hash generated above.
Populate paths from above with cache hit entries, or clear to be populated by S3 downloads

.github/workflows/tests-pr.yaml
```diff
@@ -58,7 +55,35 @@ jobs:
...
+        id: vcr-cache
+        uses: actions/cache@v4
+        with:
+          path: ${{ steps.fixture-paths.outputs.paths }}
+          key: vcr-fixtures-${{ matrix.iso }}-${{ steps.fixture-hash.outputs.hash }}
+          restore-keys: |
+            vcr-fixtures-${{ matrix.iso }}-
+      - name: Download VCR fixtures from S3
```
Shiggiddie, line 79: Refresh cache with S3 downloads if not ht

.github/workflows/tests-pr.yaml
```diff
@@ -58,7 +55,35 @@ jobs:
...
+        if: steps.vcr-cache.outputs.cache-hit != 'true'
+        run: uv run python scripts/fixtures.py download --iso ${{ matrix.iso }}
+        continue-on-error: true
       - name: Run ISO tests
+        env:
+          # Proxy for NYISO tests (NYISO blocks GitHub Actions IPs)
+          HTTPS_PROXY: ${{ matrix.iso == 'nyiso' && secrets.WEBSHARE_PROXY_URL || '' }}
+          HTTP_PROXY: ${{ matrix.iso == 'nyiso' && secrets.WEBSHARE_PROXY_URL || '' }}
```
Shiggiddie, line 108: Moved from above

gridstatus/tests/source_specific/test_aeso.py
```diff
@@ -23,7 +23,12 @@
 class TestAESO(TestHelperMixin):
     @classmethod
     def setup_class(cls):
-        cls.iso = AESO(api_key=os.getenv("AESO_API_KEY"))
+        # Use a dummy API key in CI playback mode (record_mode=none) so the
+        # class can be constructed without real credentials.
+        api_key = os.getenv("AESO_API_KEY") or (
+            "DUMMY_KEY_FOR_VCR_PLAYBACK" if RECORD_MODE == "none" else None
+        )
+        cls.iso = AESO(api_key=api_key)
```
Shiggiddie: When running in CI, we don't to use a "real" AESO_API_KEY, since there's no need to run real API requests

gridstatus/tests/source_specific/test_caiso.py
```diff
@@ -100,10 +160,8 @@ def _check_load_forecast(
         "date, end",
         [
             (
-                pd.Timestamp.today(tz=iso.default_timezone).normalize()
-                - pd.Timedelta(days=5),
-                pd.Timestamp.today(tz=iso.default_timezone).normalize()
-                - pd.Timedelta(days=2),
+                pd.Timestamp("2025-11-01", tz="US/Pacific"),
```
Shiggiddie, line 167: Where possible, relative dates have been substituted for absolute dates so we match VCR cassettes

.github/workflows/tests-nightly.yaml
```diff
@@ -0,0 +1,96 @@
+name: Nightly Tests (Live Integration)
```
Shiggiddie, line 1: This was added to split out "true integration" tests from the PR review tests.  True integration tests are too flaky given rate limits, etc.  So instead, we do a nightly full integration test that is likely to fail until we address these issues.

In other words, all of the continued tech-debt has been confined to this test run instead of PR review test runs

Makefile
```diff
@@ -6,57 +6,58 @@ clean:
 	find . -name '*~' -delete
 	find . -name '.coverage.*' -delete
 
-PYTEST_CMD := uv run pytest -s -vv -n auto --reruns 5 --reruns-delay 3 --durations=25
+PYTEST_CMD := uv run pytest -s -vv -n auto --reruns 0 --durations=25
```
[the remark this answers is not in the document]
Shiggiddie: This was a temporary change to assist in faster test runs while debugging, this has been put back

.github/workflows/tests-commit.yaml
```diff
@@ -27,15 +27,41 @@ jobs:
...
+          path: gridstatus/tests/fixtures/
+          key: vcr-fixtures-all-${{ steps.fixture-hash.outputs.hash }}
+          restore-keys: |
+            vcr-fixtures-all-
+      - name: Download all VCR fixtures from S3
+        if: steps.vcr-cache.outputs.cache-hit != 'true'
+        run: uv run python scripts/fixtures.py download
+        continue-on-error: true
+      - name: Run unit tests
+        run: make test-unit
```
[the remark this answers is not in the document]
Shiggiddie: `test-unit` job has been removed

Makefile
```diff
@@ -6,57 +6,58 @@ clean:
...
 
 .PHONY: test-ieso
 test-ieso:
-	$(PYTEST_CMD) gridstatus/tests/source_specific/test_ieso.py
+	$(PYTEST_CMD) $(UNIT_ONLY) gridstatus/tests/source_specific/test_ieso.py
 
 .PHONY: test-misc
 test-misc:
-	$(PYTEST_CMD) gridstatus/tests/source_specific/test_gridstatus.py gridstatus/tests/source_specific/test_lmp_config.py
+	$(PYTEST_CMD) $(UNIT_ONLY) gridstatus/tests/source_specific/test_gridstatus.py gridstatus/tests/source_specific/test_lmp_config.py
```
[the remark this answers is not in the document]
Shiggiddie: This was intentional. Runs in nightly only, where live integration tests are the point. Not CI-blocking.

gridstatus/tests/source_specific/test_eia.py
```diff
@@ -451,6 +451,7 @@ def _check_generators_data(
             assert pd.api.types.is_integer_dtype(df[col])
 
 
+@pytest.mark.integration
```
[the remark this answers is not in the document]
Shiggiddie: Test is @pytest.mark.integration, although in the future we should address to remove cassette entirely in all likelihood

.github/workflows/tests-nightly.yaml
```diff
@@ -0,0 +1,72 @@
...
+            vcr-fixtures-all-
+      - name: Download all VCR fixtures from S3
+        if: steps.vcr-cache.outputs.cache-hit != 'true'
+        run: uv run python scripts/fixtures.py download
+        continue-on-error: true
+      - name: Run all tests with coverage
+        env:
+          # Proxy for NYISO tests (NYISO blocks GitHub Actions IPs)
+          HTTPS_PROXY: ${{ secrets.WEBSHARE_PROXY_URL }}
+          HTTP_PROXY: ${{ secrets.WEBSHARE_PROXY_URL }}
```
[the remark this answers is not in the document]
Shiggiddie: Valid observation but nightly runs one combined test-cov, we can't conditionalize per-ISO without restructuring the nightly workflow.  This area is generally tech debt central, so we can address outside of the scope of this PR

.github/workflows/tests-pr.yaml
```diff
@@ -38,14 +38,12 @@ jobs:
...
       ERCOT_PUBLIC_API_SUBSCRIPTION_KEY: ${{ secrets.ERCOT_PUBLIC_API_SUBSCRIPTION_KEY }}
+      ERCOT_ESR_API_SUBSCRIPTION_KEY: ${{ secrets.ERCOT_ESR_API_SUBSCRIPTION_KEY }}
       MISO_API_PRICING_SUBSCRIPTION_KEY: ${{ secrets.MISO_API_PRICING_SUBSCRIPTION_KEY }}
       MISO_API_LOAD_GENERATION_AND_INTERCHANGE_SUBSCRIPTION_KEY: ${{ secrets.MISO_API_LOAD_GENERATION_AND_INTERCHANGE_SUBSCRIPTION_KEY }}
       ISONE_API_USERNAME: ${{ secrets.ISONE_API_USERNAME }}
       ISONE_API_PASSWORD: ${{ secrets.ISONE_API_PASSWORD }}
       AESO_API_KEY: ${{ secrets.AESO_API_KEY }}
-      # Proxy for NYISO tests (NYISO blocks GitHub Actions IPs)
-      HTTPS_PROXY: ${{ matrix.iso == 'nyiso' && secrets.WEBSHARE_PROXY_URL || '' }}
-      HTTP_PROXY: ${{ matrix.iso == 'nyiso' && secrets.WEBSHARE_PROXY_URL || '' }}
```
Shiggiddie, line 48: This is moved to the per-ISO level for "nyiso" specifically instead of top level proxy

.github/workflows/tests-pr.yaml
```diff
@@ -38,14 +38,12 @@ jobs:
       ERCOT_API_USERNAME: ${{ secrets.ERCOT_API_USERNAME }}
       ERCOT_API_PASSWORD: ${{ secrets.ERCOT_API_PASSWORD }}
       ERCOT_PUBLIC_API_SUBSCRIPTION_KEY: ${{ secrets.ERCOT_PUBLIC_API_SUBSCRIPTION_KEY }}
+      ERCOT_ESR_API_SUBSCRIPTION_KEY: ${{ secrets.ERCOT_ESR_API_SUBSCRIPTION_KEY }}
```
Shiggiddie, line 41: Many of our classess still require valid keys of some sort to instantiate even if we don't need them to make requests

.github/workflows/tests-pr.yaml
```diff
@@ -58,41 +56,74 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: ${{ steps.fixture-paths.outputs.paths }}
+          key: vcr-fixtures-${{ matrix.iso }}-${{ steps.fixture-hash.outputs.hash }}
+          restore-keys: |
+            vcr-fixtures-${{ matrix.iso }}-
+      - name: Download VCR fixtures from S3
+        if: steps.vcr-cache.outputs.cache-hit != 'true'
+        run: uv run python scripts/fixtures.py download --iso ${{ matrix.iso }}
+        continue-on-error: true
```
Shiggiddie, line 82: We attempt to grab fixtures from GH Action's free cache before before pulling from S3 if we can help it

gridstatus/caiso/daily_energy_storage.py
```diff
@@ -30,6 +31,10 @@ def load_daily_energy_storage_report(
     tz: str,
     verbose: bool = False,
 ) -> tuple[str, pd.Timestamp]:
+    if isinstance(date, str) and date.lower() in ("latest", "today"):
+        raise NotSupported(
+            "Daily energy storage reports are only available for historical dates",
+        )
```
Shiggiddie, line 37: This was added to ensure that the expected error from the pre-existing test `test_daily_energy_storage_latest_not_supported` is fulfilled.

I suspect at some point changes wer emade to the code that broke the test, but because we have so many failing tests we didn't notice 😅

gridstatus/tests/source_specific/test_aeso.py
```diff
@@ -23,7 +23,10 @@
 class TestAESO(TestHelperMixin):
     @classmethod
     def setup_class(cls):
-        cls.iso = AESO(api_key=os.getenv("AESO_API_KEY"))
+        api_key = os.getenv("AESO_API_KEY") or (
+            dummy_credential("AESO_API_KEY") if RECORD_MODE == "none" else None
```
Shiggiddie, line 27: `dummy_credential` is meant to cover three scenarios:

1. Real key present (dev machine or integration CI): use it for real API calls.
2. No key, playback mode (record_mode == "none" in PR CI): use the dummy, since VCR intercepts the HTTP request and replays from cassette before any auth header is actually used. Cassettes don't contain real credentials anyway.
3. No key, recording mode: pass None, which lets the constructor crash which is what we want since you need a real key to record.

gridstatus/tests/source_specific/test_aeso.py
```diff
@@ -32,10 +35,13 @@ def _check_supply_and_demand(self, df: pd.DataFrame) -> None:
 
         assert df.dtypes["Time"] == f"datetime64[ns, {self.iso.default_timezone}]"
 
+    @pytest.mark.skip(
+        reason="AESO API credentials revoked / 401 - https://www.notion.so/33de835f42aa81d699e4c0e82dd008e1"
+    )
```
Shiggiddie, line 40: Throughout the PR, for a test that isn't currently passing, a tech debt task has been created so we can follow up

gridstatus/tests/source_specific/test_aeso.py
```diff
@@ -32,10 +35,13 @@ def _check_supply_and_demand(self, df: pd.DataFrame) -> None:
...
+    @pytest.mark.integration
```
Shiggiddie, line 41: This designates a test not suitable for unit testing since it needs real API esponses

gridstatus/tests/source_specific/test_aeso.py
```diff
@@ -660,13 +703,13 @@ def _check_transmission_outages(self, df: pd.DataFrame) -> None:
 
         assert (df["Interval End"] >= df["Interval Start"]).all()
 
+    @pytest.mark.integration
     def test_get_transmission_outages_latest(self):
         """Test getting latest transmission outages data."""
-        with api_vcr.use_cassette("test_get_transmission_outages_latest.yaml"):
```
Shiggiddie, line 665: VCR cassette usage is removed for tests that use relative dates

gridstatus/tests/source_specific/test_caiso.py
```diff
@@ -100,10 +164,8 @@ def _check_load_forecast(
...
+                pd.Timestamp("2025-11-03", tz="US/Pacific"),
```
Shiggiddie, line 168: Where possible, relative dates have been replaced with static absolute dates to enable VCR cassette matching

gridstatus/tests/source_specific/test_caiso.py
```diff
@@ -23,6 +23,70 @@ class TestCAISO(BaseTestISO):
 
     trading_hub_locations = CAISO().trading_hub_locations
 
+    # --- BaseTestISO overrides: mark inherited today/latest/relative-date tests ---
```
Shiggiddie, line 26: A base test that calls self.iso.get_fuel_mix("latest") will either:                                                                                                                                     
                                                                                                                                               
* Make a real HTTP request to CAISO (no cassette to play back) sometimes flaky.
* Produce a dynamically-named cassette that shifts every run
                                                                                                                                               
Marking them `@pytest.mark.integration` excludes them from the -m "not integration" filter in the PR test matrix, while leaving them visible to the nightly integration workflow that does hit live APIs.

gridstatus/tests/base_test_iso.py
```diff
@@ -16,7 +16,12 @@ def local_today(self):
         return self.local_now().date()
 
     def local_start_of_day(self, date: str | pd.Timestamp | None):
-        return pd.Timestamp(date).tz_localize(self.iso.default_timezone).normalize()
+        ts = pd.Timestamp(date)
+        if ts.tzinfo is None:
+            ts = ts.tz_localize(self.iso.default_timezone)
+        else:
+            ts = ts.tz_convert(self.iso.default_timezone)
+        return ts.normalize()
```
Shiggiddie, line 24: `local_start_of_day` is supposed to return midnight in the ISO's local timezone, but the previous code only localized tz-naive inputs instead of tz-naive AND caller passed a tz-aware timestamp in a different timezone

gridstatus/tests/vcr_utils.py
```diff
@@ -67,25 +104,114 @@ def clean_cassettes(path: str):
     os.makedirs(path, exist_ok=True)
 
 
+class _SkipOrCassette:
```
Shiggiddie, line 107: We can eadjust how this works, but the idea is that we don't want failures due to S3 files not existing to immediately result in test failures (maybe this is the wrong call?)

gridstatus/spp.py
```diff
@@ -2728,6 +2728,7 @@ def get_interchange_real_time(
         df = df.rename(
             columns={
                 "GMTTIME": "GMTTime",
+                "#GMTTIME": "GMTTime",
```
Shiggiddie, line 2731: Found in testing to exist

gridstatus/miso_api.py
```diff
@@ -1381,6 +1381,9 @@ def _get_medium_term_load_forecast(
         if "interval" in df.columns:
             df = df.drop(columns=["interval"])
 
+        if "init" in df.columns:
+            df = df.drop(columns=["init"])
```
Shiggiddie, line 1385: Was very confused why this wasn't failing in production :/

scripts/fixtures.py
Shiggiddie, line 1: Code for uploading VCR cassets to and downloading from S3

gridstatus/tests/source_specific/test_caiso.py
Kladar: just noting that most of the relative dates are there because historical data rolls off and becomes unavailable at a certain point, so updating the cassettes will not be possible on certain datasets with absolute historical dates 
edit: "most" might not be correct, but many

.github/workflows/tests-pr.yaml
WillKoehrsen: If we need comments to explain these steps, maybe the comments should be in the workflow.

gridstatus/tests/source_specific/test_caiso.py
WillKoehrsen: In addition to Ryan's comments, we use relative dates on purpose since ISOs frequently change upstream data. I would prefer our tests break in these cases until we update the scraper.

CONTRIBUTING.md
```diff
@@ -114,6 +114,78 @@ All pull requests will be reviewed by one of our maintainers before being merged
...
+# Override the 4xx/5xx safety scan if you really need to:
+make fixtures-upload iso=<iso> force=1
+```
+
+5. **Commit and push** — the cassette files are NOT committed to git. CI downloads them from S3 (with GHA caching).
+
+### Important Guidelines
+
+- **Use fixed dates** — never use `"today"`, `"latest"`, or `pd.Timestamp.now()` in VCR-wrapped tests. These generate different URLs each run, breaking cassette matching.
+- **Use recent dates** — ISO APIs typically expire data after 30-90 days. Use dates within the last few weeks.
```
WillKoehrsen, line 168: These two points are contradictions. If we use fixed dates, the tests quickly become outdated. 

> ISO APIs typically expire data after 30-90 days

This is not true across ISOs.

CONTRIBUTING.md
```diff
@@ -114,6 +114,78 @@ All pull requests will be reviewed by one of our maintainers before being merged
...
+```
+
+3. **Verify the cassette is clean** — `make fixtures-upload` scans every cassette for `code: 4xx`/`code: 5xx` responses and refuses to upload if any are found. You can also scan manually:
+
+```bash
+grep "code: [45]" gridstatus/tests/fixtures/<iso>/vcr_cassettes/test_get_my_new_dataset_2026-03-15.yaml
+# Should return nothing
+```
+
+4. **Upload to S3** — requires AWS credentials:
```
WillKoehrsen, line 155: Why is the burden on the developer to upload cassettes? In my mind, cassettes should be uploaded to S3 when the test suite runs in CI so there is a single upload source.

scripts/fixtures.py
```diff
@@ -0,0 +1,300 @@
...
+    python scripts/fixtures.py upload [--iso ISO_NAME] [--force]
+    python scripts/fixtures.py cache-paths [--iso ISO_NAME]
+    python scripts/fixtures.py manifest [--iso ISO_NAME]
+
+Commands:
+    download     Download fixtures from S3 (public, no credentials needed).
+                 Exits non-zero if any per-source sync fails.
+    upload       Upload fixtures to S3 (requires AWS credentials). Refuses
+                 by default if any cassette contains a 4xx/5xx response;
+                 pass --force to override.
```
WillKoehrsen, line 15: Who is allowed to upload fixtures to S3?

CONTRIBUTING.md
```diff
@@ -114,6 +114,78 @@ All pull requests will be reviewed by one of our maintainers before being merged
...
+- **Use fixed dates** — never use `"today"`, `"latest"`, or `pd.Timestamp.now()` in VCR-wrapped tests. These generate different URLs each run, breaking cassette matching.
+- **Use recent dates** — ISO APIs typically expire data after 30-90 days. Use dates within the last few weeks.
+- **Don't upload error cassettes** — if the API returned 4xx/5xx errors, delete the cassette and re-record.
+- **Mark live-only tests as integration** — if a test inherently needs live data (e.g., uses relative dates, or the API uses cache-busting params), add `@pytest.mark.integration`.
+- **Cassette naming** — use descriptive names including the date: `test_get_lmp_historical_2026-03-15.yaml`.
+
+### Makefile Targets
+
+```bash
+make fixtures-download               # Download all fixtures from S3
```
WillKoehrsen, line 176: Since this data is public, what is preventing a malicious actor from repeatedly downloading these cassettes to incur a large AWS bill for Grid Status?

.github/workflows/tests-nightly.yaml
```diff
@@ -0,0 +1,99 @@
+name: Nightly Tests (Live Integration)
+
+# Runs the full live-integration test suite with coverage once per day at
+# midnight UTC. This was previously the test-all-coverage job in
+# tests-pr.yaml but it's too slow and depends on too many flaky live ISO
+# APIs to be a per-PR gate. Failures here are signal for follow-up tickets,
+# not blockers for individual PRs.
+#
+# Can also be triggered manually via the Actions tab (workflow_dispatch).
+
```
Kladar, line 10: Since we're already running this each night and it does the heavy lifting of grabbing from source, I'm thinking it might be a good chance to also update the fixtures ("blue-green" style) by writing them to another bucket and then doing the swap (or failing-out, leaving the existing fixtures fully intact). 

This could also be a good place to lock down the bucket access as well, since I think write-permissions are another point of discussion, and only give github actions (+ an internal admin) write permissions to the bucket, and we ideally just run this action if we need to update the fixtures manually.

CONTRIBUTING.md
WillKoehrsen: I've blocked all public access to the bucket. We should be using Cloudflare R2 which has zero egress costs.

https://www.cloudflare.com/developer-platform/products/r2/

gridstatus/tests/vcr_utils.py
```diff
@@ -67,25 +104,114 @@ def clean_cassettes(path: str):
...
+    # Several ISO endpoints (ERCOT's IceDocListJsonWS and public-reports, CAISO's
+    # outlook/history CSVs) append a random ``_=NNN`` cache-buster query param
+    # on every request. Use a custom URI matcher that strips any underscore-
+    # prefixed query param so cassettes replay cleanly across runs.
+    vcr_instance = vcr.VCR(
+        cassette_library_dir=cassette_dir,
+        record_mode=record_mode,
+        match_on=["ercot_uri", "method"],
+        before_record=lambda request: before_record_callback(request, source),
+        filter_headers=[
```
WillKoehrsen, line 208: We are not adequately scrubbing requests leading to leaked credentials.

Shiggiddie:
Refreshing approach, outline here:
https://www.notion.so/gridstatus/S3-Fixture-Gameplan-349e835f42aa80b299f3cd4e32d4b1a2?utm_content=349e835f-42aa-80b2-99f3-cd4e32d4b1a2&utm_campaign=T06MGAMJLG1&pvs=6

PR #845, opened 2026-04-09
Fix CAISO Fuel Regions

WillKoehrsen:
## Summary

- Fixes CAISO().get_fuel_regions() by switching to `requests.get` for the URL to solve SSL verification problems
- Restores SSL verification for all CAISO requests as this was supposed to be a temporary patch: https://github.com/gridstatus/gridstatus/pull/772

### Validation

- Run tests with `VCR_RECORD_MODE=all uv run pytest -s -vv gridstatus/tests/source_specific/test_caiso.py` 
  - We need to run all the tests because the SSL verification we added back in affects most CAISO datasets
- Run `CAISO().get_fuel_regions()` and assure it fetches the data

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -364,7 +364,7 @@ class CAISO(ISOBase):
 
         retry_num = 0
         while retry_num < max_retries:
-            r = requests.get(url, verify=False)
+            r = requests.get(url, verify=True)
 
             if r.status_code == 200:
                 break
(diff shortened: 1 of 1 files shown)


gridstatus/caiso/caiso.py
```diff
@@ -364,7 +364,7 @@ def _get_oasis(
 
         retry_num = 0
         while retry_num < max_retries:
-            r = requests.get(url, verify=False)
```
WillKoehrsen, line 367: This was always intended as a temporary patch to resolve SSL issues with CAISO OASIS. Those problems seem to have been resolved so we should make sure we are verifying requests. 

https://github.com/gridstatus/gridstatus/pull/772

Kladar (approved):

Kladar:
Going to merge this so we get the fix

PR #846, opened 2026-04-10
fix(ENG 3638): replace NAs in daily storage

Kladar:
## Summary
There are some NA values in integers columns, and these also affect the time sampling index creation when they appear. Replaces them with null


Kladar:
Going to do this elsewhere

PR #847, opened 2026-04-13
feat(ENG 2022): Add the PJM Emergency Postings dataset

Kladar:
## Summary
Adds a method to grab the PJM Emergency Postings from the XML feed.

```
from gridstatus import PJM
pjm = PJM()

df = pjm.get_emergency_postings()
print(df)
print(df.dtypes)
```

### Details
I went back and forth on this one for a while on what route to take. [As the help page indicates](https://www.pjm.com/markets-and-operations/etools/emerg-procedure), there is an XML feed and a PJM CLI. The PJM CLI is a large dependency that I'd rather avoid. And so while there is an XML feed, it does require some involved session management. Cookies stay alive for 30 minutes with inactivity, 8 hours with activity, and passing these around is a bit of a pain. 

However, all of the information is available to GUEST users, so really all of that session management is not needed to access the data. And the HTML doesn't have every piece of information, the XML does. They even provide an XML button on the HTML page that is publicly available. 
CleanShot 2026-04-13 at 11 03 10@2x

So really the most straightforward path is to simply request the page, and then download the XML and parse it. That way we can avoid session management and keep alive tactics, the method works for this opensource scraper, and the XML has all the information we need. So that's what I do here. 

PJM just asks that automated scrapers don't hit more than 1200 requests per minute, which is easily abided by. 
CleanShot 2026-04-13 at 11 16 56@2x

### Tests

```
uv run pytest gridstatus/tests/source_specific/test_pjm.py -k "emergency" -vv
```

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -23,6 +23,7 @@ from gridstatus.gs_logging import logger
 from gridstatus.lmp_config import lmp_config
 from gridstatus.pjm_constants import (
     DEFAULT_RETRIES,
+    EMERGENCY_POSTINGS_GUEST_DASHBOARD_URL,
     HUB_NODE_IDS,
     LOCATION_TYPES,
     PRICE_NODE_IDS,
@@ -72,10 +73,6 @@ class PJM(ISOBase):
     load_forecast_endpoint_name = "load_frcstd_7_day"
     load_forecast_historical_endpoint_name = "load_frcstd_hist"
     load_forecast_5_min_endpoint_name = "very_short_load_frcst"
-
-    EMERGENCY_POSTINGS_GUEST_DASHBOARD_URL = (
-        "https://emergencyprocedures.pjm.com/ep/pages/dashboard.jsf"
-    )
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


gridstatus/pjm.py
```diff
@@ -3790,6 +3793,129 @@ def get_sync_reserve_events(
...
+        root = ET.fromstring(xml_bytes)
+        rows: list[dict] = []
+        for msg in root.iter("EmergencyMessage"):
+            mid = msg.findtext("messageId")
+            if mid is None:
+                continue
+
+            eff_start = msg.findtext("effectiveStartTime") or msg.findtext(
+                "applicableStartTime",
+            )
```
WillKoehrsen: What's the difference between the applicable and effective start time? Should we be capturing both these pieces of information?

Image

WillKoehrsen: For this example, in the dashboard PJM has an Effective Start Time but not an Effective End Time. However, our dataframe has both an interval start and end with the start coming from the `effectiveStartTime` and the end from the `applicableEndTime`. This seems potentially confusing if our data has an end time but PJM's dashboard does not.

Image

Image

Kladar: Good catch, yeah I think the `effectiveStartTime` is very similar to publish time and there is almost never an `effectiveEndTime` (they are just cancelled). But `applicableStartTime`/`endTime` are the actual event boundaries. I've asked the markets team what we want to do.

Kladar: Okay, I agree with the new approach we decided on, which is to keep all 4 columns (effective start, effective end, applicable start, applicable end) and name them as such. Hopefully good now.

WillKoehrsen (approved):
Code and data looks good and tests pass locally.

PR #850, opened 2026-04-15
fix(ENG 3638): handle all encountered value types

Kladar:
## Summary
Coerces all the encountered null, nan, and na types encountered in the chart.

PR #851, opened 2026-04-15
Add /code-review Claude Code slash command

Shiggiddie:
## Summary

Adds a `/code-review` Claude Code slash command tailored to this library's conventions.

The command walks a 6-step flow (collect inputs → load guidelines → build a mental model → review the diff → validate against best practices → report out) and flags risk zones specific to this repo:

- **VCR cassettes**: fixed recent dates (not `today`/`latest`), no 4xx/5xx responses in cassettes, S3 sync via `make fixtures-upload iso=<iso>` (cassettes are not committed to git)
- **Python 3.10–3.12 compatibility**: no syntax that requires newer versions
- **Public API stability**: renamed/removed methods are breaking changes for library users
- **ISO parser consistency**: returned DataFrame columns/dtypes/timezones stay aligned across ISOs
- **Test markers**: `@pytest.mark.integration` / `@pytest.mark.slow` applied to live-API tests

Runs `make lint` as part of validation and outputs a numbered **Next Steps** list so reviewers can reply with issue numbers.

## Test plan

- [x] Open a Claude Code session on a branch with a non-trivial diff and run `/code-review`; confirm output follows the 6-step flow and ends with a numbered Next Steps list

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a new markdown command definition only, with no runtime code or behavior changes.
> 
> **Overview**
> Introduces a new Claude Code slash command definition at `.claude/commands/code-review.md` that prescribes a 6-step review workflow (diff collection, guideline loading, risk-zone checks, validation, and a structured report).
> 
> The instructions emphasize repo-specific review pitfalls (VCR cassette handling, Python 3.10–3.12 compatibility, parser/timezone consistency, API stability) and require output to end with a numbered "Next Steps" list for easy follow-up.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf5e7fa5688f2509f95f8e86302087dd9318863c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/commands/code-review.md
+++ b/.claude/commands/code-review.md
@@ -1,0 +1,46 @@
+## Review code changes against CLAUDE.md guidelines to catch issues and ensure maintainability
+
+Follow the steps below and only output the final Report Out section.
+
+1. **Collect Inputs**
+   - Review changes in this branch only
+   - Base branch for the diff: `$ARGUMENTS` (default to `main` if empty)
+   - Use `git diff <base>...HEAD` to show all changes
+   - Note any context the user provided (goal, constraints, known tradeoffs)
+
+2. **Load Guidelines**
+   - Read the workspace `../CLAUDE.md` for monorepo-wide rules
+   - Read any repo-level `AGENTS.md` / `CONTRIBUTING.md` if touched areas need it
+   - No repo-local `CLAUDE.md` exists here — rely on conventions evident in nearby code
+
+3. **Build a Mental Model**
+   - Identify what the change is trying to accomplish
+   - Map what's touched: ISO parser (`gridstatus/<iso>.py`), base classes, HTTP helpers, tests, fixtures
+   - Flag risk zones:
+     - **VCR cassettes**: new tests must use fixed recent dates (not `today`/`latest`), cassettes must not contain 4xx/5xx responses, and must be uploaded via `make fixtures-upload iso=<iso>` — they are NOT committed to git
+     - **Python version compat**: library supports Python 3.10–3.12; no 3.13-only syntax (no `type` statement aliases, etc.)
+     - **ISO API breakage**: parsers that silently return empty DataFrames on schema changes
+     - **Timezone handling**: every ISO has its own local tz; confirm tz-aware timestamps
+     - **Network retries and rate limits**
+     - **Public API surface**: renamed/removed methods are breaking changes for library users
+
+4. **Review the Diff**
+   - Go file by file
+   - Verify correctness and intent
+   - Look for bugs, logic errors, regressions, and common AI-generated mistakes (unused abstractions, inconsistent naming, unnecessary complexity, over-broad `except Exception`)
+   - For parser changes: confirm returned DataFrame columns/dtypes match sibling ISOs
+   - For new tests: confirm `api_vcr.use_cassette(...)` is used with a fixed recent date; confirm cassette recording plan is explicit
+   - For `@pytest.mark.integration` / `@pytest.mark.slow` usage: confirm tests that hit live APIs are marked so CI can skip them
+
+5. **Validate Against CLAUDE.md + Best Practices**
+   - Run `make lint` and surface any failures
+   - For changed parser files, confirm `make test-<iso>` is runnable (don't execute unless the user asks)
+   - Flag important best-practice issues not covered by guidelines
+   - Cite the relevant guideline / convention for each issue when applicable
+
+6. **Report Out**
+   - List all issues directly. NEVER add a higher-level summary of the review
+   - Pair each issue with a clear, actionable fix
+   - Order issues with sections by priority (blocking, important, nit) with global numbering so each issue can unambiguously be referenced
+   - At the end, list out each issue one by one with a concise (no more than 20 words) description, so the user can reply with numbers. Call this section "Next Steps". The last line should say "Which issue(s) do you want to fix first?".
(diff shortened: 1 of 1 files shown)


Shiggiddie:
Example output for PR#850

PR #850 — kladar/fix/caiso-storage-types (base: main)                                                                                        
                                                                                                                                               
  Blocking                                                                                                                                     
                                                                                                                                               
  None. No public API changes (all touched functions are underscore-prefixed / module-private), no cassette churn, no Python-version-specific  
  syntax, test coverage added for new behavior.                                                                                                
                                                                                                                                               
  Important                                                                                                                                    
                                               
  1. Silent NaN coercion loses debug signal when CAISO ships an unexpected shape. _coerce_chart_element catches TypeError/ValueError from      
  float(v) and returns nan with no log — so if CAISO later switches to nested arrays, dicts, or a totally new encoding, downstream sees "just
  NaNs" with no signal that parsing is broken. Add a logger.warning("unexpected chart element type=%s value=%r", type(v).__name__, v) in the   
  final except branch (and possibly when a bare string can't be parsed and isn't in the known sentinel set). Keeps the NaN fallback but gives
  operators something to grep for.             
  2. Worst-case fetch time is ~8 minutes per day. 8 candidate URLs × timeout=60 = up to 480 s if CAISO stops responding on all of them.
  Consider (a) a shorter per-request timeout (10–15 s is plenty for a static HTML file), and/or (b) a total wall-clock budget across the       
  candidate loop. As-is, an ETL backfilling a year could stall for hours on a single CAISO outage.
  3. No negative-path test covering "all candidates 404". The three new fetch tests all expect a success on some candidate. Add a fourth test  
  that asserts NoDataFoundException is raised with the final HTTP status in the message when every candidate returns 404. That path is easy to 
  regress (e.g. swap response is None or response.status_code != 200 to just response.status_code != 200 and pre-loop response = None still
  leaves you there).                                                                                                                           
  4. types-requests added to mypy's additional_dependencies may surface new errors elsewhere. Adding stubs to the mypy hook typically uncovers
  previously-silent type issues across the project. Confirm pre-commit run --all-files (or make lint) still passes on this branch, and if
  additional type fixes were needed elsewhere, they belong in this PR so pre-existing main stays green.

  Nit

  5. _coerce_chart_element name implies it could return non-float. It always returns float (NaN on failure). Consider _to_float_or_nan or      
  _as_chart_float for clarity at call sites.
  6. PR body describes only half the change. Summary says "Coerces all the encountered null, nan, and na types encountered in the chart" — but 
  the diff also adds a 4-slug URL fallback scheme for the CAISO report URL. Add a paragraph to the PR description so future git-blame /        
  changelog work can find both.                
  7. "Details" section in the PR body is empty. (Also true of the sibling PR on gs-etl.) Either delete the heading or fill it in.              
  8. _downsample_* now always returns a new list on the invalid-length short-circuit (return list(values)). Previously returned values         
  directly. Functionally safer (no aliasing), but a subtle behavior change — worth a one-word note in the PR body in case any caller relied on 
  identity.                                                                                                                                    
  9. Consider representing candidate_urls as (slug_type, url) pairs. Currently it's a flat list of 8 strings with seen_urls dedup. A dict[str, 
  str] (slug type → URL) or a tuple-of-tuples documents each variant's name (helpful for logging "matched on slug_legacy") and avoids the need 
  for a post-hoc dedup set.                    
  10. raise RuntimeError("unreachable") is dead code. The preceding raise NoDataFoundException fires in every remaining branch. On Python 3.11+
   typing.assert_never is more idiomatic, but given this library supports 3.10, leave as-is — worth a # pragma: no cover so coverage doesn't   
  complain.                                    
  11. Sentinel string set could include "--", "null" lowercase via .upper() — already handled. Good. Consider whether CAISO ever uses "?" or   
  blank-looking whitespace (NBSP); the t = v.strip() handles ASCII whitespace but not U+00A0. Low-probability, low-impact.                     
  12. Timeout literal 60 duplicated across 8 sites (via loop: once now). If you go with issue #2's shorter timeout, make it a module-level
  constant for easy tuning.                                                                                                                    
                                                                                                              
  Next Steps                                                                                                                                   
                                                                                                              
  1. Log a warning when _coerce_chart_element falls back to NaN for an unexpected type.                                                        
  2. Tighten per-request timeout and/or add a total-time budget across the URL fallback loop.                 
  3. Add a test asserting NoDataFoundException when every candidate URL returns 404.                                                           
  4. Confirm pre-commit / mypy stays green repo-wide after adding types-requests; include any fan-out fixes in this PR.                        
  5. Rename _coerce_chart_element to something that communicates "always-float".                                                               
  6. Expand the PR description to mention the URL-slug fallback work.                                                                          
  7. Fill in or remove the empty "Details" heading in the PR body.                                                                             
  8. Call out the new-list-copy behavior change in _downsample_* short-circuit in the PR body.                                                 
  9. Represent candidate URLs as named pairs so logging can report which slug type matched.                                                    
  10. Mark the "unreachable" RuntimeError with # pragma: no cover.                                                                             
  11. Consider whether NBSP / other odd whitespace needs handling in the sentinel check.                                                       
  12. Extract the 60s timeout into a module-level constant.                                                                                    
                                                                                                                                               
  Which issue(s) do you want to fix first?

Kladar (approved):

PR #852, opened 2026-04-15
fix(ENG 2023): ERCOT Messages fixes

Kladar:
## Summary
Some fixes for historical data and retrying/backoff of hitting the wayback machine


gridstatus/ercot.py
```diff
@@ -3367,11 +3378,10 @@ def _fetch_operations_messages_from_url(
...
 
-        return (
-            df[["Time", "Notice", "Type", "Status"]]
-            .sort_values("Time")
-            .reset_index(drop=True)
-        )
+        expected_cols = ["Time", "Notice", "Type", "Status"]
+        available_cols = [c for c in expected_cols if c in df.columns]
+
+        return df[available_cols].sort_values("Time").reset_index(drop=True)
```
WillKoehrsen: If one of the columns is missing, do we want to insert it as all NANs? That way we have a consistent output from the method.

WillKoehrsen (approved):
One minor comment otherwise should be good.

gridstatus/ercot.py
Kladar: makes sense, updated

Kladar:
ruff is being a bit weird with the ercot api notebook file but I think it's fine

PR #855, opened 2026-04-17
Use HTTPS for CAISO OASIS requests

djchen:
## Summary
- switch CAISO OASIS requests from `http://` to `https://`
- keep the existing verified TLS request path intact

## Why
CAISO OASIS is reachable over HTTPS with valid certificate verification, so these requests should not use plaintext HTTP.


WillKoehrsen (approved):
CAISO data retrieval still works with this update. Thanks for the contribution!

PR #857, opened 2026-04-20
feat(ENG-2896): Add ERCOT Monthly CRR Auction scrapers (5 datasets)

Kladar:
## Summary

Adds five new ERCOT scrapers for the monthly CRR auction bundle published via MIS report NP7-803-M (report_type_id 11201). Each monthly ZIP contains five CSVs; one public method per CSV.

New public methods on `Ercot`:

- `get_crr_auction_bids_offers_monthly`
- `get_crr_base_loading_monthly`
- `get_crr_binding_constraints_monthly`
- `get_crr_market_results_monthly`
- `get_crr_source_sink_shadow_prices_monthly`

Each accepts a `date` (and optional `end`) that is interpreted as a month-start (Central Time) via `@support_date_range(frequency=None)`

`_get_monthly_crr_auction_frames` downloads every matching monthly
ZIP via the existing `utils.get_zip_folder` helper, passes it to
`_handle_monthly_crr_auction_zip` to split the bundle into per-dataset
DataFrames, then each public method renames source columns to Grid Status column names
and stamps tz-aware `Interval Start` / `Interval End` using the auction month.

## Implementation notes

- New constant `MONTHLY_CRR_AUCTION_RESULTS_RTID = 11201` in `ercot.py`.
- Friendly names like `JUN2025MonthlyCRRAuctionResults` are parsed with a
  dedicated helper (`_parse_crr_auction_month`) since
  `parse_timestamp_from_friendly_name` expects underscores.
- `Path` is derived as `Source + "-" + Sink` for the three datasets
  that expose Source and Sink columns (auction bids/offers, base loading,
  market results).

## Testing

- Added six unit tests (`TestErcot`) that mock `_get_documents` and
  `utils.get_zip_folder` with small synthetic ZIPs. The raw monthly
  CRR ZIPs are ~25-35 MB each, so fixtures can be large.
- Live smoke tested against the real June 2025 monthly CRR auction
  via `scripts/ercot_crr_monthly_tester.py` (gitignored):
  - auction bids/offers: 397,648 rows
  - base loading: 56,460 rows (CRR ID unique)
  - binding constraints: 929 rows
  - market results: 74,588 rows (CRR ID unique)
  - source/sink shadow prices: 2,595 rows

## Test plan

- [x] `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot.py -k crr`
  (6 tests, all passing, no network)

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -6439,10 +6439,16 @@ class Ercot(ISOBase):
         )
 
         all_df = []
-        for month_start, raw in frames:
+        for _month_start, raw in frames:
             df = raw.copy()
-            df["Interval Start"] = month_start
-            df["Interval End"] = month_start + pd.DateOffset(months=1)
+            df["Interval Start"] = pd.to_datetime(
+                df["StartDate"],
+                format="%m/%d/%Y",
+            ).dt.tz_localize(self.default_timezone)
+            df["Interval End"] = pd.to_datetime(
+                df["EndDate"],
+                format="%m/%d/%Y",
+            ).dt.tz_localize(self.default_timezone)
             df = df.rename(
                 columns={
                     "BidType": "Bid Type",
@@ -6593,10 +6599,16 @@ class Ercot(ISOBase):
         )
 
         all_df = []
-        for month_start, raw in frames:
+        for _month_start, raw in frames:
             df = raw.copy()
-            df["Interval Start"] = month_start
-            df["Interval End"] = month_start + pd.DateOffset(months=1)
+            df["Interval Start"] = pd.to_datetime(
+                df["StartDate"],
+                format="%m/%d/%Y",
+            ).dt.tz_localize(self.default_timezone)
+            df["Interval End"] = pd.to_datetime(
+                df["EndDate"],
+                format="%m/%d/%Y",
+            ).dt.tz_localize(self.default_timezone)
             df = df.rename(
                 columns={
                     "CRR_ID": "CRR ID",
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -187,6 +187,11 @@ DAM_TOTAL_ENERGY_SOLD_RTID = 12334
 # https://www.ercot.com/mp/data-products/data-product-details?id=np1-301
 COP_ADJUSTMENT_PERIOD_SNAPSHOT_RTID = 10038
 
+# Monthly CRR Auction Results (bundled ZIP containing bids/offers, base loading,
+# binding constraints, market results, and source/sink shadow prices CSVs)
+# https://www.ercot.com/mp/data-products/data-product-details?id=NP7-803-M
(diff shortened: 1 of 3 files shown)


gridstatus/tests/source_specific/test_ercot.py
WillKoehrsen, line 1: None of these tests use live source data. If the source data drifts as frequently happens, these tests will continue to pass.

gridstatus/ercot.py
```diff
@@ -6319,6 +6324,359 @@ def _process_cop_adjustment_period_snapshot_60_day_data(
...
+            date=date,
+            end=end,
+            verbose=verbose,
+        )
+
+        all_df = []
+        for month_start, raw in frames:
+            df = raw.copy()
+            df["Interval Start"] = month_start
+            df["Interval End"] = month_start + pd.DateOffset(months=1)
```
WillKoehrsen: Interval End in the dataframes does not match the EndDate in the files. We should use the dates provided in the source data.

Image
Image

WillKoehrsen: We've also been moving away from using timestamps when we don't need the precision.

gridstatus/tests/source_specific/test_ercot.py
Kladar: Ah yeah just caught this as well, it is because many of the files can be quite large (40+ MB), so it decided against cassettes. I am trying out cassettes implementation now.

Kladar: Okay, shifted back to a VCR framework, passing for me. I updated the command in the description as well. My `add-dataset` skill did allow for judgement calls in using the VCR approach so I've tightened up that language.

gridstatus/ercot.py
Kladar: Okay, I've switched it to pull the `endDate` from the 2 datasets that have explicit end dates and reduced the precision down to `date` from `Timestamp` across the datasets.

WillKoehrsen (approved):

gridstatus/tests/vcr_utils.py
```diff
@@ -67,6 +67,28 @@ def clean_cassettes(path: str):
     os.makedirs(path, exist_ok=True)
 
 
+def _strip_cache_buster_params(uri: str) -> str:
```
WillKoehrsen, line 70: Nice, good find.

gridstatus/tests/vcr_utils.py
```diff
@@ -67,6 +67,28 @@ def clean_cassettes(path: str):
...
+    """Remove cache-buster query params (keys starting with ``_``) from a URI.
+
+    ERCOT's MIS listing endpoint appends a ``&_<epoch>`` query param on every
```
WillKoehrsen, line 73: This should also work with the CAISO cache buster we use.

https://github.com/gridstatus/gridstatus/blob/93c055532756b778cae6eafccd9a10f7098caa7f/gridstatus/caiso/caiso.py#L95-L103

PR #530, opened 2025-01-23
upgrade lxml to 5.3.0

tylernisonoff:
## Summary

Upgrades lxml dependency from 4.x.x to 5.x.x

### Details

Not sure if you have any reason for not upgrading this, but, for a variety of reasons, we have to upgrade to 5.x.x, and this conflicts with gridstatus as a sub-dependency. 

I ran as many tests as I could locally to confirm this doesn't break anything, but couldn't run them all. As such, I figured I'd just open a PR and let CI run tests while discussing with the team here to determine if this could be accepted.

PR #862, opened 2026-04-24
Relax lxml pin to allow lxml 6.x

jhovell:
## Summary

Change `lxml ~=5.3` → `lxml >=5.3,<7` in `pyproject.toml` so downstream projects on lxml 6.x can install `gridstatus` from pypi.

## Motivation

We need lxml 6.x in our project for unrelated reasons, and we've been pinning `gridstatus` to a git commit on `main` (just before [#792](https://github.com/gridstatus/gridstatus/pull/792) landed) to work around the `~=5.3` constraint. That commit has been running in our production image against lxml 6 for roughly three months without issue.

The only lxml usage in gridstatus is `from lxml import etree as lxml_etree` in `gridstatus/ieso.py`. The `etree` Python-level API is stable across the lxml 5 → 6 transition, so the functional risk of accepting 6.x is minimal.

## History

- Jan 2025: @tylernisonoff (my colleague) opened [#530](https://github.com/gridstatus/gridstatus/pull/530) to upgrade lxml from 4.x → 5.x; shipped in 0.29.1. Every release from 0.29.1 through 0.34.0 used `lxml >= 5.3.0` — open-ended upper bound, which accepted lxml 6.x fine.
- Feb 2026: [#792 "Fix Version Specifications"](https://github.com/gridstatus/gridstatus/pull/792) retightened every dependency with `~=` operators as part of a hygiene pass. This changed `lxml >= 5.3.0` → `lxml ~=5.3` (equivalent to `>=5.3,<6`), first shipped in 0.35.0.
- Consequence: no released version since 0.35.0 installs with lxml 6.x.

This PR re-opens the upper bound but caps at `<7` so a future major lxml release would still have to be explicitly accepted.

## Test plan

Happy to run your test suite against lxml 6 locally and post results — just let me know if that's helpful. Alternatively CI should cover most of it.

Opening as draft to let CI run and gather any maintainer feedback before requesting review.


WillKoehrsen:
Thanks for the PR @jhovell. Do you mind running `uv.sync` and committing the changes to `uv.lock` (if any) in this PR?

jhovell:
@WillKoehrsen good catch — done. My initial pass used an older uv (0.7.12) which produced a noisy diff that included a `revision = 3 → 2` downgrade and re-added some `manylinux2014_s390x` wheel rows that newer uv had pruned. Upgraded to uv 0.11.7 and re-ran `uv sync` for a clean one-line lockfile change (just the `requires-dist` mirror of the pyproject `lxml` constraint update).

WillKoehrsen (approved):

PR #866, opened 2026-05-01
feat(ENG 3787): Add CAISO EDAM Wind and Solar Forecast scraper

Kladar:
## Summary
- Adds `CAISO.get_edam_wind_solar_forecast`, a scraper for the new CAISO OASIS `EDAM_WND_SLR_FORECAST_GRP` GroupZip report (day-ahead, hourly wind and solar forecasts by BAA for balancing areas participating in the extended day-ahead market).
- Registers `edam_wind_solar_forecast` in `OASIS_DATASET_CONFIG` with the `GroupZip` path, `version=1`, `resultformat=6`, and `groupid=EDAM_WND_SLR_FORECAST_GRP`.
- VCR-fixturized tests covering `today` and a recorded historical date (May 1, 2026 — the first market day with available data).

Output columns: `Interval Start`, `Interval End`, `Publish Time`, `BAA`, `Solar`, `Wind`.

I add publish time with our existing `_add_forecast_publish_time` helper method based on the datetime given in the one file available, since it isn't provided with the CSV result. Worth a look.
CleanShot 2026-05-01 at 14 00 47@2x

## Test plan
- [x] `uv run pytest gridstatus/tests/source_specific/test_caiso.py -k edam_wind_solar_forecast`

```

iso = CAISO()

df = iso.get_edam_wind_solar_forecast(date="2026-05-01", end="2026-05-02")
print(df.head())
print(df.dtypes)
```

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -1331,6 +1331,69 @@ class CAISO(ISOBase):
             publish_time_offset=pd.Timedelta(minutes=22.5),
         )
 
+    def get_edam_wind_solar_forecast(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        params: dict | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """Day-ahead, hourly, BAA-level wind and solar forecasts for balancing
+        areas participating in the extended day-ahead market (EDAM).
+
+        Data at: http://oasis.caiso.com/mrioasis/logon.do at System Demand >
+        Wind and Solar Forecast (EDAM).
+
+        Args:
+            date (str | pd.Timestamp): date to return data
+            end (str | pd.Timestamp | None, optional): last date of range to return data.
+            params (dict | None, optional): overrides for OASIS query parameters
+                (e.g. ``{"baa_grp_id": None}`` omits ``baa_grp_id`` from the URL).
+            verbose (bool, optional): print out url being fetched. Defaults to False.
+
+        Returns:
+            pandas.DataFrame: hourly EDAM wind and solar forecasts by BAA
+        """
+        data = self.get_oasis_dataset(
+            dataset="edam_wind_solar_forecast",
+            date=date,
+            end=end,
+            params=params,
+            verbose=verbose,
+            raw_data=False,
+        )
+
+        df = data.rename(
+            columns={
+                "BAA_GRP_ID": "BAA",
+                "SOLAR": "Solar",
+                "WIND": "Wind",
+            },
+        )
+
+        df["Solar"] = pd.to_numeric(df["Solar"], errors="coerce")
+        df["Wind"] = pd.to_numeric(df["Wind"], errors="coerce")
+
+        df["Publish Date"] = (
+            df["Interval Start"].dt.tz_convert(self.default_timezone).dt.normalize()
+            - pd.Timedelta(days=1)
+        ).dt.tz_convert("UTC")
+
+        return (
+            df[
+                [
+                    "Interval Start",
+                    "Interval End",
+                    "Publish Date",
+                    "BAA",
+                    "Solar",
+                    "Wind",
+                ]
+            ]
+            .sort_values(["Interval Start", "BAA"])
+            .reset_index(drop=True)
+        )
+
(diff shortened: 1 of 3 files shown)


gridstatus/caiso/caiso.py
```diff
@@ -1331,6 +1331,74 @@ def get_renewables_forecast_rtpd(
...
+        df["Solar"] = pd.to_numeric(df["Solar"], errors="coerce")
+        df["Wind"] = pd.to_numeric(df["Wind"], errors="coerce")
+
+        df = self._add_forecast_publish_time(
+            df,
+            current_time=current_time,
+            publish_time_offset_from_day_start=pd.Timedelta(
+                hours=11,
+                minutes=5,
+            ),  # NOTE: only one day of data to go off.  We can adjust
```
Shiggiddie: Do we have any updated insight for this data now?

Whether or not, we should probably drop this comment?

Shiggiddie (approved):
Nit to clean up a comment, otherwise looking good 👍

gridstatus/caiso/caiso.py
```diff
@@ -1331,6 +1331,74 @@ def get_renewables_forecast_rtpd(
...
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        params: dict | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """Day-ahead, hourly, BAA-level wind and solar forecasts for balancing
+        areas participating in the extended day-ahead market (EDAM).
+
+        Data at: http://oasis.caiso.com/mrioasis/logon.do at System Demand >
+        Wind and Solar Forecast (EDAM).
```
WillKoehrsen: Am I missing something here? I don't see an EDAM option under System Demand > Wind and Solar Forecast.

Image

gridstatus/caiso/caiso.py
```diff
@@ -1331,6 +1331,74 @@ def get_renewables_forecast_rtpd(
...
+            },
+        )
+
+        df["Solar"] = pd.to_numeric(df["Solar"], errors="coerce")
+        df["Wind"] = pd.to_numeric(df["Wind"], errors="coerce")
+
+        df = self._add_forecast_publish_time(
+            df,
+            current_time=current_time,
+            publish_time_offset_from_day_start=pd.Timedelta(
```
WillKoehrsen: The EDAM solar and wind forecast is supposed to be published every 30 minutes between 6 am and 10 am.

ATLAS REFERENCE > Publications > OASIS Publications Schedule

Image

gridstatus/caiso/caiso.py
Kladar: Yeah hidden in Energy > EDAM menu
CleanShot 2026-05-04 at 09 21 39@2x

WillKoehrsen: Okay, so the doc string is incorrect.

Kladar: oh just noticed that's what you were commenting on. fixed

gridstatus/caiso/caiso.py
Kladar: updated, and working on updating the publish time column now.

gridstatus/caiso/caiso.py
WillKoehrsen: However, every file I've downloaded has the `<TimeDate>2026-05-04T16:05:30-00:00</TimeDate>`. If that is a UTC time, it corresponds to 9:05 US/Pacific.

Kladar: working on this because the xml file seems to have publish time in it where the CSV does not, but doesn't seem like there are multiple files for a day. 
CleanShot 2026-05-04 at 09 40 37


also working on a @support_date_range decorator that gets the right data

Kladar: it seems the xml is the only place where the publish time is available, so I think it makes sense to ditch the get_oasis path in favor of going directly for the xml

Kladar: okay, I've gone the xml route and skip `_get_oasis`, but match the `_get_oasis()` retry pattern. I think in theory this should work even if they start publishing according to the pattern in the docs (and also works in the current 1-file-per-day pattern)

WillKoehrsen (approved):
Tests pass locally and data looks good.

PR #869, opened 2026-05-07
fix(ENG 3843): SPP RT binding constraint timeliness

Kladar:
## Summary
The SPP Binding Constraints Real Time 5 Min has 2 codepaths that can either fetch daily files or interval files. We had some logic to figure out which to use based on the query, but this was not doing the correct switching such that daily files were pulled more often than they should have been (the daily files only have data once the day is complete, while the interval files have up-to-date data but are very slow if used to get more than a few intervals). 

Updates the method to do interval data pulling more often with a wider window and some helper functions to figure out the case.

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -156,6 +156,27 @@ def fill_baa_column(df, load_col):
     return df
 
 
+def _binding_constraints_real_time_5_min_frequency(args_dict: dict) -> str:
+    """support_date_range frequency for SPP real-time binding constraints.
+
+    Splits at 5-min boundaries only for same-day recent windows so the
+    caller's explicit end is preserved through decorator chunking. Otherwise
+    splits per day, matching the source's daily-file granularity.
+    """
+    iso = args_dict["self"]
+    end = args_dict.get("end")
+    if end is None:
+        return "DAY_START"
+
+    start = utils._handle_date(args_dict["date"], iso.default_timezone)
+    end = utils._handle_date(end, iso.default_timezone)
+    if start.normalize() != end.normalize():
+        return "DAY_START"
+    if not utils.is_within_last_days(start, 1, iso.default_timezone):
+        return "DAY_START"
+    return "5_MIN"
+
+
 class SPP(ISOBase):
     """Southwest Power Pool (SPP)"""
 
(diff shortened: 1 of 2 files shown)


gridstatus/spp.py
```diff
@@ -2555,36 +2581,55 @@ def get_binding_constraints_real_time_5_min(
...
+        )
+
+    def _resolve_recent_interval_window(
+        self,
+        start: pd.Timestamp,
+        end: pd.Timestamp | None,
+    ) -> tuple[pd.Timestamp, pd.Timestamp]:
+        """Window for today/yesterday interval fetches.
+
+        Today's first ~2.25 hours live in yesterday's daily file, so start
```
WillKoehrsen: Looking at the files for today and yesterday, they start at the beginning of the day.

Image
Image

gridstatus/spp.py
```diff
@@ -2531,7 +2552,12 @@ def _process_binding_constraints_day_ahead_hourly(self, url: str) -> pd.DataFram
 
         return df[cols_to_keep].sort_values(["Interval Start", "Constraint Name"])
 
-    @support_date_range("DAY_START")
+    # NB: SPP RTBM publishes per-interval CSVs and daily aggregates. Daily files
+    # are faster (one request per day) but lag publication, so we use interval
+    # files for today/yesterday and daily files for older history. Today's first
+    # ~2.25 hours don't have interval files yet (they live in yesterday's daily
+    # file), so today's interval start is floored to 02:10 local.
+    @support_date_range(_binding_constraints_real_time_5_min_frequency)
```
WillKoehrsen: Seeing some strange behavior when calling the function: the same file is fetched 12 times resulting in the same data duplicated 12 times.

Image

gridstatus/spp.py
Kladar: Hmm, yeah I think they made some changes to this since I implemented. They now start at the start of the day and the interval files are only available for the current month (the previous month interval files are gone), which I don't think was the case originally.

CleanShot 2026-05-07 at 14 20 45@2x

I've updated this now

gridstatus/spp.py
Kladar: yeah actually you can see here the 2:15am start time of the files. I also get this when I try to query the current day 
CleanShot 2026-05-07 at 14 43 25

There is something funky going on since it shouldn't be requesting 262 files, but in the case above I think it is a 2:15am start time to the day that we are running into that isn't present in the files themselves but is present in the CSV API. Might be why originally I had it there, now that I'm starting to remember.

WillKoehrsen: That screenshot was using the old code. The new code correctly handles the query.

Image

Kladar: Actually, okay, the files are there for pre-2:15 it seems. Let me fix all this

Kladar: Yeah I was running old code on accident as well woops. Okay, I _think_ this is working now, just pushed the latest

WillKoehrsen (approved):
Data coming out of the function looks good now.

PR #870, opened 2026-05-11
feat(ENG-2896): Add ERCOT Annual CRR Auction scrapers (5 datasets)

Kladar:
## Summary
- Adds five `get_crr_*_annual` scrapers for ERCOT's long-term CRR auction (NP7-802-M, report_type_id 11203), mirroring the existing monthly CRR set.

### Details
Each annual auction publishes one ZIP per (auction year, strip, sequence) clearing event: 6 sequences (capacity bands 70/55/40/30/20/10%) by 2 strips (1st / 2nd) = 12 ZIPs per auction year. The helper functions parse the friendly name to derive auction year, strip (1 or 2), and sequence (1-6), iterate over every matching CSV inside each ZIP, and stamp `Sequence` and `Strip` columns onto every returned row so downstream consumers can disambiguate the otherwise-overlapping clearings. CSV prefixes inside the ZIP match the monthly bundle (`Common_AuctionBidsAndOffers_`, `Common_BaseLoading_`, etc.). 

The Base Loading dataset additionally splits CSVs by covered month — for that dataset `Interval Start`/`Interval End` come from the trailing `_AUCTION_<MMM>_<YYYY>.csv` segment of each filename rather than the friendly name. Bids/offers and market results keep their per-row `StartDate`/`EndDate` semantics; binding constraints and source/sink shadow prices keep their per-row `CalendarPeriod` (`MMM_YYYY`) semantics. The methods are decorated with `@support_date_range(frequency="YEAR_START")` and filter the MIS document list by auction year.

To run tests:

```bash
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "ercot" -k "crr and annual"
```

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -6863,11 +6863,11 @@ class Ercot(ISOBase):
         all_df = []
         for _year_start, strip, sequence, _basename, raw in frames:
             df = raw.copy()
-            df["Interval Start"] = pd.to_datetime(
+            df["Start Date"] = pd.to_datetime(
                 df["StartDate"],
                 format="%m/%d/%Y",
             ).dt.tz_localize(self.default_timezone)
-            df["Interval End"] = pd.to_datetime(
+            df["End Date"] = pd.to_datetime(
                 df["EndDate"],
                 format="%m/%d/%Y",
             ).dt.tz_localize(self.default_timezone)
@@ -6888,8 +6888,8 @@ class Ercot(ISOBase):
         df = pd.concat(all_df, ignore_index=True)
         return df[
             [
-                "Interval Start",
-                "Interval End",
+                "Start Date",
+                "End Date",
                 "Path",
                 "Source",
                 "Sink",
@@ -6914,7 +6914,7 @@ class Ercot(ISOBase):
         """Base loading for each annual (long-term) ERCOT CRR Auction.
 
         Each auction zip publishes one base-loading CSV per covered month;
-        ``Interval Start`` / ``Interval End`` are derived from the trailing
+        ``Start Date`` / ``End Date`` are derived from the trailing
         ``_AUCTION_<MMM>_<YYYY>.csv`` segment of the file name.
 
         Source: https://www.ercot.com/mp/data-products/data-product-details?id=NP7-802-M
@@ -6938,8 +6938,8 @@ class Ercot(ISOBase):
                 datetime.datetime.strptime(month_year, "%b_%Y"),
             ).tz_localize(self.default_timezone)
             df = raw.copy()
-            df["Interval Start"] = month_start
-            df["Interval End"] = month_start + pd.offsets.MonthEnd(0)
+            df["Start Date"] = month_start
+            df["End Date"] = month_start + pd.offsets.MonthEnd(0)
             df = df.rename(
                 columns={
                     "CRR_ID": "CRR ID",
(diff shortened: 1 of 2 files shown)


gridstatus/ercot.py
```diff
@@ -6707,6 +6730,436 @@ def _process_cop_adjustment_period_snapshot_60_day_data(
...
+            "auction_bids_offers",
+            date=date,
+            end=end,
+            verbose=verbose,
+        )
+
+        all_df = []
+        for _year_start, strip, sequence, _basename, raw in frames:
+            df = raw.copy()
+            df["Interval Start"] = pd.to_datetime(
```
Kladar: Going to switch these to `Start Date` and `End Date` to avoid renaming downstream, one sec...

Kladar: Done

gridstatus/ercot.py
```diff
@@ -6707,6 +6730,436 @@ def _process_cop_adjustment_period_snapshot_60_day_data(
...
+        if not results:
+            raise NoDataFoundException(
+                "No Annual CRR Auction documents found for "
+                f"dataset={dataset_key}, date={date}, end={end}",
+            )
+
+        results.sort(key=lambda item: (item[0], item[1], item[2], item[3]))
+        return results
+
+    @support_date_range(frequency="YEAR_START")
```
WillKoehrsen: From what I can tell, files are released monthly. Are we assuming that if a user requests a value mid-year, such as 2025-06-01, they actually want values for all of 2025?

Image

Image

Kladar: Good question, 
_TL;DR we could do some more filtering to give the user a slightly more accurate request (but I do think we have to get all the yearly files regardless of request and stick with a `YEAR_START` date range support)._ 

 I'm not quite sure on what the right answer is in what data to return based on a given date (though I think we can do slightly better than the current approach). The files are published monthly, but they are typically for future auction years, which have the Strip and Sequence delineation, and the actual datasets themselves will have a half year in a single file. For instance, the `20271st6AnnualAuctionSeq3CRRAuctionResults` file, published `1/2/2026`, in the bids/offers file, has data for January to June, 2027. 
CleanShot 2026-05-11 at 14 18 32@2x
 
As does the `20271st6AnnualAuctionSeq4CRRAuctionResults` file, published `7/31/2025`. So right now we get all the zip files for a given auction year (12, 2 strips and 6 sequences), and then just concat them together. 

So I think we can 
1. just get the right strip for a requested date range, like avoid 6 of 12 downloads if the date is in January, just get strip 1 files for the given year rather than both strips
2. do some row level filtering for just the date range requested once all the data is concatenated

So I'll try those out, but I think it's mostly slightly more selective file getting and row-based filtering. We do have to get multiple files even if a single date is requested, and I don't think we can go to a `MONTH_START` or similar decorator. (It might make sense to add a publish time column, but we didn't in monthly and I don't think that's as necessary.).

Kladar: Actually yeah with just optimizing grabbing the correct strip for the request (strip 1 if Jan-June, strip 2 if july-dec), we do a `frequency=None`, but main points hold.

WillKoehrsen: Alright, we like to default to returning more data than requested so I'm fine with the current implementation.

WillKoehrsen (approved):

PR #880, opened 2026-05-20
[ENG-3798] Add get_marginal_fuel_type to ISONEAPI

fjlanasa:
## Summary
Adds `ISONEAPI.get_marginal_fuel_type` parallel to `get_fuel_mix`. Uses the same `/genfuelmix/{current,day/{date}}` endpoints, but pivots the `MarginalFlag` field into one nullable boolean column per fuel category instead of MW values.

## Design
- One row per API tick (irregular cadence, ~120–360 per day).
- Fuel columns are pandas nullable `boolean` (TRUE / FALSE / NA).
- No `fillna` — when a tick omits a fuel's record, that cell stays NA. Distinguishes "fuel not reported" from "explicitly not marginal."
- Cross-day concat may produce object-dtype columns with NA where a fuel category is absent on some days (e.g. Coal post-2024); preserved as NA rather than synthesized.

## Tests
- `test_get_marginal_fuel_type_latest` (1-row "latest" call, strict `dtype == bool`).
- `test_get_marginal_fuel_type_date_range` parameterized on the existing `DST_CHANGE_TEST_DATES`; allows object dtype with bool/NA values to accommodate cross-day concat.
- Cassettes recorded locally; covered by the existing gitignored `**/vcr_cassettes/` pattern.

## Test plan
- [ ] CI green on the new tests
- [ ] Spot-check that `dtype == bool` for single-day calls
- [ ] Spot-check that `MarginalFlag == 'Y'` rows in the raw API become `True` in the pivot

### DIFF
--- a/gridstatus/isone_api/isone_api.py
+++ b/gridstatus/isone_api/isone_api.py
@@ -280,6 +280,52 @@ class ISONEAPI:
         return mix_df
 
     @support_date_range("DAY_START")
+    def get_marginal_fuel_type(
+        self,
+        date: str | pd.Timestamp = "latest",
+        end: str | pd.Timestamp | None = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """Return marginal-fuel flags per timestamp, one boolean column per fuel type.
+
+        Args:
+            date (str | pd.Timestamp): The start date for the data request. Use "latest" for most recent data.
+            end (str | pd.Timestamp | None): The end date for the data request. Only used if date is not "latest".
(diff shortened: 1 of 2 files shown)


Kladar (approved):
Nice, thanks for this. I caught up on the slack thread that pertains to the decisions here, so looks good in that regard, even if this one is kind of funky. 

Since many of our source-specific tests fail in CI, helpful to add a test command to the PR description to run just the new tests. In this case something like:
```
VCR_RECORD_MODE="all" pytest -vv gridstatus/tests/source_specific/test_isone_api.py -k "marginal"
```

PR #882, opened 2026-05-20
fix(ENG 3910): Reduce PJM unverified lmp queries against dataminer

Kladar:
## Summary
In getting PJM LMPs, we make calls to `iso.get_lmp(market=REAL_TIME_5_MIN, locations="ALL")`. Inside `get_lmp`, the verified endpoint `rt_fivemin_hrl_lmps` is queried first; on `NoDataFoundException`, the code falls back to `rt_unverified_fivemin_lmps`. These queries hit the API in a nonperformant way, and is requested that we fix it. 

We want to avoid 
1. Re-querying the same immutable historical 5-minute data, which is done via the 5-minute throttle blocking subsequent calls until at least 5 min have elapsed
2. Querying multi-hour windows at a time, which is done with `5MinutesAgo` API keyword, is by definition exactly one 5-minute interval
3. Querying on timestamps that are not on the 5-minute mark, also via `5MinutesAgo` as a server-side keyword that PJM snaps for us

### Details
This setup makes it such that we cannot query the unverified LMP endpoint via the fallback more than once per 5 minutes. Here's how it would go:

Time | Verified LMP API call | Unverified LMP API call | Outcome
-- | -- | -- | --
t=0s | empty | `throttle: last=None` → fires, returns 5MinutesAgo (no new data either) | 1 unverified hit
t=5s | empty | `throttle: last=0s`, 5s ago < 5min → blocked, raises | 0 unverified hits
t=10s | empty | blocked | 0
… (every 5s) … | empty | blocked | 0
t=300s | empty | throttle: 5min elapsed → fires | 1 unverified hit

Lastly, this adds a `latest_only` kwarg for the the `_get_pjm_json()` method, such that would could conceivably apply this to other datasets if they are also problematic. 

```
VCR_RECORD_MODE="all" pytest -vv gridstatus/tests/source_specific/test_pjm.py::TestPJMUnverifiedFiveminFallback
```


gridstatus/pjm.py
```diff
@@ -541,13 +546,31 @@ def get_lmp(
...
+                if (
+                    last_call is not None
+                    and (now_utc - last_call) < self.UNVERIFIED_FIVEMIN_LMPS_THROTTLE
+                ):
+                    raise NoDataFoundException(
+                        f"No data found for {market_endpoint}: throttled "
+                        f"(last call {now_utc - last_call} ago)",
+                    )
+                self._last_unverified_fivemin_lmps_call_utc = now_utc
+                latest_only = True
```
Shiggiddie: We're unconditionally setting `latest_only = True`

But what if we have a worker coming back online after downtime and we need to catch-up on several possible intervals.  Won't this only result in a single `5MinutesAgo` and drop the rest?

gridstatus/pjm.py
```diff
@@ -92,6 +94,8 @@ def __init__(
         if not self.api_key:
             raise ValueError("api_key must be provided or set in PJM_API_KEY env var")
 
+        self._last_unverified_fivemin_lmps_call_utc: pd.Timestamp | None = None
```
Shiggiddie: Is this the right level to throttle, at the instance-global level?

I'm thinking if this were to run outside of a single-dataset worker, what might happen?

Kladar: I've swapped this to be at the class level such that multiple instances of `PJM` would still get the throttle, and verified 5 min lmps is the only path right now that would be able to access the throttle.

gridstatus/pjm.py
Kladar: Yeah good point, the verified tends to lag behind the unverified by a few days, so I've added a bit more lookback logic to handle the two cases:

- Live window (<= `UNVERIFIED_FIVEMIN_LMPS_LIVE_WINDOW`, i.e. at most one
          5-min interval): send `datetime_beginning_ept=5MinutesAgo` and
          throttle to at most one call per `UNVERIFIED_FIVEMIN_LMPS_THROTTLE` so
          a tight worker loop can't hammer the endpoint at the tip of the data.

- Catch-up window (longer): snap start/end to the 5-minute mark and
          page through in `UNVERIFIED_FIVEMIN_LMPS_CHUNK`-sized windows so the
          worker can backfill missed intervals after downtime without sending
          multi-hour requests Dataminer rejects, and without dropping any of
          the multiple intervals the wider window asked for.

Kladar: Confirming this does the unverified hourly chunk catch up when lagged (this is running on the testing DB, so it was by definition behind)

```
tartRow': 1, 'rowCount': 50000, 'datetime_beginning_ept': '05/22/2026 06:25to05/22/2026 07:25'}
2026-05-22 12:51:34 - INFO - Requesting https://api.pjm.com/api/v1/rt_unverified_fivemin_lmps with {'params': {'fields': 'congestion_price_rt,datetime_beginning_ept,datetime_beginning_utc,marginal_loss_price_rt,occ_check,pnode_id,pnode_name,ref_caseid_used_multi_interval,total_lmp_rt,type', 'startRow': 1, 'rowCount': 50000, 'datetime_beginning_ept': '05/22/2026 06:25to05/22/2026 07:25'}, 'headers': {'Ocp-Apim-Subscription-Key': '647ac3b441824215b252980cf004d42a'}}
```

Kladar:
Added some debug logging for knowing which code path we're on in scraper production, seems to be working. 
CleanShot 2026-05-22 at 13 14 49

gridstatus/pjm.py
```diff
@@ -500,46 +506,47 @@ def get_lmp(
...
-
-        # returns on the latest version of the data
-        params["row_is_current"] = "TRUE"
-
-        try:
             data = self._get_pjm_json(
                 market_endpoint,
-                start=date,
-                end=end,
+                start=pd.Timestamp.utcnow(),
```
kmax12: do we need to pass start/end if it is `latest_only`? maybe there is no cleaner way though

Kladar: I think it's easier to just add a `get_lmp_real_time_unverified_5_min` method. There are a lot of edge cases:
1. unverified live (with throttle)
2. unverified catch-up (if scraper is down)
3. verified

I'm finding it spaghettifies quickly if we don't build with those edge cases in mind up front.

gridstatus/pjm.py
```diff
@@ -923,7 +1024,10 @@ def _get_pjm_json(
         final_params = params.copy()
         final_params.update(default_params)
 
-        if start is not None:
+        if latest_only:
+            final_params[f"{filter_timestamp_name}_ept"] = "5MinutesAgo"
```
kmax12: is this correct to had here? this the `get_lmp_real_time_unverified_hourly` method?

gridstatus/pjm.py
```diff
@@ -459,7 +459,13 @@ def get_lmp(
         params = {}
 
         if market == Markets.REAL_TIME_5_MIN:
-            market_endpoint = "rt_fivemin_hrl_lmps"
+            # `date="latest"` hits the unverified endpoint directly with
+            # `5MinutesAgo`; the verified endpoint hasn't published the tip yet
+            # by definition. Any non-latest date uses the verified endpoint.
+            if date == "latest":
```
kmax12: do we need to support this here? i don't think we use it ourselves. might be simpler to remove for now because I think we can also revert the other changes to this function then

Kladar: yep agreed, yeah will remove.

gridstatus/pjm.py
kmax12: i buy that. i'd love to split up all these `get_lmp` methods :)

gridstatus/pjm.py
```diff
@@ -619,6 +620,105 @@ def _add_pnode_info_to_lmp_data(self, data: pd.DataFrame) -> pd.DataFrame:
 
         return data
 
+    @support_date_range(frequency="365D")
+    def get_lmp_real_time_unverified_5_min(
+        self,
+        date: str | pd.Timestamp,
+        end: str | pd.Timestamp | None = None,
+        locations: str | list | None = "hubs",
```
kmax12, line 627: this is just a comment for future clean up, but i think we can remove all suport for locations/location_type since we don't use them ourselves

gridstatus/tests/source_specific/test_pjm.py
```diff
@@ -3347,3 +3347,68 @@ def test_get_marginal_emission_rates_5_min_historical_date_range(
             self._check_marginal_emission_rates_5_min(df)
             assert df["Interval Start"].min() >= date
             assert df["Interval End"].max() <= end
+
+    """get_ftr_option_paths_monthly"""
+
+    def _check_ftr_option_paths_monthly(self, df):
```
kmax12: is this unintentional?

kmax12 (approved):
LGTM

gridstatus/tests/source_specific/test_pjm.py
Kladar: removed

PR #884, opened 2026-05-29
feat(ENG 3945): Add ERCOT Available Resource Planned Outage Capacity scrapers

Kladar:
## Summary
- Adds `get_planned_outage_capacity_7_day` scraper for ERCOT (NP3-162-CD)
- Adds `get_planned_outage_capacity_future` scraper for ERCOT (NP3-161-CD)

### Details
These two reports come from the ERCOT EMIL MIS archive and share the same six capacity columns (Max/Approved/Received POC Non IRR Non PUN and Max/Approved/Received POC IRR), so they share a column-rename map and a common dataframe handler. `get_planned_outage_capacity_7_day` is the hourly report giving maximum resource capacity available for planned outages within 7 days of the operating day; it has both OperatingDate and HourEnding plus a DSTFlag, so it reuses `parse_doc` (after renaming OperatingDate to DeliveryDate) to build hourly `Interval Start`/`Interval End`. `get_planned_outage_capacity_future` is the daily report that begins 7 days out and extends up to 60 months ahead; it has only an OperatingDate with no hour column, so intervals are built directly as day-aligned `Interval Start`/`Interval End` (which correctly span 23h/24h/25h across DST transitions rather than a fixed 24h). Both methods accept the report publish time as the `date` argument and support a publish-time range via `end`, fetching every report published in that window. The `Received` columns are spelled correctly here; note the Notion ticket lists the Non IRR Non PUN one as "Receieved".

To run tests:
```bash
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "ercot" -k "planned_outage_capacity"
```

### DIFF
--- a/gridstatus/ercot.py
+++ b/gridstatus/ercot.py
@@ -4233,13 +4233,17 @@ class Ercot(ISOBase):
         Returns:
             pandas.DataFrame: A DataFrame with hourly planned outage capacity data
         """
-        return self._get_hourly_report(
-            start=date,
-            end=end,
-            report_type_id=PLANNED_OUTAGE_CAPACITY_7_DAY_RTID,
-            extension="csv",
-            handle_doc=self._handle_planned_outage_capacity_7_day,
-            verbose=verbose,
+        return (
+            self._get_hourly_report(
+                start=date,
+                end=end,
+                report_type_id=PLANNED_OUTAGE_CAPACITY_7_DAY_RTID,
+                extension="csv",
+                handle_doc=self._handle_planned_outage_capacity_7_day,
+                verbose=verbose,
+            )
+            .sort_values(["Interval Start", "Publish Time"])
+            .reset_index(drop=True)
         )
 
     def _handle_planned_outage_capacity_7_day(
@@ -4281,13 +4285,17 @@ class Ercot(ISOBase):
         Returns:
             pandas.DataFrame: A DataFrame with daily planned outage capacity data
         """
-        return self._get_hourly_report(
-            start=date,
-            end=end,
-            report_type_id=PLANNED_OUTAGE_CAPACITY_FUTURE_RTID,
-            extension="csv",
-            handle_doc=self._handle_planned_outage_capacity_future,
-            verbose=verbose,
+        return (
+            self._get_hourly_report(
+                start=date,
+                end=end,
+                report_type_id=PLANNED_OUTAGE_CAPACITY_FUTURE_RTID,
+                extension="csv",
+                handle_doc=self._handle_planned_outage_capacity_future,
+                verbose=verbose,
+            )
+            .sort_values(["Interval Start", "Publish Time"])
+            .reset_index(drop=True)
         )
 
     def _handle_planned_outage_capacity_future(
@@ -4325,7 +4333,7 @@ class Ercot(ISOBase):
                 "AggReceivedResourcePOCIRR": "Received POC IRR",
             },
         )
-        df = df.sort_values(["Interval Start", "Publish Time"]).reset_index(drop=True)
+
         return df[
             [
                 "Interval Start",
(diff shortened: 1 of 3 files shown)


Kladar:
Here is an example tester script that pulls from source and outputs dfs and metadata. This is not normally something I would check in to this repo since they are ephemeral scripts. Easy to create and iterate on for what you care about whenever you need. 

```
import pandas as pd

from gridstatus import Ercot

iso = Ercot()

start = pd.Timestamp.now(tz=iso.default_timezone).normalize() - pd.Timedelta(days=2)
end = start + pd.Timedelta(days=1)

print("=== 7 day (NP3-162-CD) ===")
df_7_day = iso.get_planned_outage_capacity_7_day(date=start, end=end, verbose=True)
print(df_7_day)
print(df_7_day.dtypes)
print("rows:", len(df_7_day))
print("publish times:", df_7_day["Publish Time"].nunique())
print("dupes:", df_7_day.duplicated(subset=["Publish Time", "Interval Start"]).sum())

print("\n=== future (NP3-161-CD) ===")
df_future = iso.get_planned_outage_capacity_future(date=start, end=end, verbose=True)
print(df_future)
print(df_future.dtypes)
print("rows:", len(df_future))
print("publish times:", df_future["Publish Time"].nunique())
print("dupes:", df_future.duplicated(subset=["Publish Time", "Interval Start"]).sum())
print("interval span:", df_future["Interval Start"].min(), "->", df_future["Interval Start"].max())
```

gridstatus/tests/source_specific/test_ercot.py
```diff
@@ -1882,6 +1882,79 @@ def test_get_hourly_resource_outage_capacity(self):
...
+        "Approved POC IRR",
+        "Received POC IRR",
+    ]
+
+    def _check_planned_outage_capacity(self, df):
+        assert df.columns.tolist() == self.PLANNED_OUTAGE_CAPACITY_COLUMNS
+
+        assert df.dtypes["Interval Start"] == "datetime64[ns, US/Central]"
+        assert df.dtypes["Interval End"] == "datetime64[ns, US/Central]"
+        assert df.dtypes["Publish Time"].kind == "M"
```
fjlanasa: Should we assert the timezone here too?

gridstatus/ercot.py
```diff
@@ -4203,6 +4211,135 @@ def _handle_hourly_resource_outage_capacity_df(
...
+            columns={
+                "MaxDailyResourcePOCnonIRRnonPUN": "Max POC Non IRR Non PUN",
+                "AggApprovedResourcePOCnonIRRnonPUN": "Approved POC Non IRR Non PUN",
+                "AggReceivedResourcePOCnonIRRnonPUN": "Received POC Non IRR Non PUN",
+                "MaxDailyResourcePOCIRR": "Max POC IRR",
+                "AggApprovedResourcePOCIRR": "Approved POC IRR",
+                "AggReceivedResourcePOCIRR": "Received POC IRR",
+            },
+        )
+        df = df.sort_values(["Interval Start", "Publish Time"]).reset_index(drop=True)
```
fjlanasa: This dataframe ends up being resorted within `_get_hourly_report` using `sort_values('Published Time')`, so within a single 'Published Time', 'Interval Start' can be out of order. Does that matter? If so, maybe we should call this `df.sort_values(["Interval Start", "Publish Time"]).reset_index(drop=True)` inside of `get_planned_outage_capacity_future` and `get_planned_outage_capacity_7_day`.

gridstatus/tests/source_specific/test_ercot.py
Kladar: updated

gridstatus/ercot.py
Kladar: good catch, updated

fjlanasa (approved):
LGTM

PR #885, opened 2026-05-31
fix(ercot): handle 60-Day SCED Resource AS Offers column rename

WillKoehrsen:
## What changed
ERCOT renamed the AS-price column suffixes in the 60-Day SCED Resource AS Offers report in late March 2026 (`_URS/_DRS/_NS/_RRSPF/_RRSUF/_RRSFF` → `_REGUP/_REGDN/_NSPIN/_RRSPFR/_RRSUFR/_RRSFFR`). `process_sced_resource_as_offers` infers `Curve Type` from which suffix columns are non-zero, so on the renamed files every offer type collapsed to `Online`, producing duplicate `(SCED Timestamp, Resource Name, Curve Type)` primary keys.

Fix: rename the new suffixes back to the original names at the top of the function. The existing curve-type classification and curve-extraction logic is unchanged and now works for both old and new files. (The pre-rename inference was already correct — only the column names changed.)

Suffix mapping is evidenced by the raw headers (old vs new files are byte-identical except these 6 suffixes) and by ERCOT's RRS sub-type definitions (RRSPFR = Primary Frequency Response, RRSUFR = Under Frequency Relay, RRSFFR = Fast Frequency Response).

## How to validate
```
pytest gridstatus/tests/source_specific/test_ercot.py -k ScedResourceAsOffers
```
12 pass (9 existing old-suffix tests + 3 new-suffix tests: same curve types, no duplicate PKs, identical output to the old layout). Also verified against real disclosure files for OD 2026-01-10, 2026-02-15 (old suffixes) and 2026-03-31 (new suffixes): 0 duplicate PKs on all three.

### DIFF
--- a/gridstatus/ercot_60d_utils.py
+++ b/gridstatus/ercot_60d_utils.py
@@ -1479,6 +1479,28 @@ def process_sced_resource_as_offers(
             "pg_array_as_string" returns PG array strings like '{{mw,price},{mw,price}}'
             directly, using ~3x less peak memory.
     """
+    # ERCOT renamed the AS-price column suffixes in late March 2026
+    # (_URS->_REGUP, _DRS->_REGDN, _NS->_NSPIN, _RRSPF->_RRSPFR,
+    # _RRSUF->_RRSUFR, _RRSFF->_RRSFFR). Rename them back to the original names
+    # so the curve-type and curve-extraction logic below works unchanged for
+    # both old and new files.
+    new_to_old_suffix = {
+        "_REGUP": "_URS",
+        "_REGDN": "_DRS",
+        "_RRSPFR": "_RRSPF",
+        "_RRSUFR": "_RRSUF",
+        "_RRSFFR": "_RRSFF",
+        "_NSPIN": "_NS",
+    }
+
+    def _rename_suffix(col):
+        for new_suffix, old_suffix in new_to_old_suffix.items():
+            if col.endswith(new_suffix):
+                return col[: -len(new_suffix)] + old_suffix
+        return col
+
+    df = df.rename(columns=_rename_suffix)
+
     # First create a curve_type column with the logic:
     # regulation down : values only in _DRS columns and not in other columns
     # offline: values in _NS and optionally _ECRS columns and not in other columns
--- a/gridstatus/tests/source_specific/test_ercot.py
+++ b/gridstatus/tests/source_specific/test_ercot.py
@@ -4522,6 +4522,68 @@ class TestProcessScedResourceAsOffers:
         assert result.columns.tolist() == SCED_RESOURCE_AS_OFFERS_COLUMNS
 
 
+def _to_new_suffixes(df):
+    """Rename AS-price columns to ERCOT's post-March-2026 suffixes."""
+    old_to_new = {
+        "_URS": "_REGUP",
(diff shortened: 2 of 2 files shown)


gridstatus/ercot_60d_utils.py
```diff
@@ -1479,6 +1479,28 @@ def process_sced_resource_as_offers(
...
+    # so the curve-type and curve-extraction logic below works unchanged for
+    # both old and new files.
+    new_to_old_suffix = {
+        "_REGUP": "_URS",
+        "_REGDN": "_DRS",
+        "_RRSPFR": "_RRSPF",
+        "_RRSUFR": "_RRSUF",
+        "_RRSFFR": "_RRSFF",
+        "_NSPIN": "_NS",
+    }
```
WillKoehrsen, line 1494: @cwaldoch The columns for SCED Resource AS Offers recently changed from

```
"SCED Timestamp","Repeated Hour Flag","Resource Name","PRICE1_URS","PRICE1_DRS","PRICE1_RRSPF","PRICE1_RRSUF","PRICE1_RRSFF","PRICE1_NS","PRICE1_ECRS","QUANTITY_MW1",...
```

to 

```
"SCED Timestamp","Repeated Hour Flag","Resource Name","PRICE1_REGUP","PRICE1_REGDN","PRICE1_RRSPFR","PRICE1_RRSUFR","PRICE1_RRSFFR","PRICE1_NSPIN","PRICE1_ECRS",...
```

Can you confirm the mapping we have in this dictionary is correct?

WillKoehrsen: I looked for market notices regarding this change but wasn't able to find any.

cwaldoch: what the heck. . .  I'll review

looks right on first glance, but want to confirm `URS` and `DRS`

cwaldoch:
I still see REGUP and REGDN in a file from ERCOT? 
image

gridstatus/ercot_60d_utils.py
WillKoehrsen: `URS` and `DRS` were the old names.

Kladar: Confirming I see the change from `URS`/`DRS` to `REGUP`/`REGDN`
**Before**
CleanShot 2026-06-01 at 07 57 06@2x

**After**
CleanShot 2026-06-01 at 07 58 16

Kladar (approved):
Confirming I see the change from `URS`/`DRS` to `REGUP`/`REGDN`
**Before**
CleanShot 2026-06-01 at 07 57 06@2x

**After**
CleanShot 2026-06-01 at 07 58 16

Though I agree, probably worth Connor approval as well

WillKoehrsen:
> I still see REGUP and REGDN in a file from ERCOT?

@cwaldoch `REGUP` and `REGDN` are the new names. The old names were `URS` and `DRS`.

cwaldoch:
> > I still see REGUP and REGDN in a file from ERCOT?
> 
> @cwaldoch `REGUP` and `REGDN` are the new names. The old names were `URS` and `DRS`.

ah yeah I see that now, sorry, misunderstood originally

given their location in the file and such it seems like all of those assignments are correct

PR #886, opened 2026-06-03
feat(ENG 3533): Add MISO API Generation Fuel Mix by region scrapers

Kladar:
## Summary
- Adds `get_generation_fuel_mix_by_region_real_time` and `get_generation_fuel_mix_by_region_day_ahead` scrapers for MISO

I ended up splitting this into 2 datasets, even if it's a little strange to have a real time and day ahead fuel mix (these are estimates put out by MISO, and they are on a 1 day lag typically, such that they are published the day after the market interval date). 

### Details
These methods hit the MISO Generation Fuel Mix api endpoints, which contains hourly generation fuel mix broken out by region (Central, North, South) for both the real-time state estimator and the day-ahead cleared market. Each method returns a long-by-area frame with one row per `(Interval Start, Region)`, with `Region` taking the values Central, North, South, and derived Total (the system-wide total, which equals the sum of the three physical areas). Fuel columns are Coal, Gas, Nuclear, Hydro, Wind, Solar, Other, and Storage, plus a Total. The data is published the day after the market date. The South area does not report Storage, so that value is null for South. Timestamps are localized to MISO's fixed EST offset. 

To run tests: `VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "miso" -k "generation_fuel_mix_by_region"`


gridstatus/miso.py
```diff
@@ -853,6 +853,205 @@ def get_interconnection_queue(self, verbose: bool = False) -> pd.DataFrame:
...
+            "Gas",
+            "Nuclear",
+            "Hydro",
+            "Wind",
+            "Solar",
+            "Other",
+            "Storage",
+        ]
+
+        # Column offsets for each area block in the raw sheet. South does not report a
```
WillKoehrsen: These offsets seem a little brittle. Have we verified they work on historical files?

WillKoehrsen:
This dataset is going to be retired when MISO does the Market Reports retirement. 

[MISO Data Exchange Information718679.pdf](https://github.com/user-attachments/files/28725002/MISO.Data.Exchange.Information718679.pdf)
[Data Exchange Report to Endpoint Mapping726669.pdf](https://github.com/user-attachments/files/28725003/Data.Exchange.Report.to.Endpoint.Mapping726669.pdf)

Screenshot 2026-06-08 at 15 22 51

We have already implemented the API endpoints although we may need to add the `MISO` region by summing up the other regions.

https://github.com/gridstatus/gridstatus/blob/ae6b1606b38985b070f960cec387c38280a9ccd7/gridstatus/miso_api.py#L1133

https://github.com/gridstatus/gridstatus/blob/ae6b1606b38985b070f960cec387c38280a9ccd7/gridstatus/miso_api.py#L802

I'm wondering if the addition in this PR is necessary or if we can only update the `MISOAPI` methods.

gridstatus/miso.py
Kladar: Good call out, they do change slightly around 2023-09-07 with wind being added to the south area (data goes back to 2023-01-01). I've made the column discovery dynamic, and the backfill now works for the entire range of historical data. 
CleanShot 2026-06-08 at 14 08 32

Kladar:
Okay, this updates the MISOAPI methods to take over instead of adding the market reports. My experiments showed no latency between the API and the market reports for data, and the same historical date range available. One difference is that the API refers to these as "Regions" rather than "Areas", so going with that nomenclature. Also the API only returns the region data, so I create a MISO total column as well to match the market reports. 

Also adds the miso_api file to the mypy ignore list.

WillKoehrsen (approved):
Data looks good and tests pass locally. For posterity, could we update the PR title and description before merging?

PR #892, opened 2026-06-15
feat(ENG 4137/4140/4143): Add NYISO, SPP, and MISO dataset scrapers

Kladar:
## Summary
- Adds `get_btm_installed_capacity` for NYISO (ENG-4137)
- Adds `get_solar_and_wind_forecast_by_reserve_zone` for SPP (ENG-4140)
- Adds `get_load_forecast_mid_term_by_region` for MISO (ENG-4143)

### Details
**NYISO BTM Installed Capacity** pulls daily behind-the-meter installed capacity by zone from MIS P-70Clist (`btminstalledcapacitytracking`).

**SPP Solar and Wind Forecast by Reserve Zone** fetches hourly flat files from `resource-forecast-by-reserve-zone`. The `date` parameter is the publish hour (file timestamp). Publish time is derived from the filename. Each file contains ~24 hours of actuals (intervals before publish time) and ~7 days of forecasts. **NaN actuals are expected** for current and future intervals — actuals are only populated for intervals before publish time. Forecast columns are null only on the horizon edge (dropped when all value columns are null).

**MISO Load Forecast by Region** wraps the Data Exchange API with publish-date semantics: `date` is the forecast run init/publish date, and the method queries the 7-day forecast horizon for that vintage. Returns long-format rows by region and LRZ.

To run tests:
```bash
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "btm_installed_capacity"
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "solar_and_wind_forecast_by_reserve_zone"
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "load_forecast_mid_term_by_region"
```

### DIFF
--- a/gridstatus/miso_api.py
+++ b/gridstatus/miso_api.py
@@ -1524,7 +1524,9 @@ class MISOAPI:
                 "Local Resource Zone",
                 "init",
             ]:
-                data[col] = data[col].astype(float)
+                data[col] = (
+                    pd.to_numeric(data[col], errors="coerce").round().astype("Int64")
+                )
 
         data = data.sort_values(
             ["Interval Start", "Publish Time", "Region", "Local Resource Zone"],
@@ -1672,6 +1674,18 @@ class MISOAPI:
                 "LRZ8_9_10 MTLF",
             ]
         ].sum(axis=1)
+
+        mtlf_cols = [
+            "LRZ1 MTLF",
+            "LRZ2_7 MTLF",
+            "LRZ3_5 MTLF",
+            "LRZ4 MTLF",
+            "LRZ6 MTLF",
+            "LRZ8_9_10 MTLF",
+            "MISO MTLF",
+        ]
+        for col in mtlf_cols:
+            df[col] = pd.to_numeric(df[col], errors="coerce").round().astype("Int64")
 
         return df
 
--- a/gridstatus/miso_api.py
+++ b/gridstatus/miso_api.py
@@ -1522,6 +1522,7 @@ class MISOAPI:
                 "Publish Time",
                 "Region",
                 "Local Resource Zone",
+                "init",
             ]:
                 data[col] = data[col].astype(float)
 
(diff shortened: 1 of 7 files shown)


gridstatus/miso_api.py
```diff
@@ -1555,6 +1556,59 @@ def get_medium_term_load_forecast_hourly(
...
+        for offset in range(1, 8):
+            forecast_date = publish_time + pd.DateOffset(days=offset)
+            df = self._get_medium_term_load_forecast(
+                date=forecast_date,
+                verbose=verbose,
+                publish_time=publish_time,
+                time_resolution=HOURLY_RESOLUTION,
+            )
+            if not df.empty:
+                all_dfs.append(df)
```
WillKoehrsen: Querying for multiple days is what separates this method from `get_medium_term_load_forecast_hourly`?

gridstatus/nyiso.py
```diff
@@ -347,6 +349,40 @@ def get_btm_solar_forecast(
...
+
+        Source: https://mis.nyiso.com/public/P-70Clist.htm
+        """
+        data = self._download_nyiso_archive(
+            date=date,
+            end=end,
+            dataset_name=BTM_INSTALLED_CAPACITY_DATASET,
+            filename="btminstalledcapacitytracking",
+            verbose=verbose,
+        )
```
WillKoehrsen, line 373: `"latest"` gives the error: `DateParseError: Unknown datetime string format, unable to parse: latest, at position 0` which I think means it's not supported?

WillKoehrsen (approved):

gridstatus/miso_api.py
Kladar: Ah good find, I didn't realize we had that (and it should probably be `mid_term` rather than `medium_term`? but digression). But yeah same data, slightly different slice, "by forecasted day" vs "by publish vintage across the all horizons". I could see us combining them, keeping both, or using the existing method, happy to go any direction (none standing out to me as obviously best right now).

gridstatus/nyiso.py
Kladar: Yeah not supported, I throw a custom error now for "latest"

gridstatus/miso_api.py
WillKoehrsen: I think we should only have one method to avoid confusion and I like your implementation since it fetches the full forecast. We may want to make the maximum offset a parameter though so if you only want one day out you can get that.

Kladar: Okay I've added the parameter and removed the existing `get_medium_term_load_forecast_hourly` method and its tests. I think that method was an outside contribution to the repo so there might be users for it. I added the removal to the changelog as a breaking change.

PR #894, opened 2026-06-16
feat(ENG 4133): Add CAISO Aggregated Generation Outages scraper

Kladar:
## Summary
- Adds `get_aggregated_generation_outages` scraper for CAISO

### Details
This adds a new CAISO OASIS GroupZip scraper for the `AGGR_OUTAGE_SCH_GRP` report (Energy → System → Aggregated Generation Outages). Each daily query returns roughly 30 days of forward-looking hourly outage MW by fuel category (Hydro, Renewable, Thermal, and related groupings) and trading hub (SP15, NP15, ZP26, PACE, PACW). The method uses the standard OASIS GET flow via `get_oasis_dataset`, maps `REPORT_DATE_GMT` to `Publish Time`, and returns Interval Start, Interval End, Publish Time, Fuel Category, Trading Hub, and MW.

To run tests:
```bash
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "caiso" -k "aggregated_generation_outages"
```

### DIFF
--- a/gridstatus/tests/source_specific/conftest.py
+++ b/gridstatus/tests/source_specific/conftest.py
@@ -1,23 +1,36 @@
+import os
 import time
 
 import pytest
 
-from gridstatus.tests.vcr_utils import RECORD_MODE
+CAISO_OASIS_RECORD_COOLDOWN_SECONDS = 30
 
-CAISO_OASIS_RECORD_COOLDOWN_SECONDS = 15
+_last_caiso_oasis_record_time: float | None = None
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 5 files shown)


gridstatus/tests/source_specific/conftest.py
Kladar, line 1: not entirely sure this is needed, but I couldn't seem to get `real_sleep` to actually sleep between the support date range date requests on the API, which would fail the test. The API rate limit affects back-to-back tests but also back-to-back requests in the same test, which we need to do for this dataset given the group_zip structure of the responses. That this hasn't come up before makes me think we don't actually need it, but it does work to get these tests passing.

WillKoehrsen (approved):

PR #902, opened 2026-06-24
fix(ENG-4133): Pivot CAISO aggregated generation outages to wide schema

Kladar:
## Summary
- Pivot `get_aggregated_generation_outages` from long format (`Fuel Category` + `MW`) to wide format with per-category columns (`Thermal`, `Renewable`, `Hydro`, `Aggregated`, `Not Available`)
- One row per `(Interval Start, Publish Time, Trading Hub)` instead of per fuel category
- Rename source `Not Avail` to `Not Available`

## Test plan
- [x] `VCR_RECORD_MODE="all" pytest -vv gridstatus/tests/source_specific/test_caiso.py -k "outages"`

### DIFF
--- a/gridstatus/caiso/caiso.py
+++ b/gridstatus/caiso/caiso.py
@@ -2291,6 +2291,12 @@ class CAISO(ISOBase):
             if column not in df.columns:
                 df[column] = np.nan
 
+        component_sum = df[["Hydro", "Not Available", "Renewable", "Thermal"]].sum(
+            axis=1,
+            min_count=1,
+        )
+        df["Aggregated"] = df["Aggregated"].combine_first(component_sum)
+
         return df[
             [
                 "Interval Start",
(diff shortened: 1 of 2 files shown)


Kladar:
The `Not Avail` to `Not Available` rename is optional, but seems nice.

gridstatus/caiso/caiso.py
```diff
@@ -2266,10 +2313,12 @@ def get_aggregated_generation_outages(
         sleep: int = 4,
         verbose: bool = False,
     ) -> pd.DataFrame:
-        """Return hourly aggregated generator outages by fuel category and trading hub.
+        """Return hourly aggregated generator outages by trading hub.
```
WillKoehrsen, line 2322: Is `aggregated` not the sum of outages for the trading hub?

Image

Kladar: It seems like `aggregated` only has values for ZP26 and the others never have an `aggregated` value. So maybe it makes sense to create a `Total` column that adds them to avoid the sparse `aggregated` column

Kladar: ...or just put that calculated total in the `aggregate` column

WillKoehrsen: I think for a dataset called `aggregated`, we should have a total column. As to whether that should be a new column or put the total in `aggregate`, I don't have strong feelings.

Kladar: I agree, and I think the aggregated column is the most logical place, since it already exists and can simply be made more dense by summing the other columns (and then ZP26 just doesn't have disaggregation, which seems fine from a user standpoint).

Implemented the sum in the aggregate column

WillKoehrsen: That makes sense. Also, having an `aggregated` and `total` column could be confusing especially since `aggregated` is a sparse column.

WillKoehrsen (approved):
👍🏽

PR #909, opened 2026-06-30
feat(ENG 4236): Add MISO Area Control Error scraper

Kladar:
## Summary
- Adds `get_area_control_error` scraper for MISO

### Details
Adds `MISO.get_area_control_error`, which returns the system-level Area Control Error (ACE) in MW from the public MISO API (`https://public-api.misoenergy.org/api/Ace`). ACE is published as instantaneous observations roughly every 30 seconds. The source endpoint only serves a rolling ~2-hour window of recent observations and has no historical/date query path, so only `date="latest"` is supported (a `NotSupported` error is raised otherwise), matching the pattern used by `get_interchange_5_min`. The source `instantEST` timestamps are parsed in the MISO default EST timezone into a `Time` column, and `value` is converted to a numeric `Area Control Error` column. The returned DataFrame has columns `Time` and `Area Control Error`.

This method lives in the `MISO` class rather than `MISOAPI` because ACE comes from the public, unauthenticated `public-api.misoenergy.org` host (the same host as `get_fuel_mix`, `get_load`, and `get_interchange_5_min`), whereas `MISOAPI` is a client for the authenticated Data Exchange APIM endpoints with subscription keys and pagination that do not apply here.

To run tests:
```bash
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "miso" -k "area_control_error"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 #### ERCOT
 * ERCOT Available Resource Planned Outage Capacity 7 Day and Future datasets in [#884](https://github.com/gridstatus/gridstatus/pull/884)
 
+#### MISO
+* MISO Area Control Error dataset via `MISO.get_area_control_error`
+
 ## v0.36.0 - April 20, 2026
 
 ### Additions (New Features/Datasets)
--- a/gridstatus/miso.py
+++ b/gridstatus/miso.py
@@ -1951,6 +1951,52 @@ class MISO(ISOBase):
             .reset_index(drop=True)
         )
 
+    @support_date_range(frequency=None)
+    def get_area_control_error(
+        self,
+        date: str | pd.Timestamp = "latest",
+        end: str | pd.Timestamp = None,
+        verbose: bool = False,
+    ) -> pd.DataFrame:
+        """Get the MISO Area Control Error (ACE).
+
+        Real-time system-level area control error in MW, published as instantaneous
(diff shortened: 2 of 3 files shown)


WillKoehrsen (approved):
Nice, data looks good.

WillKoehrsen:
Extremely minor nit:

The description says to run tests use:

```
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "miso" -k "area_control_error"
```

However, `-k` can only hold one value at a time so the `-k "miso"` is ignored and this also picks up the PJM area control error tests.

Screenshot 2026-06-30 at 12 52 34

To run only the miso area control error tests, we can use `-k "miso and area_control_error"`

Screenshot 2026-06-30 at 12 56 52

Kladar:
> Extremely minor nit:
> 
> The description says to run tests use:
> 
> ```
> VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "miso" -k "area_control_error"
> ```
> 
> However, `-k` can only hold one value at a time so the `-k "miso"` is ignored and this also picks up the PJM area control error tests.
> Screenshot 2026-06-30 at 12 52 34
> 
> To run only the miso area control error tests, we can use `-k "miso and area_control_error"`
> Screenshot 2026-06-30 at 12 56 52

TIL, nice, I'll add a note to the claude skill/cursor command

PR #911, opened 2026-07-01
feat(ENG-4238): Add date-range support for PJM emergency postings

Kladar:
## Summary
- Adds date range support to `get_emergency_postings()` that fetch via PJM's public REST endpoint (`/ep/rest/public/posting`). Originally we found the dashboard parsing to be faster and easier, but the public rest endpoint also works. 
- Existing behavior is now called `latest`, since we can get the most recent messages via that original pathway and the downstream approach is mostly unchanged. 

## Why
The live ETL only captured recent postings because the dashboard export defaults to a rolling 2–3 day filter. Historical backfill requires querying by date range; the REST API supports this without JSF session management. This option didn't seem available when originally implemented, but is helpful for backfills. No need to boil the ocean here.

## Test plan
- [x] `uv run pytest gridstatus/tests/source_specific/test_pjm.py -k emergency -vv`
- [ ] Merge and tag; gs-etl PR depends on this rev

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -3778,7 +3778,7 @@ class PJM(ISOBase):
             .reset_index(drop=True)
         )
 
-    @support_date_range(frequency="YEAR_START")
+    @support_date_range(frequency=None)
     def get_emergency_postings(
         self,
         date: str | pd.Timestamp,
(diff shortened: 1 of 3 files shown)


Kladar:
One second, need to think about this

WillKoehrsen (approved):
Nice, I was able to fetch data back to May 2004 and tests pass locally.

PR #912, opened 2026-07-01
feat(ENG 3096): Add EIA Facility Fuel scraper

Kladar:
## Summary
- Adds `get_facility_fuel` scraper for EIA

### Details
Pulls monthly electric power operations for individual power plants from the EIA v2 API `electricity/facility-fuel` endpoint (Form EIA-923 data). Each row is a plant/fuel/prime-mover combination for a month and includes net and gross generation, fuel consumption in physical units and BTUs, and average heat content, along with the units columns the API provides. The data includes EIA's roll-up rows (`ALL` fuel and prime mover codes and the synthetic plant code 99999 state-fuel-level increment rows), which are kept as reported. Both `fuel2002` (as Fuel Code) and `fuelType` (as Fuel Type) are kept because a plant month can report the same 2002-era fuel code under multiple modern fuel types (for example WAT as both HYC and HPS), so the fuel type is needed to uniquely identify rows. Pagination is sequential with a stable sort to avoid the API's unstable ordering across requests. The Updated At column is the publication time of each row's period, taken from the Last-Modified header of the per-year EIA-923 zip file on the EIA-923 page (checking the current, early-release, archive, and pre-2008 form 906/920 URL locations), falling back to the fetch time if no file is found. Data is available back to 2001.

To run tests:
```bash
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "eia and facility_fuel"
```

### DIFF
--- a/gridstatus/eia.py
+++ b/gridstatus/eia.py
@@ -683,8 +683,6 @@ class EIA:
             "data": FACILITY_FUEL_DATA_FIELDS,
             "offset": 0,
             "length": page_size,
-            # pagination breaks if not sorted because
-            # api doesn't return in stable order across requests
             "sort": [
                 {"column": col, "direction": "asc"}
                 for col in ["period", "plantCode", "fuel2002", "fuelType", "primeMover"]
--- a/gridstatus/eia.py
+++ b/gridstatus/eia.py
@@ -38,6 +38,13 @@ HENRY_HUB_NATURAL_GAS_SPOT_PRICES_PATH = "natural-gas/pri/fut"
 # Physical location of Henry Hub is Louisiana
 HENRY_HUB_TIMEZONE = "US/Central"
 
+FACILITY_FUEL_ZIP_URLS = [
+    "https://www.eia.gov/electricity/data/eia923/xls/f923_{year}.zip",
+    "https://www.eia.gov/electricity/data/eia923/xls/f923_{year}er.zip",
+    "https://www.eia.gov/electricity/data/eia923/archive/xls/f923_{year}.zip",
+    "https://www.eia.gov/electricity/data/eia923/archive/xls/f906920_{year}.zip",
+]
+
 
 class EIA:
     BASE_URL = "https://api.eia.gov/v2/"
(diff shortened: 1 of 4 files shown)


gridstatus/eia.py
```diff
@@ -637,6 +641,115 @@ def get_henry_hub_natural_gas_spot_prices(self, date, end=None, verbose=False):
...
+        headers = {
+            "X-Api-Key": self.api_key,
+            "X-Params": json.dumps(params),
+        }
+
+        if verbose:
+            logger.info(f"Fetching data from {url}")
+            logger.info(f"Params: {params}")
+
+        updated_at = pd.Timestamp.utcnow().floor("s")
```
WillKoehrsen: I'm not sure about this `updated_at` value since it's more an `accessed_at` time.

If we want to get an updated time, I found some options:

There are some release dates here, but they may be hard to scrape.

https://www.eia.gov/electricity/data/eia923/

Image

If you download the zip file and look at the file properties, you can see the created and modified time.

https://www.eia.gov/electricity/data/eia923/xls/f923_2026.zip

Image

Kladar: Okay, I evaluated these options and this is now using the last modified from the response for the zip files. I think this is a good tradeoff between complexity and accuracy. From Claude:

> The Updated At column comes from the publication time of the per-year EIA-923 zip file rather than the time we fetched the data. The API itself exposes no update timestamp anywhere (the route metadata returns updated: None and the data response has nothing), so we use the zip distribution channel as a proxy for when EIA last published each year's data. For each year in the requested range we send a single HEAD request and read the Last-Modified header off the zip, no download needed. The URL moves through EIA's release cycle, so we try four known locations in order: xls/f923_{year}.zip for the current year, xls/f923_{year}er.zip for the early release, archive/xls/f923_{year}.zip once finalized, and archive/xls/f906920_{year}.zip for 2001-2007 when the data came from forms 906/920. Missing URLs 301-redirect to a landing page instead of returning a 404, so we check for a 200 with the header present and fall back to fetch time with a logged warning if no file is found. Every month in a year shares one timestamp since the file is per-year, and each monthly republish bumps the whole year's timestamp, so it should be read as "the release this row's year belongs to was last published at this time" rather than "this exact row was last revised at this time." One known assumption: the API and the zip files could update a day or two apart, so the timestamp reflects the zip channel's release event, which we take as close enough for a monthly dataset revised on a months-long cycle. Spot checks across the history look right: 2001 stamps as 2011-06-03 (last archive touch), 2015 as 2022-08-05, 2024 as 2026-01-12, 2025 as 2026-06-30 (early release), and 2026 as 2026-06-24 (current file).

WillKoehrsen (approved):
Data looks good including the `updated_at` time. Tests pass locally.

PR #918, opened 2026-07-15
fix(PJM Emergency Postings): capture `is_drill` field on PJM Emergency Postings

Kladar:
## Summary
Adds the field from the parsed XML report

### DIFF
--- a/gridstatus/pjm.py
+++ b/gridstatus/pjm.py
@@ -3881,7 +3881,6 @@ class PJM(ISOBase):
                     {
                         "Message ID": int(mid.strip()),
                         "Message Type": msg.findtext("messageType"),
-                        "Is Drill": is_drill,
                         "Priority": msg.findtext("priority"),
                         "Emergency Message": msg.findtext("message"),
                         "Region": rn,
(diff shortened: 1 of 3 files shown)


gridstatus/pjm.py
```diff
@@ -3874,11 +3874,14 @@ def _parse_emergency_xml(self, xml_bytes: bytes) -> pd.DataFrame:
                 [r.findtext("regionName") for r in regions] if regions else [None]
             )
 
+            is_drill = (msg.findtext("pjmDrill") or "false").strip().lower() == "true"
```
WillKoehrsen, line 3877: It looks like we're not capturing the "DRILL" in the "Message Type". 

Image
Image

Kladar: I think it's just a banner that gets created when the `pjmDrill` field is true. On the page, `DRILL -` is prefixed onto Message Type (and there’s a separate pink Drill banner in the message body) when `pjmDrill` is true.

In the XML for message 103331, `messageType` is just:

```
Manual Load Dump Action - Capacity Emergency - NERC EEA 3
```

with `<pjmDrill>true</pjmDrill>` as its own field. Do we want to manually add it to our message type parsing?

kmax12: my vote is to just add the boolean column

Kladar: that is also my vote

kmax12 (approved):
LGTM

PR #924, opened 2026-07-22
feat(ERCOT): Add published_after to price correction getters

WillKoehrsen:
## What changed

All nine ERCOT price correction getters accept an optional `published_after` timestamp, passed through to `_get_documents`, which filters the MIS document listing before any file is downloaded. Callers that track what they have already ingested can skip re-reading every listed file — the electrical bus corrections alone are ~9M rows across 80+ files per run. When nothing new is listed, the getters raise `NoDataFoundException` as they already do for an empty listing.

The LMP price correction handler now processes files in publish order, so when multiple files carry a row for the same key the latest published row appears last. Rows are returned as-is — duplicate-key rows in the source files are left for consumers to resolve.

## How to validate

```
uv run pytest gridstatus/tests/source_specific/test_ercot.py -k "published_after or price_corrections"
```


Kladar (approved):
Makes sense. The downstream use of this is probably where I have to understand more but this seems good. 

Note to self: these tests break after 2100-01-01.

PR #930, opened 2026-07-30
Adopt ruff 0.16's default ruleset

WillKoehrsen:
Two things: adopt ruff 0.16's default rule set, and stop the pre-commit hook and CI from running different ruff versions.

## The version was split-brained

The pre-commit hook was pinned to `v0.13.2` while `uv.lock`, which is what CI's `make lint` resolves, gave `0.11.5`. The local hook was enforcing a different rule set than CI. Both now sit on the same **exact** version, and the config comment says why they have to match.

## Config

Drops the hand maintained `select` in favour of `extend-select`: inherit the 0.16 defaults, name only the families outside them. Future default additions then arrive with a version bump instead of a config edit.

## Real bugs this surfaced

- **`ercot.py`** defined `_make_offer_curve` inside `for as_name in AS_PRODUCTS` and closed over the loop variable. It happens to work because `.apply()` runs it in the same iteration, but `as_name` is now bound as a default so the closure captures the right value by construction rather than by luck.
- **`test_isone_api`** asserted `location_id == location_id`. That test asserted nothing at all. It now checks the map's real contract, that every zone resolves to an `int`.
- **`base.py`** called `method(date="today", *args, **kwargs)`. Python allows it, but it evaluates as `method(*args, date="today", ...)`, so a non-empty `args` would collide on `date`. Now written in the order it actually executes.
- **Four self-assignments** (`end = end`, `date = date`, and `start = start = ...` twice in a docs notebook) were dead lines left behind by edits that lost their other half.
- **`ercot_api`** leaked two file handles via `json.load(open(...))`.
- **Two test helpers** took a mutable list literal as a default. It is only ever read through `set()`, so it is now a module level tuple.
- **`eia.py`** had two handlers that caught `HTTPError` only to re-raise it.

## Ignored, with reasons

| Rule | Why |
|---|---|
| `RUF012` | class level mutable defaults here are declarative dataset config |
| `SIM117` | nested `with` blocks read better than one merged statement |
| `TRY002`, `TRY003` | vanilla exceptions with messages at the raise site are the style here |
| `TRY004` | swapping `ValueError` for `TypeError` changes what callers catch |
| `BLE001` | this library scrapes HTML, CSV, XML and PDFs from a dozen ISOs, where nearly every broad handler either moves to the next document or re-raises as `RequestException` on purpose. Keeping it on would mean ~20 inline noqas rather than a signal. |

## The bulk

Mechanical: **140 unused `noqa` directives** (`E501` is globally ignored here, so every `# noqa: E501` was dead), typing modernisation to PEP 585/604, 28 implicit `Optional`s made explicit, and the usual comprehension and f-string cleanups.

One typing change shortened a signature enough that `ruff-format` collapsed it onto one line and `add-trailing-comma` then re-expanded it. The committed form is where the two hooks agree.

## How to validate

- `make lint`
- `uv run pre-commit run --all-files`, now with ruff at 0.16.0
- `make test` on Python 3.11, the version CI uses

The parts worth a close read are the four behaviour changes listed above: the `as_name` closure binding in `ercot.py`, the argument reordering in `base.py`, the `.items()` rewrite in `ieso.py`, and the `KeyError` narrowing in `test_get_iso_invalid`. Everything else is mechanical.


Kladar (approved):
TIL about `!s`, `!r`, and `next()`.

Lots of nice little improvements here, nothing that raised any red flags for breaking changes (but we can keep an eye on it once merged).

PR #931, opened 2026-07-31
Add Energy Storage to SPP fuel mix combine

Kladar:
## Summary
- GenMix2Hour includes `Energy Storage Market` / `Energy Storage Self`
- Add Energy Storage to `process_gen_mix` `columns_to_combine` so non-detailed fuel mix exposes a single `Energy Storage` column
- Tests allow Energy Storage as an optional column (present in GenMix2Hour, absent from GenMix365)

## Test plan
- [ ] `get_fuel_mix(date="latest")` includes `Energy Storage` and not Market/Self columns
- [ ] `get_fuel_mix_detailed(date="latest")` keeps `Energy Storage Market` / `Energy Storage Self`
- [ ] Historical GenMix365 fetches still succeed without Energy Storage columns

### DIFF
--- a/gridstatus/spp.py
+++ b/gridstatus/spp.py
@@ -3084,6 +3084,7 @@ def process_gen_mix(df: pd.DataFrame, detailed: bool = False) -> pd.DataFrame:
         "Waste Disposal Services",
         "Wind",
         "Waste Heat",
+        "Energy Storage",
         "Other",
     ]
 
--- a/gridstatus/tests/source_specific/test_spp.py
+++ b/gridstatus/tests/source_specific/test_spp.py
@@ -79,6 +79,15 @@ class TestSPP(BaseTestISO):
         "Other Self",
     ]
 
+    FUEL_MIX_OPTIONAL_COLS = [
+        "Energy Storage",
+    ]
+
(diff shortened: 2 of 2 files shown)


gridstatus/spp.py
```diff
@@ -393,38 +406,55 @@ def _get_fuel_mix(
...
+                columns=["Short Term Load Forecast", "Average Actual Load", "Time"],
+                errors="ignore",
+            )
+            dfs.append(df_part)
 
-        df = df.drop(
-            columns=["Short Term Load Forecast", "Average Actual Load", "Time"],
-            errors="ignore",
-        )
+        df = pd.concat(dfs, ignore_index=True)
```
kmax12: personally, ive grown to feel that that the magic of combining files shouldn't be in these scraper. i think it's easier to reason about when each of these `get_*` method corresponds to a single resource on the other end. 

if the caller wants to combine them, they can, but i feel like less magic is better at this layer. 

I'd be in favor of moving a head with the version of this PR that just adds support for the new column, but I'm not sold on this concatenation and deduping as of now.

WillKoehrsen (approved):

PR #938, opened 2026-08-24
Raise Correct Error When No Data Found for EIA

WillKoehrsen:
## Summary

- Raises `NoDataFoundException` when no data is found for an EIA dataset allowing downstream callers to handle this exception

### DIFF
--- a/gridstatus/tests/source_specific/test_eia.py
+++ b/gridstatus/tests/source_specific/test_eia.py
@@ -5,6 +5,7 @@ import pandas as pd
 import pytest
 
 import gridstatus
+from gridstatus.base import NoDataFoundException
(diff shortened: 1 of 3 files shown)


gridstatus/tests/source_specific/test_eia.py
```diff
@@ -186,6 +187,27 @@ def test_fuel_type():
...
+
+    with pytest.raises(
+        NoDataFoundException,
+        match="No EIA data found for electricity/rto/interchange-data",
+    ):
+        eia.get_dataset(
+            dataset="electricity/rto/interchange-data",
+            start=start,
+            end=end,
+        )
```
Kladar, line 208: could probably just fixturize an empty dataframe rather than calling source here

Kladar (approved):

PR #939, opened 2026-08-24
Handle the Offer Type column in ERCOT 60-day SCED Resource AS Offers files

WillKoehrsen:
## What changed

ERCOT's 60-Day SCED Disclosure Resource AS Offers files starting with Operating Day June 25, 2026 (published August 24, 2026) carry an explicit `Offer Type` column (`ONRES`/`OFFNS`/`REGDN`). The previous logic inferred the curve type from which price columns held values and raised `ValueError: Unknown curve type found` on these files, because they contain:

- 370,400 padding rows with no price or quantity values (every resource is padded out to all three offer types)
- 2,700 ECRS-only rows appearing under both `ONRES` and `OFFNS`, which no price-column pattern can tell apart

`process_sced_resource_as_offers` now maps `Offer Type` directly to the curve type when the column is present, drops the empty padding rows, and raises on unrecognized offer type values. Files without the column keep the existing inference path unchanged.

## How to validate

- `pytest gridstatus/tests/source_specific/test_ercot.py -k TestProcessScedResourceAsOffers` (6 new tests, 12 existing pass unchanged)
- `Ercot().get_60_day_sced_disclosure(date=pd.Timestamp('2026-06-25', tz='US/Central'), process=True)` completes against the real file: 1,188,350 rows, curve type counts match the per-offer-type non-empty row counts, no duplicate (SCED Timestamp, Resource Name, Curve Type) keys, no all-empty curve rows.


Kladar (approved):
Tests passing

PR #940, opened 2026-08-28
Handle ERCOT's capped/uncapped SCED price column split

WillKoehrsen:
## What changed

On 2026-08-27 ERCOT implemented NPRR1290 and NPRR1323 and renamed the price column in three SCED reports with a `Capped` prefix, adding an `Uncapped` counterpart:

| Report | Before | Now |
|---|---|---|
| SCED System Lambda (NP6-322-CD) | `SystemLambda` | `CappedSystemLambda`, `UncappedSystemLambda` |
| LMPs by Electrical Bus (NP6-787-CD) | `LMP` | `CappedLMP`, `UncappedLMP` |
| Real-Time MCPC by SCED Interval (NP6-332-CD) | `MCPC` | `CappedMCPC`, `UncappedMCPC` |

All three parsers index the price column by name, so they raised `KeyError` on every file published after the cutover.

The capped value is the same series these methods have always returned, so it keeps its existing column name and no existing output changes meaning. The uncapped value is exposed as a new column (`Uncapped System Lambda`, `Uncapped LMP`, `Uncapped MCPC`). The rename is applied only when the capped name is present, so pre-cutover files still parse and get a null uncapped value.

Only the electrical bus LMP report changed. `_handle_lmp_df` is shared with `get_lmp_by_settlement_point`, so the uncapped column is gated on `location_type` (threaded down from `_get_lmp`) rather than sniffed from the frame. That keeps settlement point output exactly as it was, and keeps the bus output schema stable across the cutover boundary instead of flickering by fetch date.

`ErcotAPI._handle_lmp_by_bus` gets the same treatment, since that path serves older dates.

`_handle_sced_system_lambda` is split into a `_handle_sced_system_lambda_df` taking a DataFrame, matching the existing `_handle_lmp` / `_handle_lmp_df` split, so the frame logic is unit-testable without cassettes.

## How to validate

Against live ERCOT data:

```python
iso = gridstatus.Ercot()
now = pd.Timestamp.now(tz=iso.default_timezone)

# Post-cutover: uncapped column populated
iso.get_sced_system_lambda(now - pd.Timedelta(minutes=30), now)
iso.get_mcpc_sced(now - pd.Timedelta(minutes=30), now)
iso.get_lmp_by_bus(now - pd.Timedelta(minutes=30), now)

# Pre-cutover (MIS retains ~5 days): parses, uncapped is NaN
pre = (now - pd.Timedelta(days=4)).normalize() + pd.Timedelta(hours=12)
iso.get_sced_system_lambda(pre, pre + pd.Timedelta(minutes=30))

# Unchanged
iso.get_lmp_by_settlement_point(now - pd.Timedelta(minutes=30), now)
```

I ran all of the above. Post-cutover rows carry both values (identical right now, since they only diverge when the price would exceed VOLL), pre-cutover rows return the uncapped column as NaN, and `get_lmp_by_settlement_point` columns are unchanged.

Unit tests:

```
uv run pytest gridstatus/tests/source_specific/test_ercot.py -k CappedUncapped
```

7 tests covering both layouts for each of the three reports, plus a guard that settlement point output does not gain the column.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 * ERCOT 2-Day Aggregate Gen Summary, Load Summary, and Output Schedule datasets via `Ercot.get_aggregate_gen_summary_2_day`, `Ercot.get_aggregate_load_summary_2_day`, and `Ercot.get_aggregate_output_schedule_2_day`
 * ERCOT LMP Price Corrections by settlement point and by electrical bus via `Ercot.get_lmp_by_settlement_point_price_corrections` and `Ercot.get_lmp_by_bus_price_corrections`
 * ERCOT price correction getters accept `published_after` to skip documents already ingested
+* ERCOT uncapped SCED prices via the new `Uncapped System Lambda`, `Uncapped LMP`, and `Uncapped MCPC` columns on `Ercot.get_sced_system_lambda`, `Ercot.get_lmp_by_bus`, and `Ercot.get_mcpc_sced`. ERCOT began publishing these alongside the capped prices on August 27, 2026 with NPRR1290 and NPRR1323. The two values differ only when the price would exceed the Value of Lost Load. Files published before that date have no uncapped value, so the column is null for earlier data.
 
 #### MISO
 * MISO Area Control Error dataset via `MISO.get_area_control_error`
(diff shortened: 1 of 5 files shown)


kmax12 (approved):
ran locally as expected

PR #945, opened 2026-09-02
Handle the capped/uncapped column split in ERCOT SCED price corrections

WillKoehrsen:
On 2026-09-01 the two ERCOT price correction reports keyed by SCED run (`RTM_MCPC_SCED` and `RTM_EBLMP`) followed the 2026-08-27 SCED report change from #940: the original and corrected price columns gained a `Capped` prefix and `Uncapped` counterparts. The bus report also renamed `ElectricBusName` to `ElectricalBus` and `DSTFlag` to `RepeatedHourFlag`. `get_mcpc_sced_price_corrections` raised `KeyError` on the new files, and `get_lmp_by_bus_price_corrections`, which fetches files from both sides of the cutover in one call, ended up with two `RepeatedHourFlag` columns after concatenation and failed in `tz_localize` with `Buffer has wrong number of dimensions`.

Each file is now normalized to the legacy column names before concatenation. The capped value keeps its existing column name, matching #940, and the uncapped value is exposed as new `Uncapped MCPC Original` / `Uncapped MCPC Corrected` and `Uncapped LMP Original` / `Uncapped LMP Corrected` columns that are null for pre-cutover files. The settlement point report did not change, so its output is untouched.

## Validation

- `TestCappedUncappedPriceCorrectionColumns` feeds synthetic pre-cutover, post-cutover, and mixed files through the real handlers.
- `VCR_RECORD_MODE=all uv run pytest gridstatus/tests/source_specific/test_ercot.py -k price_corrections` passes against live ERCOT.
- Fetching both reports with `published_after=2026-07-23` (the window that failed in production) returns the July and September files in one frame, with uncapped values null only for the July rows.


Kladar (approved):
lgtm

PR #947, opened 2026-09-11
feat(ENG 4562): Add CAISO AS Prices Real Time 15 Min scraper

Kladar:
## Summary
- Adds `get_as_prices_real_time_15_min` scraper for CAISO

### Details
Fetches OASIS PRC_INTVL_AS (market_run_id=RTM), which returns one operating hour of 15-minute ancillary service prices per request. The output is pivoted on ANC_TYPE into the same wide product columns as `get_as_prices` (Non-Spinning Reserves, Regulation Down/Up, Regulation Mileage Down/Up, Spinning Reserves) for the six CAISO AS regions.

To run tests:
```bash
VCR_RECORD_MODE=all uv run pytest -vvv gridstatus/tests/ -k "caiso and as_prices_real_time_15_min"
```


Kladar:
One sec, dropping the `Time` column since we don't need it

Kladar:
okay, good to go

gridstatus/caiso/caiso.py
```diff
@@ -2449,6 +2449,100 @@ def get_as_prices(
...
+            sleep=sleep,
+            verbose=verbose,
+            raw_data=False,
+        )
+
+        columns = [
+            "Interval Start",
+            "Interval End",
+            "Region",
+            "Market",
```
WillKoehrsen: It looks like the Market is always "RTM". Is this information we need to record?

gridstatus/caiso/caiso.py
```diff
@@ -2449,6 +2449,100 @@ def get_as_prices(
...
+            return pd.DataFrame(columns=columns)
+
+        df = df.rename(
+            columns={
+                "ANC_REGION": "Region",
+                "MARKET_RUN_ID": "Market",
+            },
+        )
+
+        as_type_map = {
```
WillKoehrsen: Do we use this map elsewhere? It might be easier to maintain if defined once.

gridstatus/caiso/caiso.py
Kladar: Dropped it. Every value is RTM, so it just wastes space. PK is now Interval Start + Region.

gridstatus/caiso/caiso.py
Kladar: Good call — it was duplicated in `get_as_prices`, this method, and `get_as_procurement`. Moved to `AS_TYPE_MAP` in `caiso_constants.py` (includes the XML `*_CLR_PRC` aliases this report sometimes uses).

WillKoehrsen (approved):
👍🏽

